PR #389, opened 2024-05-03
Watch dayBoundary and forward updates to Qalendar.

### LINKED ISSUE #384: Fix booking hours on schedule and calendar pages
Currently where there should be nothing there's empty time. 

We should only show non-empty time with about 2 hour padding on either side. This is working on bookings page iirc, but I might have broke it with the new calendar component.

image

MelissaAutumn:
Fixes #384 

Adds a watch for dayBoundary and forwards the data to qalendar.time. In Qalendar time is created from the config value and is not referenced directly again. So we need to poke at it manually.

image

It looks a bit silly with one event, but we can adjust that as needed I think!

(Please give it a test 😄 )

### DIFF
--- a/frontend/src/components/CalendarQalendar.vue
+++ b/frontend/src/components/CalendarQalendar.vue
@@ -254,15 +254,15 @@ const calendarEvents = computed(() => {
  * @type {ComputedRef<{start, end}>}
  */
 const dayBoundary = computed(() => {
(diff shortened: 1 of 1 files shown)


malini (approved):

devmount (approved):
Tested. Works on my end, thanks!

PR #399, opened 2024-05-08
Disable the "Generate Zoom Meeting" checkbox if they don't have a zoom account connected.

MelissaAutumn:
Fixes #389 

Spotted via a Sentry bug report and then confirmed it was a bug. Easy fix thankfully.


devmount (approved):
Good catch! Makes totally sense to only enable that input when a zoom account is connected 😅

PR #400, opened 2024-05-08
Invite panel

### LINKED ISSUE #362: Build out mini-admin panel
For alpha we'll need an admin panel to manage invite code generation, and maybe user lookup/disable. 

We can just use the default tailwind table styling:
https://tailwindcss.com/docs/table-layout

We'll need a few endpoints: 
 * [x] Query all invites
 * [x] Generate `n` invites
 * [x] Revoke a specific invite
 
 (Maybe out of scope for alpha):
 * List all subscribers (data limited to id, username, invite code used (or None), date created)
 * Disable a specific subscriber

Admin panel is gated via a custom middleware check on the backend and some hard coded strings on the frontend. So if someone really wanted to look at the admin panel they probably could, but the backend function would prevent any of the functionality (like showing, or performing actions) from working. 

We can fine-tune our database models later...

devmount:
## Description of the Change

This change introduces admin only views. Those are currently protected by a dependency checking, if the current users email address is whitelisted in a .env var `APP_ADMIN_ALLOW_LIST` (comma separated list of email addresses).

New route `/admin/invites`:
image

Invite codes list is paginated, filterable, and codes can be bulk generated, copied and revoked.

New route `/admin/subscribers`:
Untitled

For now, subscribers can only be listed. Showing invite status and disabling them is still todo.

## Benefits

Administration tools for quick management of invites and subscribers.

## Applicable Issues

Closes #362


MelissaAutumn:
I'll fix up the PR with any changes needed!

MelissaAutumn:
I accidentally rebased this on my old changes, so I had to go through and download / apply patch files from the now headless remote. It should be fixed now, if not a little out of order. 

Apologies!

MelissaAutumn:
Idea pivot! Have a subscriber panel with a invite input at the footer. This invite function will create the subscriber and email them about Appointment. 

I wrote a generic data table component that takes in structured data and displays it. There's some extra functionality I decided to cut for now, but are still implemented in the data table. I'm sure there's an easier way to do filtering than keeping the source list and a filtered version but I cannot think of a good way to do it right now. 

If someone could pretty-ify the placement of the data filters that would be great 😅 

image

backend/src/appointment/routes/auth.py
```diff
@@ -237,6 +241,13 @@ def me(
         timezone=subscriber.timezone, avatar_url=subscriber.avatar_url
     )
 
+
+@router.post("/permission-check")
+def permission_check(_: Subscriber = Depends(get_admin_subscriber)):
+    """Checks if they have admin permissions"""
+    return True
```
MelissaAutumn, line 248: The check is done in get_admin_subscriber, and will error out. FastAPI has a weird relationship with using DI over middleware...

MelissaAutumn:
Additionally all admin checks will fail if `APP_ADMIN_ALLOW_LIST` is empty.

devmount:
No worries about the rebase, everything looks fine now. Sorry you had some additional effort because of that.

The new subscriber panel is nice, also simplifies things for not having different admin pages for now. I fixed up some alignment and set the horizontal overflow scrollbar to auto (as for some systems, the scrollbar would be always visible with "scroll" 🙈 ). Also updated the language files (my IDE sorted the JSON keys, sorry about that 😇 ).

So looking fine now! Idea for the data filters: We could integrate them into the table header beneath the column label. But I think it's also fine to have them above the table 🤷🏻

MelissaAutumn:
Thank you! Also thank you for re-sorting the language files as always!

PR #407, opened 2024-05-10
Fix all day events bleeding over to following day:

MelissaAutumn:
* Google defines all day as "Midnight" to "Midnight + 1 day"
* Subtract 1 minute from the end time.

Fixed:
image

Sample all day event:
```
{
	"11": {
		"title": "Test All Day",
		"start": "2024-05-10T00:00:00",
		"end": "2024-05-11T00:00:00",
		"all_day": true,
		"tentative": false,
		"description": "",
		"calendar_title": "<redacted>",
		"calendar_color": "#9fe1e7",
		"location": null,
		"uuid": null
	}
}
```

### DIFF
--- a/backend/src/appointment/controller/calendar.py
+++ b/backend/src/appointment/controller/calendar.py
@@ -323,13 +323,15 @@ class CalDavConnector(BaseConnector):
             start = e.vobject_instance.vevent.dtstart.value
             # get_duration grabs either end or duration into a timedelta
             end = start + e.get_duration()
+            # if start doesn't hold time information (no datetime), it's a whole day
+            all_day = not isinstance(start, datetime)
 
             events.append(
                 schemas.Event(
                     title=title,
                     start=start,
                     end=end,
-                    all_day=not isinstance(start, datetime),
+                    all_day=all_day,
                     tentative=tentative,
                     description=e.icalendar_component["description"] if "description" in e.icalendar_component else "",
                 )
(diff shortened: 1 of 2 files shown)


devmount (commented):
Good catch! Works for Google, but not for CalDAV it seems?

image

MelissaAutumn:
Do you have a sample dump of what that calendar event looks like in ical?

Fastmail seems to use Duration instead of setting an end time (which we don't actually handle, so that fix is coming up!)

devmount:
This is the dump:

```ical
BEGIN:VEVENT
CREATED:20240513T080055Z
LAST-MODIFIED:20240513T080115Z
DTSTAMP:20240513T080115Z
UID:cg9ce9g7-arhe-o5li-47jh-2rqa03fyca1j
SUMMARY:AllDayTest
DTSTART;VALUE=DATE:20240514
DTEND;VALUE=DATE:20240515
TRANSP:OPAQUE
CLASS:PUBLIC
END:VEVENT
```

I used some random public test caldav server, so this might just be a weird implementation on server side. They obviously use all day events with excluded `DTEND`.

devmount:
I was able to reproduce this bug with a NextCloud CalDAV calendar. Working on a fix.

devmount (approved):
Ok I don't know what I did wrong, but it suddenly works for all my CalDAV calendars 🙈 I guess my frontend container was in a weird state. When checking out that branch again and restarting the containers, it works now 😇 

So your fix totally works for CalDAV too, sorry for this!

PR #417, opened 2024-05-21
if $REDIS_USE_CLUSTER = True then connect to a redis cluster

### LINKED ISSUE #414: Redis Error on Stage: CROSSSLOT Keys in request don't hash to the same slot
`CROSSSLOT Keys in request don't hash to the same slot`

We currently get that error when attempting to book an appointment. We're blanking out the redis details for now but we should investigate this.

MelissaAutumn:
Should fix #414

### DIFF
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -99,6 +99,8 @@ REDIS_DB=0
 # No value = Python None
 REDIS_PASSWORD
 REDIS_USE_SSL=False
+# Connect to a redis cluster instead of a single instance
+REDIS_USE_CLUSTER=False
 
 # In minutes, the time a cached remote event will expire at.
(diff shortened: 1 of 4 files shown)


devmount (approved):
I cannot test this, but looks reasonable.

A quick question for redis: Are we affected by [their licensing change](https://redis.io/blog/redis-adopts-dual-source-available-licensing/) in any way?

Sancus:
> A quick question for redis: Are we affected by [their licensing change](https://redis.io/blog/redis-adopts-dual-source-available-licensing/) in any way?

No, we used AWS managed redis. I expect them to switch to Valkey when they roll out versions where that's relevant, currently their max supported is 7.1 which uses the old license.

PR #418, opened 2024-05-21
Don't use db models for emails

### LINKED ISSUE #405: Don't pass in models for email templates
For email templates/email background tasks we should create schemas, or additional parameters instead of relying on a model instance. 

We've had some reported issues where the database potentially timed out, causing a new instance to be created and a error generating from trying to retrieve the subscriber from the new db instance.

MelissaAutumn:
Hopefully fixes #405

### DIFF
--- a/backend/src/appointment/controller/mailer.py
+++ b/backend/src/appointment/controller/mailer.py
@@ -170,7 +170,7 @@ class ZoomMeetingFailedMail(Mailer):
 class ConfirmationMail(Mailer):
     def __init__(self, confirm_url, deny_url, attendee_name, attendee_email, date, *args, **kwargs):
         """init Mailer with confirmation specific defaults"""
-        self.attendee_name = attendee_name.title()
+        self.attendee_name = attendee_name
         self.attendee_email = attendee_email
         self.date = date
         self.confirmUrl = confirm_url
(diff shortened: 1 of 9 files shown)


backend/src/appointment/controller/mailer.py
```diff
@@ -168,10 +168,10 @@ def text(self):
 
 
 class ConfirmationMail(Mailer):
-    def __init__(self, confirm_url, deny_url, attendee, date, *args, **kwargs):
+    def __init__(self, confirm_url, deny_url, attendee_name, attendee_email, date, *args, **kwargs):
         """init Mailer with confirmation specific defaults"""
-        self.attendee = attendee
-        self.attendee.name = self.attendee.name.title()
+        self.attendee_name = attendee_name.title()
```
devmount, line 173: I suppose we don't need the title() call anymore:

```suggestion
        self.attendee_name = attendee_name
```

backend/src/appointment/routes/schedule.py
```diff
@@ -254,10 +254,10 @@ def request_schedule_availability_slot(
     db.commit()
 
     # Sending confirmation email to owner
-    background_tasks.add_task(send_confirmation_email, url=url, attendee=attendee, date=date, to=subscriber.email)
+    background_tasks.add_task(send_confirmation_email, url=url, attendee_name=attendee, date=date, to=subscriber.email)
```
devmount, line 257: We need only the name here:

```suggestion
    background_tasks.add_task(send_confirmation_email, url=url, attendee_name=attendee.name, date=date, to=subscriber.email)
```

devmount (changes requested):
Just two minor things that need to be addressed, everything else looks good! This is a weird bug, thank you for this fix!

Issue #334, opened 2024-03-22
Raise stale test error for any tests that use deprecated routes


MelissaAutumn:
We have a handy deprecated flag for routes, right now it just logs the use of them. The intent is, they might still be used somewhere in the code, and we won't to remove it soon, but may not have the time in a specific ticket or timeframe to do the actual sweep. 

However in prep we can start removing tests relating to those routes which should help us sniff out the remaining uses. 

We basically just need to uncomment this line:
https://github.com/thunderbird/appointment/pull/329/commits/01a378c440adbba3fd769abaf1ad5c64378b868b#diff-a6a06a5597c188ae6a48e3740623436efa8e65a7e2f91f8f6b6c98cd0d182b5cR150

PR #426, opened 2024-05-23
Major clean up + Use attendee names for zoom meeting

### LINKED ISSUE #404: Zoom meeting titles should reference booker name or email
Currently, we provide the schedule name to Zoom to name meetings. This results in every meeting being called something like "Andrei's Bookings".

Instead, we should call it something like "`<display name>`'s meeting with `<booker name or email>`".

MelissaAutumn:
Should fix #404 

Will need to test on stage.

This also includes a bunch of clean up, and then I discovered there weren't tests for this route...so it's a little bigger than the 2 line fix.


backend/src/appointment/routes/api.py
```diff
@@ -437,121 +437,6 @@ def read_public_appointment(slug: str, db: Session = Depends(get_db)):
     )
 
 
-@router.put("/apmt/public/{slug}", response_model=schemas.SlotAttendee, deprecated=True)
```
MelissaAutumn, line 440: Deprecated route, contains duplicate zoom meeting link code so removed it and its tests to avoid confusion.

backend/test/integration/test_schedule.py
```diff
@@ -510,3 +515,129 @@ def bust_cached_events(self, all_calendars = False):
...
+            db.commit()
+
+            availability = schemas.AvailabilitySlotConfirmation(
+                slot_id=slot_id,
+                slot_token=slot.booking_tkn,
+                owner_url=signed_url,
+                confirmed=True
+            ).model_dump()
+
+        response = with_client.put(
```
MelissaAutumn, line 571: Routes have their own db instance so we can't have them overlap, otherwise weird stuff might happen.

backend/src/appointment/database/repo/slot.py
```diff
@@ -35,12 +35,14 @@ def get_by_subscriber(db: Session, subscriber_id: int):
...
     """create new slots for appointment of given id"""
+    return_slots = []
     for slot in slots:
         db_slot = models.Slot(**slot.dict())
         db_slot.appointment_id = appointment_id
         db.add(db_slot)
+        return_slots.append(db_slot)
     db.commit()
-    return slots
+    return return_slots
```
devmount, line 45: This makes a lot more sense than returning the argument, thanks.

backend/src/appointment/routes/api.py
devmount: Yeah this is still from the single appointment feature we dropped. There might still be more around...

devmount (approved):
Great work, thanks!

backend/src/appointment/routes/api.py
MelissaAutumn: There is! but they're marked as deprecated. I need to enable this and remove them and their tests at some point: https://github.com/thunderbird/appointment/issues/334

PR #427, opened 2024-05-23
Features/397 invite codes

### LINKED ISSUE #397: Allow sign up by invite code
Allow a user to use an invite code to bypass the email allow list. There would be an additional field on our login form, if non-empty it will check against the table of invite codes, save it in state and upon subscriber creation, mark it as used and associate it.

MelissaAutumn:
Fixes #397 

Invite Admin Panel:
Screen Shot 2024-05-23 at 15 00 03-fullpage

Login:
Screen Shot 2024-05-23 at 15 00 37-fullpage

This PR re-introduces the invite admin panel, which allows an admin to generate x number of invites, revoke unused invites, and filter on invites. A user can now optionally enter an invite, and if they're a new user it will "use" that invite allowing them to register. Otherwise they'll see an error. 

There is also a navigation bar for the admin panel, and some additions to the data-table component like the ability to conditionally disable buttons.


frontend/src/locales/en.json
```diff
@@ -88,10 +89,13 @@
     "addCalendar": "Add {provider} calendar",
     "addDay": "Add day",
     "addTime": "Add time",
+    "admin-invite-codes-panel": "Invites",
+    "admin-subscriber-panel": "Subscribers",
```
devmount, line 93: A bit odd to switch from camelCase to dash-case, but I guess that's because of the route names, right?

frontend/src/views/admin/InviteCodePanelView.vue
```diff
@@ -0,0 +1,270 @@
...
+    <alert-box
+      @close="pageError = ''"
+      v-if="pageError"
+    >
+      {{ pageError }}
+    </alert-box>
+  </div>
+  <admin-nav>
+
+  </admin-nav>
```
devmount, line 19: Is there supposed to be a child element here or can we use a stand-alone tag?

frontend/src/views/admin/SubscriberPanelView.vue
```diff
@@ -14,6 +14,9 @@
       {{ pageError }}
     </alert-box>
   </div>
+  <admin-nav>
+
+  </admin-nav>
```
devmount, line 19: And same here: Is there supposed to be a child element here or can we use a stand-alone tag?

devmount (approved):
Awesome, Mel! There it is again 😅 

I added another table data type `code`, which preformats text and makes it copiable. I also completed translations and found a few minor things.

frontend/src/locales/en.json
MelissaAutumn: Yep! I should just group those. 🤔

frontend/src/views/admin/InviteCodePanelView.vue
MelissaAutumn: Reformatting / editor shenanigans, I'll clean that up!

frontend/src/locales/en.json
devmount: Good idea, maybe just a section directly for route names...?

MelissaAutumn: Yep, I'll make a ticket for it.

PR #434, opened 2024-05-28
Ensure a trailing slash for CalDAV urls

### LINKED ISSUE #343: Ensure there's a slash at the end of caldav urls
Trying to hook up my fastmail calendar with appointment was a bit of a pain. In the end I needed the calendar url from the Settings -> Calendars -> Export, and a closing `/` at the end, otherwise Fastmail will just 401 us.

So why don't we just add that closing slash if it's not there. We should test a few more caldav providers to ensure this won't break anything.

devmount:
## Description of the Change

This change adds a slash `/` to the end of CalDAV urls, when a CalDAV calendar is added or updated and the given url doesn't already end with one.

I normally test with a public CalDavZIP instance (https://calendar.robur.coop/) but that currently seems to be down. I tested with a NextCloud calendar too and that was working just fine.

## Benefits

Less complaints from picky CalDAV servers enforcing a trailing slash.

## Applicable Issues

Closes #343

### DIFF
--- a/backend/src/appointment/routes/api.py
+++ b/backend/src/appointment/routes/api.py
@@ -135,6 +135,10 @@ def create_my_calendar(
             google_tkn=external_connection.token,
         )
     else:
+        # make sure, the CalDAV url has a trailing slash
+        if calendar.url[-1] != '/':
+            calendar.url += '/'
+
         con = CalDavConnector(
             redis_instance=None,
             url=calendar.url,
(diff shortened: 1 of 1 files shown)


MelissaAutumn (commented):
Thanks that should do it! But can we centralize that logic in the CalDavConnector constructor? I don't think we need to mutate the url for the database, just adjust it in the constructor so our api calls will always include the slash.

PR #440, opened 2024-05-30
Add a preferred email option in account settings (Fixes #408)

### LINKED ISSUE #408: Change default email option
By default, Appointment uses the email associated with your Mozilla account. My Mozilla account is associated with my personal email - and I'd like to keep it that way. But I primarily use Appointment with my work email. To accept an appointment, you have to accept via the automated email, which I would like to change to go to my work email instead of the one affiliated with my Mozilla account.

TLDR: Change what email is used with Appointment without having to change the email associated with your Mozilla account.

MelissaAutumn:
Fixes #408 

This PR adds a dropdown to select the users preferred email with a tooltip explaining what that option does:

image
image

We don't have an email verification system, so I'm only pulling from FXA + Google Connections. So if you connect with your personal google account you can use that email for booking emails, and events. 

To facilitate this functionality I've added a `secondary_email` address to the subscribers table, and a `preferred_email` property to the subscribers model. The `preferred_email` property returns either the `secondary_email` or if that's null, the `email`. For any outgoing email usage you should always use `preferred_email`. `email` should only be used in context of login/auth, and `secondary_email` should only be used during updates relating to the dropdown. 

I believe I've caught all of the outgoing email usage, but feel free to let me know if I've missed anything.


backend/src/appointment/routes/account.py
```diff
@@ -52,3 +54,16 @@ def delete_account(db: Session = Depends(get_db), subscriber: Subscriber = Depen
...
     except AccountDeletionException as e:
         raise HTTPException(status_code=500, detail=e.message)
+
+
+@router.get("/available-emails")
+def get_available_emails(db: Session = Depends(get_db), subscriber: Subscriber = Depends(get_subscriber)):
+    """Return the list of emails they can use within Thunderbird Appointment"""
+    google_connections = get_by_type(db, subscriber_id=subscriber.id, type=ExternalConnectionType.google)
+
+    emails = {subscriber.email, *[connection.name for connection in google_connections]} - {subscriber.preferred_email}
```
MelissaAutumn, line 64: This should ensure there's no duplicates if they already have their preferred email set as their main email, or a google account email.

devmount (approved):
Thank you, works!

PR #441, opened 2024-05-31
Fix schedule event previews

### LINKED ISSUE #350: General Availability toggle hides calendar information and doesn't restore it
**Describe the bug**
Using the toggle just hides the calendar events and doesn't restore it when you toggle it back on.

I honestly don't know what the toggle is supposed to do? Can we remove it?

**To Reproduce**

1. Populate your calendar with events, and save your GA settings
3. Click on the General Availibility toggle from active to deactivated
4. Notice that the events are all gone
5. Click the toggle again to activate it
6. Notice that the events are still gone
7. Reload the page
8. Events are restored

**Expected behavior**
I honestly don't know, lol, other than it should at least be restored when I toggle back to the original state

**Actual behavior**
Events are hidden until page refresh.

**Screenshots**
It's this toggle.
Screenshot 2024-04-03 at 4 44 06 PM

- OS: OS X
- Browser Version [e.g. Firefox 114.0b8 (64-Bit)]: FF

devmount:
## Description of the Change

This change does two things:

- Extending the preview slots out of the currently shown month for days displayed from the previous/next month
- Keeping calendar events when creating/saving/toggling availability

image

## Benefits

More convenient display of elements in the calendar on availability configuration.

## Applicable Issues

Closes #350

### DIFF
--- a/frontend/src/views/ScheduleView.vue
+++ b/frontend/src/views/ScheduleView.vue
@@ -87,7 +87,6 @@ const schedules = ref([]);
 const firstSchedule = computed(() => (schedules.value?.length > 0 ? schedules.value[0] : null));
 const schedulesReady = ref(false);
 const getFirstSchedule = async () => {
-  calendarEvents.value = [];
   // trailing slash to prevent fast api redirect which doesn't work great on our container setup
   const { data } = await call('schedule/').get().json();
   schedules.value = data.value;
--- a/frontend/src/views/ScheduleView.vue
+++ b/frontend/src/views/ScheduleView.vue
@@ -107,6 +107,8 @@ const onDateChange = (dateObj) => {
   const start = dj(dateObj.start);
   const end = dj(dateObj.end);
 
+  activeDate.value = start.add(end.diff(start, 'minutes')/2, 'minutes');
+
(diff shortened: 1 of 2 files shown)


frontend/src/views/ScheduleView.vue
```diff
@@ -87,7 +87,6 @@ const schedules = ref([]);
 const firstSchedule = computed(() => (schedules.value?.length > 0 ? schedules.value[0] : null));
 const schedulesReady = ref(false);
 const getFirstSchedule = async () => {
-  calendarEvents.value = [];
```
devmount, line 90: I didn't find any reason to keep this here. This was the cause for calendar events to disappear on schedule save.

frontend/src/views/ScheduleView.vue
```diff
@@ -107,6 +106,8 @@ const onDateChange = (dateObj) => {
   const start = dj(dateObj.start);
   const end = dj(dateObj.end);
 
+  activeDate.value = start.add(end.diff(start, 'minutes')/2, 'minutes');
```
devmount, line 109: We're now updating the selected date on navigation too, to keep the event generation in sync with the view

MelissaAutumn (approved):
Thanks, tested and it works great!

Issue #292, opened 2024-03-11
Basic admin UI for managing accounts


malini:
**Is your feature request related to a problem? Please describe.**
We need a tool to manage our users' accounts so we can allow for banning accounts, changing emails, etc.

**Describe the solution you'd like**

- [x] Account disable + enable (for us as admin ie: banning accounts)
- [x] Generating invite codes for TBA alpha
- [ ] Changing emails (and resetting FxA token)
- [ ] Changing subscription data(plan, etc)
- changing linked calendars (maybe)

devmount:
@malini Follow-up questions:

1. Changing emails means, an admin can replace the current email-address of a subscriber with another by invalidating their session?
2. What's included in "subscription data" exactly?

Thanks!

malini:
closing this since it will be managed as part of tb-accounts

PR #443, opened 2024-05-31
Soft delete / disabling / enabling subscribers via panel

devmount:
## Description of the Change

This change extends the subscribers model by an active flag and the subscribers panel with buttons for disabling / enabling them.

image

Also introduces a new data table type `bool` for binary fields.

I added checks for active flag to the login flow, but there must be more places to check for that (maybe even as middleware? I dont know if necessary). @MelissaAutumn Could you add that if needed?

## Benefits

Control over subscribers to quickly ban or reenable them.

## Applicable Issues

Implements a part of #292

### DIFF
--- a/backend/src/appointment/database/models.py
+++ b/backend/src/appointment/database/models.py
@@ -98,15 +98,23 @@ class Base:
     time_updated = Column(DateTime, server_default=func.now(), default=func.now(), onupdate=func.now(), index=True)
 
 
-class Subscriber(Base):
+class HasSoftDelete:
+    """Mixing in a column to support deletion without removing the record"""
+    time_deleted = Column(DateTime, nullable=True)
+
+    @property
+    def is_deleted(self):
+        """A record is marked deleted if a delete time is set."""
+        return self.time_deleted is not None
+
+
+class Subscriber(HasSoftDelete, Base):
     __tablename__ = "subscribers"
 
     id = Column(Integer, primary_key=True, index=True)
     username = Column(StringEncryptedType(String, secret, AesEngine, "pkcs5", length=255), unique=True, index=True)
     # Encrypted (here) and hashed (by the associated hashing functions in routes/auth)
     password = Column(StringEncryptedType(String, secret, AesEngine, "pkcs5", length=255), index=False)
-
-    active: bool = Column(Boolean, index=True, default=True)
 
     # Use subscriber.preferred_email for any email, or other user-facing presence.
     email = Column(StringEncryptedType(String, secret, AesEngine, "pkcs5", length=255), unique=True, index=True)
--- a/backend/src/appointment/database/models.py
+++ b/backend/src/appointment/database/models.py
@@ -106,6 +106,8 @@ class Subscriber(Base):
     # Encrypted (here) and hashed (by the associated hashing functions in routes/auth)
     password = Column(StringEncryptedType(String, secret, AesEngine, "pkcs5", length=255), index=False)
 
+    active: bool = Column(Boolean, index=True, default=True)
+
     # Use subscriber.preferred_email for any email, or other user-facing presence.
     email = Column(StringEncryptedType(String, secret, AesEngine, "pkcs5", length=255), unique=True, index=True)
     secondary_email = Column(StringEncryptedType(String, secret, AesEngine, "pkcs5", length=255), nullable=True, index=True)
--- a/backend/src/appointment/database/repo/subscriber.py
+++ b/backend/src/appointment/database/repo/subscriber.py
@@ -71,8 +71,24 @@ def update(db: Session, data: schemas.SubscriberIn, subscriber_id: int):
     return db_subscriber
 
 
+def disable(db: Session, subscriber: models.Subscriber):
+    """Disable a given subscriber"""
+    subscriber.active = False
+    db.commit()
+    db.refresh(subscriber)
+    return subscriber
+
+
+def enable(db: Session, subscriber: models.Subscriber):
+    """Enable a given subscriber"""
+    subscriber.active = True
+    db.commit()
+    db.refresh(subscriber)
+    return subscriber
+
+
 def delete(db: Session, subscriber: models.Subscriber):
-    """Delete a subscriber by subscriber id"""
+    """Delete a given subscriber"""
     db.delete(subscriber)
     db.commit()
     return True
--- a/backend/src/appointment/database/repo/subscriber.py
+++ b/backend/src/appointment/database/repo/subscriber.py
@@ -77,7 +77,7 @@ def disable(db: Session, subscriber: models.Subscriber):
     # mark subscriber deleted = disable them
     subscriber.time_deleted = datetime.datetime.now(datetime.UTC)
     # invalidate session
-    # subscriber.minimum_valid_iat_time = datetime.datetime.now(datetime.UTC)
+    subscriber.minimum_valid_iat_time = datetime.datetime.now(datetime.UTC)
     db.add(subscriber)
     db.commit()
     db.refresh(subscriber)
@@ -89,15 +89,15 @@ def enable(db: Session, subscriber: models.Subscriber):
     # mark subscriber deleted = disable them
     subscriber.time_deleted = None
     # refresh session
-    # subscriber.minimum_valid_iat_time = datetime.datetime.now(datetime.UTC)
+    subscriber.minimum_valid_iat_time = datetime.datetime.now(datetime.UTC)
     db.add(subscriber)
     db.commit()
     db.refresh(subscriber)
     return subscriber
 
 
-def delete(db: Session, subscriber: models.Subscriber):
-    """Delete a given subscriber"""
+def hard_delete(db: Session, subscriber: models.Subscriber):
+    """Delete a given subscriber by actually removing their record"""
     db.delete(subscriber)
     db.commit()
     return True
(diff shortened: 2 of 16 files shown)


backend/src/appointment/database/repo/subscriber.py
```diff
@@ -71,8 +71,24 @@ def update(db: Session, data: schemas.SubscriberIn, subscriber_id: int):
     return db_subscriber
 
 
+def disable(db: Session, subscriber: models.Subscriber):
+    """Disable a given subscriber"""
+    subscriber.active = False
```
MelissaAutumn, line 76: You'll want to set the minimum_iat_time to now as well. This will cause any jwt tokens with an iat (issued at) of less than the minimum_iat_time to be invalidated. Thus logging them out.

backend/src/appointment/database/repo/subscriber.py
```diff
@@ -71,8 +71,24 @@ def update(db: Session, data: schemas.SubscriberIn, subscriber_id: int):
...
+    db.commit()
+    db.refresh(subscriber)
+    return subscriber
+
+
+def enable(db: Session, subscriber: models.Subscriber):
+    """Enable a given subscriber"""
+    subscriber.active = True
```
MelissaAutumn, line 84: Probably also here, just to refresh the jwt session.

backend/src/appointment/routes/subscriber.py
```diff
@@ -24,12 +24,20 @@ def get_all_subscriber(db: Session = Depends(get_db), _: Subscriber = Depends(ge
 @router.put("/disable/{email}")
 def disable_subscriber(email: str, db: Session = Depends(get_db), _: Subscriber = Depends(get_admin_subscriber)):
     """endpoint to disable a subscriber by email, needs admin permissions"""
-    # TODO: Add status to subscriber, and disable it instead.
-    raise NotImplementedError
+    subscriber = repo.subscriber.get_by_email(db, email)
+    if not subscriber:
```
MelissaAutumn, line 28: Can we prevent the disabling of already disabled subscribers and the subscriber themself? 😄

backend/src/appointment/routes/subscriber.py
```diff
@@ -24,12 +24,20 @@ def get_all_subscriber(db: Session = Depends(get_db), _: Subscriber = Depends(ge
...
+    
+    # Set active flag to false on the subscribers model.
+    return repo.subscriber.disable(db, subscriber)
 
+
+@router.put("/enable/{email}")
+def disable_subscriber(email: str, db: Session = Depends(get_db), _: Subscriber = Depends(get_admin_subscriber)):
+    """endpoint to disable a subscriber by email, needs admin permissions"""
     subscriber = repo.subscriber.get_by_email(db, email)
     if not subscriber:
```
MelissaAutumn, line 39: same here, can't enable an already enabled subscriber

MelissaAutumn:
Thank you! 

Regarding other checks, simple add a check at the end of `get_user_from_token` in `dependencies.auth` and that should do it!

I'd hate to ask this but can you change the active field to a time_deleted field instead, and move it into a mixin so we can sprinkle it into models as needed. (Adding the mixin to subscriber for now is fine.)

And then add a attribute function to that mixin named is_deleted or is_disabled as a boolean check for time_deleted.

For these types of statuses it's always good to have a timestamp :smile:

backend/src/appointment/database/repo/subscriber.py
devmount: Perfect, exactly what I was looking for, thanks!

backend/src/appointment/routes/subscriber.py
devmount: Of course! I added validation for this.

backend/src/appointment/routes/subscriber.py
devmount: Of course! I added validation for this too.

backend/src/appointment/database/repo/subscriber.py
```diff
@@ -71,8 +72,32 @@ def update(db: Session, data: schemas.SubscriberIn, subscriber_id: int):
     return db_subscriber
 
 
+def disable(db: Session, subscriber: models.Subscriber):
+    """Disable a given subscriber"""
+    # mark subscriber deleted = disable them
+    subscriber.time_deleted = datetime.datetime.now(datetime.UTC)
+    # invalidate session
+    # subscriber.minimum_valid_iat_time = datetime.datetime.now(datetime.UTC)
```
devmount, line 80: This is giving me a weird error:

> sqlalchemy.exc.OperationalError: (MySQLdb.OperationalError) (1292, "Incorrect datetime value: 'xSoAjP5Wa5dyM51mj3ik+DSXLXqREIQAid+KUhryiFZ4yP/YPrUTDYoKnpwkoD3b' for column 'minimum_valid_iat_time' at row 1")
> [SQL: UPDATE subscribers SET time_updated=now(), time_deleted=%s, minimum_valid_iat_time=%s WHERE subscribers.id = %s]
> [parameters: (datetime.datetime(2024, 6, 1, 9, 46, 51, 624961, tzinfo=datetime.timezone.utc), 'xSoAjP5Wa5dyM51mj3ik+DSXLXqREIQAid+KUhryiFZ4yP/YPrUTDYoKnpwkoD3b', 7)]
> (Background on this error at: https://sqlalche.me/e/14/e3q8)

Obviously somehow sqlalchemy can't handle the encrypted datetime string? Maybe you have an idea on this @MelissaAutumn ? Commented it out for now.

backend/src/appointment/database/repo/subscriber.py
```diff
@@ -71,8 +72,32 @@ def update(db: Session, data: schemas.SubscriberIn, subscriber_id: int):
...
+    db.refresh(subscriber)
+    return subscriber
+
+
+def enable(db: Session, subscriber: models.Subscriber):
+    """Enable a given subscriber"""
+    # mark subscriber deleted = disable them
+    subscriber.time_deleted = None
+    # refresh session
+    # subscriber.minimum_valid_iat_time = datetime.datetime.now(datetime.UTC)
```
devmount, line 92: Same here.

frontend/src/views/admin/SubscriberPanelView.vue
```diff
@@ -172,7 +178,7 @@ const filters = [
       if (selectedKey === 'all') {
         return mutableDataList;
       }
-      return mutableDataList.filter((data) => data.wasInvited.value.toLowerCase() === selectedKey);
+      return mutableDataList.filter((data) => data.wasInvited.value?.toString().toLowerCase() === selectedKey);
```
devmount, line 181: I fixed this to handle bool and string values at the same time.

backend/src/appointment/database/models.py
```diff
@@ -98,7 +98,17 @@ def get_columns(self) -> list:
...
+    """Mixing in a column to support deletion without removing the record"""
+    time_deleted = Column(DateTime, nullable=True)
+
+    @property
+    def is_deleted(self):
+        """A record is marked deleted if a delete time is set."""
+        return self.time_deleted is not None
+
+
+class Subscriber(HasSoftDelete, Base):
```
devmount, line 111: I added this mixin separately. Now we can give all models for which we want soft delete this class.

devmount (commented):
I added all of your suggestions, Melissa. And please: don't ever feel bad about asking me for improvements! This was great feedback from you and I happily implemented it 🙌🏻

backend/src/appointment/database/repo/subscriber.py
MelissaAutumn: That is very odd. What is the minimum_valid_iat_time field in your database? It should be a VARCHAR(255) via https://github.com/thunderbird/appointment/blob/main/backend/src/appointment/migrations/versions/2024_01_09_1652-ad7cc2de5ff8_add_minimum_valid_iat_time_to_subscribers.py

backend/src/appointment/database/repo/subscriber.py
```diff
@@ -71,8 +72,32 @@ def update(db: Session, data: schemas.SubscriberIn, subscriber_id: int):
...
+    db.add(subscriber)
+    db.commit()
+    db.refresh(subscriber)
+    return subscriber
+
+
 def delete(db: Session, subscriber: models.Subscriber):
-    """Delete a subscriber by subscriber id"""
```
MelissaAutumn, line 75: Can we swap this function to hard_delete to be explicit, I think it's only used in two places.

frontend/src/views/admin/SubscriberPanelView.vue
```diff
@@ -155,11 +161,11 @@ const filters = [
       },
       {
         name: 'Yes',
-        key: 'yes',
+        key: 'true',
       },
       {
         name: 'No',
-        key: 'no',
+        key: 'false',
```
MelissaAutumn, line 168: I built this and I didn't realize it I could it do it this way 😅

MelissaAutumn (approved):
Looks good, aside from the current errors and the rename. 😄

backend/src/appointment/database/repo/subscriber.py
devmount: Oh good point, I'll investigate.

devmount: You were totally right! Somehow I must have missed that migration and my minimum_valid_iat_time is still datetime 😅

backend/src/appointment/database/repo/subscriber.py
devmount: Of course

PR #444, opened 2024-05-31
Prod cleanup

jdbass:
## Description of the Change

Resolved inconsistencies between production IaC and running environment


MelissaAutumn (approved):
I like when code is removed 😄

PR #453, opened 2024-06-04
Ignore the port if it already exists in the host.

MelissaAutumn:
This should help with AWS credential generation. This will only occur during the aws flow and doesn't affect dev.


jdbass (approved):
Looks good to me, seems like it should work as intended

PR #456, opened 2024-06-05
Fix permission check

MelissaAutumn:
We test that the response is not 200, so it will work as-is right now. But it's just a little odd.


devmount (approved):
Looks good to me. Thanks!

PR #457, opened 2024-06-06
Chores/CS and linter fixes backend

devmount:
## Description of the Change

This applies `black` and `ruff` on our current backend. Seems like we didn't do that for quite some time 😅 Let me know, if you want to change any rules here, afaik we're just applying default rules and a few in pyproject.toml

## Benefits

Consistent code base

## Applicable Issues

None

### DIFF
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -73,7 +73,7 @@ line-length = 120
 # Allow unused variables when underscore-prefixed.
 dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
 
-# Assume Python 3.10.
+# TODO: Assume Python 3.10.
 target-version = "py310"
 
 [tool.ruff.mccabe]
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -35,13 +35,8 @@ dependencies = { file = ["requirements.txt"] }
 
 # Ruff
 [tool.ruff]
-# Enable pycodestyle (`E`) and Pyflakes (`F`) codes by default.
-select = ["E", "F"]
-ignore = []
-
-# Allow autofix for all enabled rules (when `--fix`) is provided.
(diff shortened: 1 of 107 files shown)


backend/pyproject.toml
```diff
@@ -73,7 +73,7 @@ line-length = 120
 # Allow unused variables when underscore-prefixed.
 dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
 
-# Assume Python 3.10.
+# TODO: Assume Python 3.10.
```
devmount, line 76: @MelissaAutumn Do we need to update this? I did this PR on Python 3.12 with current versions of black and ruff...

MelissaAutumn: We can at least upgrade this to 3.11. I think 3.12 is fine, but 3.11 is the docker container's target.

devmount: Alright, changed that. Thanks!

MelissaAutumn:
Thanks for the config fixes! Seems like we might not need black anymore since ruff's formatter sells itself as a drop-in replacement. 

We should switch the quote style back to single quotes and try and fix the excessive param exploding when it comes to functions that contain a dictionary or list as its sole argument.

devmount:
I removed black and added a ruff rule to use single quotes (but somehow the don't get applied on fix?). Do you know which rule to add to fix the function parameters? Thanks!

devmount:
Ok for some reason the quotes config in pyproject.toml wasn't taken into account, so I manually ran this now:

```bash
ruff format backend --config "format.quote-style = 'single'"
```

devmount:
I actually tried that too without success, but I can add it to the pyproject.toml nevertheless

devmount:
Ok, I now added the inline quotes rule as well.

MelissaAutumn:
Thanks!

MelissaAutumn (approved):

backend/src/appointment/dependencies/auth.py
```diff
@@ -18,24 +18,28 @@
...
 
     # Token has been expired by us - temp measure to avoid spinning a refresh system, or a deny list for this issue
-    if any([
-        subscriber is None,
-        subscriber.is_deleted,
-        subscriber and subscriber.minimum_valid_iat_time and not iat,
-        subscriber and subscriber.minimum_valid_iat_time and subscriber.minimum_valid_iat_time.timestamp() > int(iat)
-    ]):
+    if any(
+        [
```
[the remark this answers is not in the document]
devmount: I agree, but I couldn't find a rule for that...

PR #482, opened 2024-06-13
Fix uuids always producing the same result:

MelissaAutumn:
* Pass uuid as a callable not a result
* Fix re-create the uuid index to make them unique
* Re-generate all uuids
* Add tests to test uniqueness.

Easy mistake to miss, now fixed and tested.

### DIFF
--- a/backend/src/appointment/database/models.py
+++ b/backend/src/appointment/database/models.py
@@ -179,7 +179,7 @@ class Appointment(Base):
     __tablename__ = 'appointments'
 
     id = Column(Integer, primary_key=True, index=True)
-    uuid = Column(UUIDType(native=False), default=uuid.uuid4(), index=True)
+    uuid = Column(UUIDType(native=False), default=uuid.uuid4, index=True, unique=True)
     calendar_id = Column(Integer, ForeignKey('calendars.id'))
     duration = Column(Integer)
     title = Column(encrypted_type(String))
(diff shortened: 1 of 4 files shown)


backend/src/appointment/database/models.py
```diff
@@ -179,7 +179,7 @@ class Appointment(Base):
     __tablename__ = 'appointments'
 
     id = Column(Integer, primary_key=True, index=True)
-    uuid = Column(UUIDType(native=False), default=uuid.uuid4(), index=True)
+    uuid = Column(UUIDType(native=False), default=uuid.uuid4, index=True, unique=True)
```
devmount, line 182: SQLAlchemy also provides a Uuid type since 2.0, I think using that could prevent issues with default values maybe.
https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.Uuid

devmount (approved):
Great work, also thanks for the tests. Just one question/suggestions.

backend/src/appointment/database/models.py
devmount: Also thanks for adding `unique=True`! How could we possibly miss that 😅

MelissaAutumn: We generally don't need uniqueness because most if not everything is scoped by the subscriber. But I felt like it'd be silly not to unique a uuid.

And I can make a ticket for that in the future!

devmount: Ah right. Thank you!

PR #483, opened 2024-06-13
Add some info on creating database migrations

MelissaAutumn:
Turns out I needed this command on hand...


devmount (approved):
Haha yeah, me too! I added that to my general project notes 😂 But having it in the Readme again is convenient 👍🏻

PR #489, opened 2024-06-13
Features/448 health check v2

MelissaAutumn:
The healthcheck will now return 503's if MySQL or Redis (If configured) are missing. Otherwise you'll get a 200 `Health OK` message. Technically this message is localized so it ensures locale is working as well. Since the default locale is English, it should say "Health OK". No need to check that though.

This also fixes a local config issue with redis, and tests accidentally inheriting "None" values from your main .env.


devmount (approved):
Looks good to me!

Issue #413, opened 2024-05-15
TSify the frontend


malini:
**Is your feature request related to a problem? Please describe.**
We should move to TS before launch to help our code stability (ex: can catch property-related issues at ts-compile time vs runtime)

**Additional context**
Check Send's approach to TS and emulate closely so we make the same architectural choices between the two projects to reduce cognitive load.

MelissaAutumn:
🙏

MelissaAutumn:
High-level steps, mostly in order:
* [ ] TSify the entry point (main.js) and any configuration required for the rest of the application
* [ ] TSify routes
* [ ] TSify stores
* [ ] TSify components
* [ ] TSify view components
* [ ] TSify defintions (this will be mixed in with the other steps)
* [ ] TSify utils

devmount:
I started working on this and resolved the first step. After #492 landed, we can split the next steps and working on typing different parts in parallel.

devmount:
A little more low level steps, to keep track of remaining tasks:

- [ ] `components/` (27 files) 
    - [ ] `bookingView/`
    - [ ] `FTUE/`
    - [ ] *.vue
- [x] `composables/` (2 files)
- [x] `elements/` (33 files)
- [ ] `stores/` (9 files)
    - [ ] ftue-store.js
- [ ] `tbpro/` (13 files)
    - [ ] `elements/`
    - [ ] `icons/`
- [ ] `views/` (19 files)
    - [x] `admin/`
    - [ ] `errors/`
    - [ ] *.vue
- [ ] App.vue
- [ ] definitions
- [x] keys
- [x] main
- [x] models
- [x] router
- [x] utils

PR #492, opened 2024-06-18
Base TypeScript

devmount:
## Description of the Change

- Add TypeScript dev dependency
- Move main.js -> main.ts
- Update all configuration files to *.config.ts
- Create tsconfig.json with defaults from [Send](https://github.com/thunderbird/send-suite/blob/main/frontend/tsconfig.json)
- Fix Sentry deprecated function calls
- Fix some markdown linter issues

## Benefits

Foundation for all types and type migrations that are to come.

## Applicable Issues

Starts working on #413


frontend/src/main.ts
```diff
@@ -58,7 +58,7 @@ const apiUrl = `${protocol}://${import.meta.env.VITE_API_URL}${port}`;
 app.provide('apiUrl', apiUrl);
 app.provide('bookingUrl', `${protocol}://${import.meta.env.VITE_BASE_URL}/appointments/all/`);
 
-const loc = localStorage?.getItem('locale') ?? (navigator.language || navigator.userLanguage);
```
devmount, line 61: I couldn't find any documentation on a `navigator.userLanguage` property, so I removed it 🤷🏻

MelissaAutumn: Apparently it's IE only 😅

MelissaAutumn (approved):
Looks like a good base!

frontend/src/main.ts
devmount: Haha I see, then I guess it's safe to remove it 😅

PR #499, opened 2024-06-21
Stores TypeScript

devmount:
## Description of the Change

This change introduces types for all our db models and pinia store implementations. This is a first version which is meant to be iterated on in future PRs. Also having a `models.ts` is probably fine for now, but we might want to discuss where to store our types long term. I couldn't find anything in the send repo.

The user store is the last store that's missing. Since it's a bit bigger, I'll do that in a separate PR next week.

_Edit:_ No separate PR. Added the user store TS now too.

## Applicable Issues

Implements #413

### DIFF
--- a/backend/src/appointment/routes/account.py
+++ b/backend/src/appointment/routes/account.py
@@ -27,7 +27,7 @@ def get_external_connections(subscriber: Subscriber = Depends(get_subscriber)):
     external_connections = defaultdict(list)
 
     if os.getenv('ZOOM_API_ENABLED'):
-        external_connections['Zoom'] = []
+        external_connections['zoom'] = []
 
     for ec in subscriber.external_connections:
         external_connections[ec.type.name].append(
--- a/frontend/src/components/CalendarQalendar.vue
+++ b/frontend/src/components/CalendarQalendar.vue
@@ -174,7 +174,7 @@ const processCalendarColorScheme = (calendarTitle, calendarColor) => {
  * @param d - DayJS date object
  * @returns {*}
  */
-const applyTimezone = (d) => dj(d).utc(true).tz(dj.tz.guess());
+const applyTimezone = (d) => dj(d).utc().tz(dj.tz.guess());
 
 /**
  * Generates a list of Qalendar friendly event objects from events and appointments props.
--- a/frontend/src/components/CalendarQalendar.vue
+++ b/frontend/src/components/CalendarQalendar.vue
@@ -174,7 +174,7 @@ const processCalendarColorScheme = (calendarTitle, calendarColor) => {
  * @param d - DayJS date object
  * @returns {*}
  */
-const applyTimezone = (d) => dj.utc(d).tz(dj.tz.guess());
+const applyTimezone = (d) => dj(d).utc(true).tz(dj.tz.guess());
 
 /**
  * Generates a list of Qalendar friendly event objects from events and appointments props.
--- a/frontend/src/models.ts
+++ b/frontend/src/models.ts
@@ -97,3 +97,30 @@ export type Schedule = {
   availabilities?: Availability[];
   calendar: Calendar;
 };
+
+export type User = {
+  email: string;
+  preferredEmail: string;
+  level: number;
+  name: string;
+  timezone: string;
+  username: string;
+  signedUrl: string;
+  avatarUrl: string;
+  accessToken: string;
+  scheduleSlugs: string[];
+}
+
+export type Subscriber = {
+  username: string;
+  name: string;
+  email: string;
+  preferred_email: string;
+  level: number;
+  timezone: string;
+  avatar_url: string;
+}
+
+export type Signature = {
+  url: string;
+}
--- a/frontend/src/models.ts
+++ b/frontend/src/models.ts
@@ -47,3 +47,10 @@ export type Appointment = {
   calendar_color: string;
   active: boolean;
 };
+
+export type Calendar = {
+  id: number;
+  connected: boolean;
+  title: string;
+  color: string;
+};
--- a/frontend/src/models.ts
+++ b/frontend/src/models.ts
@@ -1,4 +1,5 @@
 import { Dayjs } from "dayjs";
+import { UseFetchReturn } from '@vueuse/core';
 
 export type Attendee = {
   id: number;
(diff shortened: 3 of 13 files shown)


backend/src/appointment/routes/account.py
```diff
@@ -27,7 +27,7 @@ def get_external_connections(subscriber: Subscriber = Depends(get_subscriber)):
     external_connections = defaultdict(list)
 
     if os.getenv('ZOOM_API_ENABLED'):
-        external_connections['Zoom'] = []
+        external_connections['zoom'] = []
```
devmount, line 30: I hope this won't cause any issues (it didn't in my local environment). Since fxa and google are lowercase, I thought I just change this for convenience.

frontend/src/components/CalendarQalendar.vue
```diff
@@ -174,7 +174,7 @@ const processCalendarColorScheme = (calendarTitle, calendarColor) => {
  * @param d - DayJS date object
  * @returns {*}
  */
-const applyTimezone = (d) => dj.utc(d).tz(dj.tz.guess());
+const applyTimezone = (d) => dj(d).utc().tz(dj.tz.guess());
```
devmount, line 177: According to [the docs](https://day.js.org/docs/en/manipulate/utc) we implemented this wrong 😅 So no clue, how this worked before, but should be fixed now!

frontend/src/models.ts
devmount, line 1: Please throw in you opinions on this (and the type names). This is just meant to be a first collection of types.

frontend/src/stores/alert-store.ts
```diff
@@ -14,25 +14,24 @@ export const useSiteNotificationStore = defineStore('siteNotification', () => {
...
-   * @param {string} checkId notification id to check agains current id
-   * @returns {boolean}
+   * @param checkId notification id to check agains current id
    */
-  const isSame = (checkId) => id.value === checkId;
+  const isSame = (checkId: string) => id.value === checkId;
 
   /**
    * Lock the given id
-   * @param {string} lockId notification id to lock in
```
devmount, line 24: IMO it's distracting to hold the types in the doc comments too, which are properly defined the line below. But please let me know if you disagree on this.

frontend/src/models.ts
```diff
@@ -0,0 +1,144 @@
...
+export type FetchAppointmentList = (url: string) => UseFetchReturn<Appointment[]> & PromiseLike<UseFetchReturn<Appointment[]>>;
+export type FetchCalendarList = (url: string) => UseFetchReturn<Calendar[]> & PromiseLike<UseFetchReturn<Calendar[]>>;
+export type FetchScheduleList = (url: string) => UseFetchReturn<Schedule[]> & PromiseLike<UseFetchReturn<Schedule[]>>;
+export type FetchExternalConnectionCollection = (url: string) => UseFetchReturn<ExternalConnectionCollection> & PromiseLike<UseFetchReturn<ExternalConnectionCollection>>;
+
+export type Error = { error: boolean|string|null };
+export type Token = {
+  access_token: string;
+  token_type: string;
+}
```
devmount, line 144: Maybe this should live in a separate api.ts file or similar.

frontend/src/stores/user-store.ts
```diff
@@ -56,11 +57,10 @@ export const useUserStore = defineStore('user', () => {
 
   /**
    * Retrieve the current signed url and update store
-   * @param {function} fetch preconfigured API fetch function
-   * @return {boolean}
+   * @param {FetchSignature} call preconfigured API fetch function
    */
-  const updateSignedUrl = async (fetch) => {
-    const { error, data: sigData } = await fetch('me/signature').get().json();
+  const updateSignedUrl = async (call: FetchAny): Promise<Error> => {
```
devmount, line 62: I'm not very happy to use `FetchAny`, but since we're piping the call function to different methods, we can't have another type signature here. Or at least I currently don't see how. I kept the jsdoc comment param type with the appropriate type. Occurs several times in this store.

devmount (commented):
I added the users store now too. Please don't hold back any change requests.

MelissaAutumn:
Sorry I'm a bit busy this week, I'll review it hopefully tomorrow.

MelissaAutumn (commented):
Looks good just some comments.

Thanks!

frontend/src/components/CalendarQalendar.vue
devmount: I thought so too! But setting `utc(true)` gives me wrong times (+2, my utc offset) in the calendar... can you test this with your timezone?

backend/src/appointment/routes/account.py
devmount: I did a case sensitive search for `Zoom` over the whole codebase, but couldn't find another place.

frontend/src/stores/booking-view-store.ts
```diff
@@ -1,21 +1,23 @@
...
+  const dj = inject<(date?: ConfigType) => Dayjs>('dayjs');
 
   // States
   const activeView = ref(bookingCalendarViews.loading);
   const activeDate = ref(dj());
   // Data
-  const selectedEvent = ref(null);
-  const appointment = ref(null);
-  const attendee = ref(null);
+  const selectedEvent = ref<Dayjs>(null);
```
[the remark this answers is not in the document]
devmount: We set `"strictNullChecks": false` in tsconfig (which is from send), so currently there is no need to set `Dayjs|null`, or am I getting you wrong?

frontend/src/stores/external-connections-store.ts
```diff
@@ -19,9 +20,9 @@ export const useExternalConnectionsStore = defineStore('externalConnections', ()
 
   /**
    * Get all external connections for current user
-   * @param {function} call preconfigured API fetch function
+   * @param call preconfigured API fetch function
    */
-  const fetch = async (call) => {
+  const fetch = async (call: FetchExternalConnectionCollection) => {
```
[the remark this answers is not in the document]
devmount: Ok, so if I understand correctly, you don't want a type signature at this place for the parameter `call`? Can you give an example of what you mean?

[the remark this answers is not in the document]
MelissaAutumn: I meant to type the request line like:

const reponse: CoolResponse = await call('cool-endpoint');

because technically call doesn't know what its going to return until that point.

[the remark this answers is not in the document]
devmount: Ah I see, yes that's a better idea, I can rewrite that 👍🏻 I wanted to do it as parameter type to tell call what we expect to receive, but since that is indeed variable it's better to type the call. Thanks!

frontend/src/stores/user-store.ts
devmount: This is fixed now since we have response types.

frontend/src/stores/external-connections-store.ts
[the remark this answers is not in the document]
devmount: Added that now!

frontend/src/models.ts
```diff
@@ -128,14 +128,14 @@ export type Signature = {
 
 // Types and aliases used for our custom createFetch API calls and return types
 export type FetchAny = (url: string) => UseFetchReturn<any> & PromiseLike<UseFetchReturn<any>>;
```
MelissaAutumn, line 130: We can probably just call this Fetch now.

MelissaAutumn (approved):
Looks good, thank you!!

PR #501, opened 2024-06-26
Update regex to allow for spaces and plus signs, and remove the hash before booking.

### LINKED ISSUE #500: Mobile booking error
**Describe the bug**
Ryan ran into this when using https://apmt.day/malini@thunderbird.net/6e711985/ to book me on **mobile**
Screenshot_20240626_110656_Firefox

Related Sentry error is [here](https://thunderbird.sentry.io/issues/5494594831/)

This worked on desktop, though.

- Android 14 running Firefox Mobile 127.0

MelissaAutumn:
Fixes #500 

This only affects you if you dig into week or day view, but we handle it by removing the hash from the url before we send it up. So this applies it in the booking function.

Also retrieve_user_url_data basically takes `/mel/a6b24d4c/` and splits it into (`mel`, `a6b24d4c`, `/mel/a6b24d4c/`). I'm starting to think we should just split by slash, but that can be another PR.

### DIFF
--- a/frontend/src/views/BookingView.vue
+++ b/frontend/src/views/BookingView.vue
@@ -183,9 +183,10 @@ const bookEvent = async (attendeeData) => {
     attendee: attendeeData,
   };
 
+  const url = window.location.href.split('#')[0];
   const request = call('schedule/public/availability/request').put({
     s_a: obj,
-    url: window.location.href,
+    url,
   });
 
   // Data should just be true here.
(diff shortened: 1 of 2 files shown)


frontend/src/views/BookingView.vue
```diff
@@ -183,9 +183,10 @@ const bookEvent = async (attendeeData) => {
     attendee: attendeeData,
   };
 
+  const url = window.location.href.split('#')[0];
```
malini, line 186: Are we guaranteed to have a '#' in the href here?

malini: submitted too early. What happens if we get an error here?

MelissaAutumn: split will still return an array with the url. if theres a hash it will be length==2 instead of 1.

malini: oh whoops my bad, thanks!

malini (approved):

PR #502, opened 2024-06-26
First Time User Experience

MelissaAutumn:
Wow it's done! I just had to battle with tailwind a bit.

This is a huge PR as it includes new components, fonts, assets based off the newly established design language for pro services. As well as slow switch away from tailwind. 

I've looked at this too long, so I'm sure there's some obvious inconsistencies I've missed. Please point them out!

Edit: Added some mobile screenshots. The ui is passable on mobile, but will need some design love in the future.

So here's some notes:

### TBPro
* Added BaseButton and its states to be used as a `button` would.
* Added PrimaryButton and SecondaryButton aliases
* Added TextInput and its states to be used as a `input` would.
* Added SelectInput and its states to be used as a `select` would.
* Added SyncCard which is a fancy checkbox list.
* Added Tooltip which is a WIP because it turns out I'm bad at coding tooltips.
* Added NoticeBar for general information with various states.
* Added Metropolis and Inter fonts.
* Added css variables to colors and sizing.
* Added BubbleSelect component.

### Additions
* Added ftue_level to Subscribers. So we can potentially run them through a FTUE wizard after major changes in the future. For now when the FTUE is done its set to 1.
* Added post-css's custom media plugin to variable-ize media query breakpoints.
* Force FTUE to be in light mode (as dark mode isn't ready yet.)
* Added FTUE with a store handling routing.

### Fixes
* The "My Link" is now properly url-encoded.
* Added some field validation to Schedule and Subscriber schemas.

Screenshots:
Screen Shot 2024-06-26 at 12 47 02-fullpage
Screen Shot 2024-06-26 at 12 47 07-fullpage
Screen Shot 2024-06-26 at 12 48 13-fullpage
Screen Shot 2024-06-26 at 12 48 45-fullpage
Screen Shot 2024-06-26 at 12 48 56-fullpage
Screen Shot 2024-06-26 at 12 49 02-fullpage

Mobile screenshots:

Screen Shot 2024-06-26 at 13 31 22-fullpage
Screen Shot 2024-06-26 at 13 31 26-fullpage
Screen Shot 2024-06-26 at 13 32 11-fullpage
Screen Shot 2024-06-26 at 13 32 17-fullpage
Screen Shot 2024-06-26 at 13 32 21-fullpage
Screen Shot 2024-06-26 at 13 32 25-fullpage


backend/src/appointment/migrations/versions/2024_06_25_2133-156b3b0d77b9_add_ftue_level_to_subscribers.py
```diff
@@ -0,0 +1,24 @@
...
+Create Date: 2024-06-25 21:33:27.094632
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '156b3b0d77b9'
+down_revision = '12c7e1b34dd6'
```
devmount, line 14: Unfortunately this collides with previous developments 😬 
We had two migrations on 2024-06-13 which change the current migration head. I'll fix the head locally on my system, but we should provide a continuous migration queue.

devmount (commented):
First of all: Thank you for this great work! It does not only implement FTUE, but also provides the base for the upcoming changes like the transition away from Tailwind 👏🏻 I feel like we should start documentation on our coding conventions (like which order we use for .vue file sections - script/template/css, how CSS is structure etc.).

But now for the review: I found a lot of small stuff, let me know if I should fix them right away. There is one issue with db migrations, and I currently have issues running the frontend container. I get the following error:

image

Do I need to rebuild the whole container from scratch?

backend/src/appointment/routes/api.py
```diff
@@ -1,3 +1,4 @@
+import logging
```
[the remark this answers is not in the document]
MelissaAutumn: There's a log statement in api.py that was missing this import

frontend/tailwind.config.js
```diff
@@ -1,4 +1,5 @@
 const defaultTheme = require('tailwindcss/defaultTheme');
+const plugin = require('tailwindcss/plugin.js');
```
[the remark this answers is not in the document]
MelissaAutumn: Nothing anymore :scream:

backend/src/appointment/migrations/versions/2024_06_25_2133-156b3b0d77b9_add_ftue_level_to_subscribers.py
MelissaAutumn: Yeah I haven't had the chance to rebase this off of main yet. I'll create a merge migration.

I'm not a fan of git-based database migrations for this reason :sweat_smile:

MelissaAutumn:
For the build stuff you'll have to blow away your node_module volume and rebuild the container. I'm not sure how other projects handle dockerized node_modules. :thinking:

backend/src/appointment/routes/api.py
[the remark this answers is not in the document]
devmount: Oh my bad! Thanks!

backend/src/appointment/migrations/versions/2024_06_25_2133-156b3b0d77b9_add_ftue_level_to_subscribers.py
devmount: Yeah, this is a pain during development... Thanks for a merge migration 🙏🏻

frontend/src/locales/en.json
```diff
@@ -120,6 +120,7 @@
     "busy": "Busy",
     "caldav": "CalDAV",
     "calendar": "Calendar",
+    "calendars": "Calendars",
```
devmount, line 123: Should we rather use i18n plural functionality here?

devmount (approved):
FTUE works on my end now! I love the design! Can't wait to implement it for the whole application 🙏🏻 

Only thing I found: After logging in, I have a bug causing input fields to be dark, but everything else light. This was fixed be reloading the page, so already having some local storage set might be the reason here.
    image

Also I added German translations.

frontend/src/tbpro/elements/BaseButton.vue
```diff
@@ -0,0 +1,147 @@
+<template>
+  <button :class="{'primary': type === 'primary', 'secondary': type === 'secondary', 'small': size === 'small'}" type="button">
```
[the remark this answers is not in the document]
devmount: Alright, makes sense!

frontend/src/locales/en.json
MelissaAutumn: Good catch!

MelissaAutumn:
@devmount I added a new string!

frontend/src/stores/external-connections-store.ts
```diff
@@ -1,6 +1,7 @@
-import { ExternalConnection, ExternalConnectionCollection, Fetch, ExternalConnectionCollectionResponse } from '@/models';
-import { defineStore } from 'pinia';
-import { ref, computed } from 'vue';
+import {ExternalConnection, ExternalConnectionCollection, Fetch, ExternalConnectionCollectionResponse} from '@/models';
+import {defineStore} from 'pinia';
+import {ref, computed} from 'vue';
```
devmount, line 3: Is this a linter thing to remove spaces around curly braces?

frontend/src/stores/schedule-store.ts
```diff
@@ -33,7 +37,7 @@ export const useScheduleStore = defineStore('schedules', () => {
     isLoaded.value = true;
 
     const { updateScheduleUrls } = useUserStore();
-    updateScheduleUrls(data.value);
+    await updateScheduleUrls(data.value);
```
devmount, line 40: > 'await' has no effect on the type of this expression. - ts(80007)

You cannot await a method that is not async.

frontend/src/stores/schedule-store.ts
```diff
@@ -1,10 +1,14 @@
...
 import { defineStore } from 'pinia';
-import { ref, computed } from 'vue';
+import { ref, computed, inject } from 'vue';
 import { useUserStore } from '@/stores/user-store';
+import { dateFormatStrings } from '@/definitions.js';
 import { Fetch, Schedule, ScheduleListResponse } from '@/models';
 
 // eslint-disable-next-line import/prefer-default-export
 export const useScheduleStore = defineStore('schedules', () => {
+  const dj = inject('dayjs');
```
devmount, line 10: This needs to be typed, otherwise TS won't know that dj is a Dayjs callable

```suggestion
  const dj = inject<(date?: ConfigType) => Dayjs>('dayjs');
```

with

```js
import { Dayjs, ConfigType } from 'dayjs';
```

devmount (commented):
Thanks for merging the TS changes in 👏🏻  Left some comments about possible issues I found.

MelissaAutumn:
Couple of things:
* Either I broke the eslint config or it wasn't setup for typescript, so I fixed that
* That then broke my ide's "Run eslint on save" functionality, oops
* I fixed that, and then got annoyed at how we type injections
* Switched dayjs to a typed Symbol
* Got annoyed it wasn't providing timezone
* Added a new type that's an instance of Dayjs class with tz mixed in maybe?
* idk it works now, we should maybe switch off dayjs in the future

Works for me, but I'll merge this onto stage and we can give it a good test.

PR #504, opened 2024-06-27
Upgrade deps

MelissaAutumn:
I ran into an issue, decided it was time to go through and upgrade some ancient deps...in the end the problem wasn't related to the deps, but now we're all nice and upgraded.

I've added a minimum node version (the latest LTS iirc), and all tests pass which is nice. We're still on Python 3.11, but I believe we could jump to 3.12 without any issue now.


frontend/src/views/BookingView.vue
```diff
@@ -138,7 +138,7 @@ const handleError = (data) => {
 
   if (data?.detail?.id === 'SCHEDULE_NOT_ACTIVE') {
     errorHeading.value = '';
-    errorBody.value = data.value.detail.message;
+    errorBody.value = data?.detail.message;
```
MelissaAutumn, line 141: This is just a legitimate error I had to fix while testing.

devmount: Good catch

devmount (approved):
Perfect, thanks!

PR #510, opened 2024-07-03
Update IaC with new cloudfront function

MelissaAutumn:
This fixes the weird validation error AWS randomly started giving us.


devmount (approved):
This just makes sure to always have a base url slash, right?

PR #512, opened 2024-07-03
Don't raise exception if we have an invite code

### LINKED ISSUE #511: When prompted with invite code the login screen complains about allow list
image

MelissaAutumn:
Fixes #511 

The first check is so we can show a nice error on the frontend, there's a second check later on. We should clean this up, but i'm waiting for a design refresh first. So here's the fix.


devmount (approved):
Oh yes, makes totally sense here 👍🏻 We might want to hook l10n in here, but might as well do that later.

PR #518, opened 2024-07-04
add some background to the CloudFront function fix

malini:
## Description of the Change

Just adding some more context if we ever have to look at this function again from the AWS console. It's not immediately clear why we add a slash

### DIFF
--- a/tofu/modules/services/frontend-infra/main.tf
+++ b/tofu/modules/services/frontend-infra/main.tf
@@ -209,7 +209,8 @@ resource "aws_cloudfront_function" "rewrite" {
       request.uri = '/index.html';
     }
 
-    // If empty, then add a slash! Required by AWS, see https://github.com/thunderbird/appointment/pull/510/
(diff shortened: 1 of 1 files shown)


tofu/modules/services/frontend-infra/main.tf
```diff
@@ -209,7 +209,7 @@ resource "aws_cloudfront_function" "rewrite" {
       request.uri = '/index.html';
     }
 
-    // If empty, then add a slash!
+    // If empty, then add a slash! Required by AWS, see https://github.com/thunderbird/appointment/pull/510/
```
MelissaAutumn, line 212:
```suggestion
    // If empty, then add a slash!
    // Required by AWS, see https://github.com/thunderbird/appointment/pull/510/
```

MelissaAutumn (approved):
It felt a little toooooooo long.

devmount (approved):

PR #519, opened 2024-07-04
Schedule booking confirmation

### LINKED ISSUE #460: "No Confirmation Needed" or "Confirmation Needed"

devmount:
## Description of the Change

This change adds a switch to toggle off booking confirmation.

image

## Benefits

If users don't need the additional layer of safety from flooding their calendar with events, they can make the booking process for potential attendees much faster.

## Applicable Issues

Closes #460

### DIFF
--- a/backend/src/appointment/database/models.py
+++ b/backend/src/appointment/database/models.py
@@ -269,7 +269,7 @@ class Schedule(Base):
     farthest_booking: int = Column(Integer, default=20160)  # in minutes, defaults to 2 weeks
     weekdays: str | dict = Column(JSON, default='[1,2,3,4,5]')  # list of ISO weekdays, Mo-Su => 1-7
     slot_duration: int = Column(Integer, default=30)  # defaults to 30 minutes
-    booking_confirmation: bool = Column(Boolean, index=True, default=True)
+    booking_confirmation: bool = Column(Boolean, index=True, nullable=False, default=True)
 
     # What (if any) meeting link will we generate once the meeting is booked
     meeting_link_provider: MeetingLinkProviderType = Column(
--- a/backend/src/appointment/database/models.py
+++ b/backend/src/appointment/database/models.py
@@ -269,6 +269,7 @@ class Schedule(Base):
     farthest_booking: int = Column(Integer, default=20160)  # in minutes, defaults to 2 weeks
     weekdays: str | dict = Column(JSON, default='[1,2,3,4,5]')  # list of ISO weekdays, Mo-Su => 1-7
     slot_duration: int = Column(Integer, default=30)  # defaults to 30 minutes
+    booking_confirmation: bool = Column(Boolean, index=True, default=True)
 
     # What (if any) meeting link will we generate once the meeting is booked
     meeting_link_provider: MeetingLinkProviderType = Column(
--- a/backend/src/appointment/database/schemas.py
+++ b/backend/src/appointment/database/schemas.py
@@ -129,6 +129,7 @@ class AppointmentOut(AppointmentBase):
     owner_name: str | None = None
     slots: list[SlotBase | SlotOut] = []
     slot_duration: int
+    booking_confirmation: bool
 
 
 """ SCHEDULE model schemas
--- a/backend/src/appointment/database/schemas.py
+++ b/backend/src/appointment/database/schemas.py
@@ -170,6 +170,7 @@ class ScheduleBase(BaseModel):
     weekdays: list[int] | None = [1, 2, 3, 4, 5]
     slot_duration: int | None = None
     meeting_link_provider: MeetingLinkProviderType | None = MeetingLinkProviderType.none
+    booking_confirmation: bool | None = True
 
     class Config:
         json_encoders = {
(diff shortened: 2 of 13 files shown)


frontend/src/locales/de.json
devmount, line 1: I added some missing German translations and (sorry) sorted json keys again 😅

frontend/src/locales/en.json
```diff
@@ -317,6 +358,7 @@
       "continueToFxa": "Enter your email above to continue to Mozilla Accounts"
     },
     "nameIsInvitingYou": "{name} is inviting you",
+    "ownerNeedsToConfirmBooking": "If this option is active, appointments are only booked until the calendar owner confirmed it.",
```
devmount, line 361: Please correct my bad English if necessary. In this context it might be even better to say "until you confirmed them", because it's displayed only for the one editing their schedule.

backend/src/appointment/database/models.py
```diff
@@ -269,6 +269,7 @@ class Schedule(Base):
     farthest_booking: int = Column(Integer, default=20160)  # in minutes, defaults to 2 weeks
     weekdays: str | dict = Column(JSON, default='[1,2,3,4,5]')  # list of ISO weekdays, Mo-Su => 1-7
     slot_duration: int = Column(Integer, default=30)  # defaults to 30 minutes
+    booking_confirmation: bool = Column(Boolean, index=True, default=True)
```
MelissaAutumn, line 272: Can we make nullable=False, I'd like to start using that on all fields that make sense.

backend/src/appointment/migrations/versions/2024_07_04_1501-fb1feb76c467_schedule_booking_confirmation.py
```diff
@@ -0,0 +1,24 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'fb1feb76c467'
+down_revision = '0c22678e25db'
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.add_column('schedules', sa.Column('booking_confirmation', sa.Boolean, default=True, index=True))
```
MelissaAutumn, line 20: Also nullable=False here, please!

backend/src/appointment/database/schemas.py
```diff
@@ -170,6 +171,7 @@ class ScheduleBase(BaseModel):
     weekdays: list[int] | None = [1, 2, 3, 4, 5]
     slot_duration: int | None = None
     meeting_link_provider: MeetingLinkProviderType | None = MeetingLinkProviderType.none
+    booking_confirmation: bool | None = True
```
MelissaAutumn, line 174: Once nullable = False you can remove the | None.

frontend/src/locales/en.json
MelissaAutumn, line 361:
```suggestion
    "ownerNeedsToConfirmBooking": "When this option is active, you will be required to confirm or decline times booked on your schedule. When this option is deactivated, all times will automatically be confirmed.",
```
I almost think this should show two information text depending on the state of the button. But I suggested both here.

MelissaAutumn: Either way it's their calendar, so we should be referring to them not the calendar owner.

backend/src/appointment/routes/schedule.py
```diff
@@ -185,6 +185,7 @@ def read_schedule_availabilities(
         owner_name=subscriber.name,
         slots=actual_slots,
         slot_duration=schedule.slot_duration,
+        booking_confirmation=schedule.booking_confirmation
```
MelissaAutumn, line 188: We should really dedicate a schema to this endpoint at somepoint 😄

MelissaAutumn (changes requested):
A few things!

backend/src/appointment/database/models.py
devmount: Of course, good thinking. Especially for all bool fields.

backend/src/appointment/routes/schedule.py
devmount: Yes we should. This is kind of grown from the single-Appointments we used to have

backend/src/appointment/routes/schedule.py
```diff
@@ -494,10 +535,3 @@ def decide_on_schedule_availability_slot(
     slot = repo.slot.book(db, slot.id)
 
     Tools().send_vevent(background_tasks, slot.appointment, slot, subscriber, slot.attendee)
-
```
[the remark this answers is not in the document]
devmount: Oh! Thank you for finding that! Totally oversaw the early return. Yes, a boolean is sufficient here, the schemas are returned by the parent functions 👍🏻

frontend/src/locales/en.json
devmount: Yes thanks. I added your suggestions. If we feel this is too much text, we can conditionally show parts of it.

frontend/src/views/BookingView.vue
```diff
@@ -41,7 +42,7 @@
     <booking-modal
       :open="showBookingModal"
       :event="selectedEvent"
-      :requires-confirmation="true"
+      :requires-confirmation="appointment?.booking_confirmation"
```
MelissaAutumn, line 45: If appointment is empty here I'm sure more things would break, but it also show that the booking was successfully booked. Should we just let it hard crash here to get some sentry information to fix any future bug?

MelissaAutumn (approved):
Looks good thanks!

frontend/src/views/BookingView.vue
devmount: Good point. I removed the `?`.

PR #521, opened 2024-07-05
Ssl cert argument

jdbass:
## Description of the Change

The SSL cert is now supplied via a github variable to allow for more easily updating the cert

## Benefits

Allows for the SSL cert to be updated without code changes


devmount (approved):
Looks good to me, thank you!

PR #527, opened 2024-07-10
Configuration to fix HMR

devmount:
## Description of the Change

Ever since we switched to Vite, my docker container struggles to properly handle hot module reload (HMR), which is a pain for development on my end. I found a configuration for the server section in vite.config.ts which restores HMR for me.

If this doesn't have any known side effects or interference with other configuration, it would improve my dev flow a lot 😇

### DIFF
--- a/frontend/vite.config.ts
+++ b/frontend/vite.config.ts
@@ -1,13 +1,24 @@
 import { sentryVitePlugin } from '@sentry/vite-plugin';
 import { fileURLToPath, URL } from 'node:url';
 
-import { defineConfig, loadEnv } from 'vite';
+import { defineConfig, loadEnv, WatchOptions } from 'vite';
 import vue from '@vitejs/plugin-vue';
 
 // https://vitejs.dev/config/
 export default defineConfig(({ mode }) => {
   // Load env file based on `mode` in the current working directory.
   const env = loadEnv(mode, process.cwd());
+
+  // Build watch options
+  const watch = {
+    usePolling: env.VITE_SERVER_WATCH_POLLING === 'true'
+  } as WatchOptions;
+  if (env.VITE_SERVER_WATCH_INTERVAL) {
+    watch.interval = Number(env.VITE_SERVER_WATCH_INTERVAL);
+  }
+  if (env.VITE_SERVER_WATCH_BINARY_INTERVAL) {
+    watch.binaryInterval = Number(env.VITE_SERVER_WATCH_BINARY_INTERVAL);
+  }
 
   return {
     plugins: [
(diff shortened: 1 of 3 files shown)


MelissaAutumn:
Looks like it may cause high cpu usage on other machines. I'll look into adding a vite-user.config.ts. I was half-way to implementing it before something else came up. But the idea is, it's not attached to git, and if it's there it gets imported mixins custom settings.

MelissaAutumn:
It's mentioned in the vite documentation
https://vitejs.dev/config/server-options#server-watch

devmount:
Update: my bad, `usePolling: true` actually did increase my CPU usage by ~20%. So thanks for pointing that out. I also found that increasing the default polling intervals reverts this effect. E.g. setting this only leads to an average increase of 5% on my end:

```
    watch: {
      usePolling: true,
      interval: 1500,
      binaryInterval: 3000,
    },
```

So yeah, having a user defined vite config which is not git tracked would be the best solution.

devmount:
Ok for now, I introduced a new frontend .env var `VITE_SERVER_WATCH_POLLING`. Polling will only be activated if this is set to true. In the future when we need more user defined configuration, we can easily switch that env to something like VITE_USER_DEFINED_CONFIG and merge a corresponding user config file in.

frontend/vite.config.ts
```diff
@@ -1,34 +1,48 @@
...
 
-  server: {
-    host: '0.0.0.0',
-  },
+    server: {
+      host: '0.0.0.0',
+      hmr: {
+        clientPort: 8080,
+      },
+      port: 8080,
```
MelissaAutumn, line 36: Are these needed for your setup? If not, I'd like to avoid explicitly setting them.

frontend/vite.config.ts
```diff
@@ -1,34 +1,48 @@
...
+      watch: {
+        usePolling: env.VITE_SERVER_WATCH_POLLING === 'true',
+        interval: 1500,
+        binaryInterval: 3000,
```
MelissaAutumn, line 40: Can we make these env vars too? I think the defaults are smaller which work nicely on linux/mac 😄

MelissaAutumn (changes requested):
Sorry for the extreme delay. Just a few more changes and it's good.

frontend/vite.config.ts
devmount: You're right, I don't need them!! I' remove it, thanks!

frontend/vite.config.ts
devmount: Of course, I'm on it

devmount:
No worries! I added two more env vars for intervals and implemented at while keeping defaults

MelissaAutumn (approved):

PR #531, opened 2024-07-10
Bake in the github short sha as the container's release version.

MelissaAutumn:
I believe this should work. Tested the container side of things locally, without providing one it's an empty env variable, when providing something it fills it in correctly.


jdbass (approved):
updated the new reference to gihub_sha to match existing references for consistency within the file

devmount (approved):
Looking good!

PR #533, opened 2024-07-10
Remove hardcoded cert

jdbass:
## Description of the Change

removes hardcoded cert so it can be updated via github variables


MelissaAutumn (approved):
looks good, thanks!

PR #534, opened 2024-07-10
corrected ssl cert variable reference

jdbass:
## Description of the Change

Corrected an error in moving the ssl cert to a variable


MelissaAutumn (approved):
👍

PR #536, opened 2024-07-10
Waiting List Admin Panel (Readonly)

MelissaAutumn:
To get visibility on the waiting list, I've added a read-only admin panel view. 

image

Bit of a rush, we'll add the "Accept x number of peeps" as a later feature.

### DIFF
--- a/frontend/src/views/admin/WaitingListPanelView.vue
+++ b/frontend/src/views/admin/WaitingListPanelView.vue
@@ -1,36 +1,3 @@
-<template>
-  <div class="flex w-full justify-center">
-    <alert-box
-      @close="pageNotification = ''"
-      v-if="pageNotification"
-      :scheme="alertSchemes.success"
-    >
-      {{ pageNotification }}
-    </alert-box>
-    <alert-box
-      @close="pageError = ''"
-      v-if="pageError"
-    >
-      {{ pageError }}
-    </alert-box>
-  </div>
-  <admin-nav/>
-  <div v-if="displayPage">
-    <data-table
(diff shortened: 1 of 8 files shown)


frontend/src/views/admin/WaitingListPanelView.vue
```diff
@@ -0,0 +1,216 @@
...
+      :loading="loading"
+    >
+    </data-table>
+  </div>
+  <div v-else class="flex size-full min-h-[75vh] items-center justify-center">
+    <loading-spinner/>
+  </div>
+</template>
+
+<script setup>
```
devmount, line 34: Do we want to put the script tag at the top for all (new) single file components?

devmount (approved):
Thank you! Just some minor questions.

frontend/src/views/admin/WaitingListPanelView.vue
```diff
@@ -0,0 +1,216 @@
...
+      },
+    ],
+    /**
+     * Callback function, filter the list by selectedKey and return it back to the table
+     * @param selectedKey
+     * @param mutableDataList
+     * @returns {*}
+     */
+    fn: (selectedKey, mutableDataList) => {
+      console.log(mutableDataList)
```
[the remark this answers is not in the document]
MelissaAutumn: whoops, thanks!

frontend/src/views/admin/WaitingListPanelView.vue
MelissaAutumn: Sure! I think that's probably a good plan.

PR #537, opened 2024-07-10
Replace yarn with npm

MelissaAutumn:
This replaces yarn 1.* with whatever version of npm I have installed (10.7.0). 

With our typescriptification of the frontend yarn has started producing weird errors with eslint. I was unable to lint, and upgrading to yarn 4 was a little complicated. So let's just switch to good ol' npm. 

You'll have to nuke the `appointment_node_modules` volume and rebuild the container.


frontend/.eslintrc.cjs
```diff
@@ -6,7 +6,7 @@ module.exports = {
     browser: true,
     es2021: true,
   },
-  parser: '@typescript-eslint/parser',
+  parser: 'vue-eslint-parser',
```
MelissaAutumn, line 9: Additional eslint fixes.

devmount (approved):
Yarn used to be great, but I recently switched back to npm in several projects too 🤷🏻  Thank you.

PR #539, opened 2024-07-11
added recursive flag to frontend zip command

jdbass:
## Description of the Change

added recursive flag to frontend zip command


MelissaAutumn (approved):
Looks good!

PR #549, opened 2024-07-15
Use server_default for migrations, and fix a logic error from my suggestion

MelissaAutumn:
This is what I get for looking at things and going "eh, that checks out." without manually checking it out.

The migration wasn't working with default, so I swapped it to server_default along with their true() function. Otherwise existing schedules would have this set as false 😱 

And fixed an error from my suggestion that we can't really work-around with the existing store setup. So adding those question marks back in!


devmount (approved):
Oh interesting! Might be my bad, I thought I tested this and it worked 🤔 Thank you!

PR #552, opened 2024-07-17
Fix timezone conversion

### LINKED ISSUE #542: Timezone in bookings is not converted
**Describe the bug**
The timezone in the booking page seems to be UTC, not my current timezone

**To Reproduce**

1. Set Europe/Berlin
2. Have someone book a time at 10am Europe/Berlin
3. Go to https://appointment.day/appointments

**Expected behavior**
Time should be converted to my current local timezone, or I guess the primary timezone I have set in the preferences. Or at the very least indicated.

**Actual behavior**
* Time in the email is `Mon Jul 15 10:00:00 2024`
* Time in the list is 8:00 to 8:30am

**Screenshots**
image
image

devmount:
## Description of the Change

We had an issue where times on all appointments from the appointment store were displayed in the wrong timezone, if the timezone between owner and bookee was different. I think I did that wrong last time I touched this 😅 

Also I removed the "Your booking will be confirmed by the calendar owner." hint from the booking model if confirmation was disabled.

## Benefits

Correct times again.

## Applicable Issues

Closes #542

### DIFF
--- a/frontend/test/stores/appointment-store.test.js
+++ b/frontend/test/stores/appointment-store.test.js
@@ -89,6 +89,9 @@ describe('Appointment Store', () => {
   test('timezone', async () => {
     const apmt = useAppointmentStore();
     await apmt.fetch(createFetch({ baseUrl: API_URL }));
+    // TODO This check depends on the timzone set in the user store.
+    // If no user timezone is set it defaults to UTC, which is currently the case.
+    // So we might want to adjust the github workflow runner to set a timezone for testing.
     expect(apmt.appointments[0].slots[0].start.toISOString()).toBe('3000-01-01T09:00:00.000Z');
   });
 
--- a/frontend/test/stores/appointment-store.test.js
+++ b/frontend/test/stores/appointment-store.test.js
@@ -86,6 +86,12 @@ describe('Appointment Store', () => {
     expect(apmt.pendingAppointments.length).toBe(1);
   });
 
+  test('timezone', async () => {
+    const apmt = useAppointmentStore();
+    await apmt.fetch(createFetch({ baseUrl: API_URL }));
+    expect(apmt.appointments[0].slots[0].start.toISOString()).toBe('3000-01-01T09:00:00.000Z');
+  });
+
   test('reset', async () => {
     const apmt = useAppointmentStore();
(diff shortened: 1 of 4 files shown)


MelissaAutumn:
Looks good, can you add a new test for appointmentStore's fetch function that tests the timezone conversion.

devmount:
Of course! Haha sorry always forgetting about the tests 😅

frontend/test/stores/appointment-store.test.js
```diff
@@ -86,6 +86,12 @@ describe('Appointment Store', () => {
     expect(apmt.pendingAppointments.length).toBe(1);
   });
 
+  test('timezone', async () => {
+    const apmt = useAppointmentStore();
+    await apmt.fetch(createFetch({ baseUrl: API_URL }));
+    expect(apmt.appointments[0].slots[0].start.toISOString()).toBe('3000-01-01T09:00:00.000Z');
```
devmount, line 92: Is this sufficient? I think this only tests the fallback to tz.guess which is UTC in the testing environment. Any idea on how to simulate the users timezone here?

devmount (commented):
@MelissaAutumn Added a test, but it might not be that useful (see comment) 😅

frontend/test/stores/appointment-store.test.js
MelissaAutumn: https://day.js.org/docs/en/timezone/set-default-timezone this might work? I haven't looked at the code. If not then maybe we should start looking at other date libraries 🙀

MelissaAutumn: We could also adjust the github workflow runner to set a timezone. That might be a good fallback solution hmmm.

devmount: Setting a default timezone after the dj objects are initialize won't have any effect, so I think your idea with the gh workflow runner timezone is the best/least effort solution. I added a hint in the code. Thank you for your good thoughts on this!

MelissaAutumn: Sounds good, we can probably fix this later!

MelissaAutumn (approved):

PR #556, opened 2024-07-17
Add some initial user metric collection points

### LINKED ISSUE #550: add Sentry metrics to get usage stats
In order to target and prioritize our engineering and design efforts, we need to know more about our users. What types of devices/browsers they're using, etc.

The most important question right now are what devices/os/browser are people using so we can most effectively build our UI experiences for our users.

We should minimally submit (perhaps on page load?) the following to sentry:
- browser type/version
- os type/version
- device type/version
- browser screen size

**Describe alternatives you've considered**
Not much at this point. Sentry's metrics are new, so there might be some challenges that are unknown, but we're going with Sentry since it's already hooked up for us

MelissaAutumn:
Fixes #550 

Looks like: https://thunderbird.sentry.io/metrics/?interval=1h&project=4505428280279040&statsPeriod=7d

I'd like to eventually have some of this data open similar to how we handle https://stats.thunderbird.net. But for now its all bundled in Sentry.

It seems tags are the easiest to query against so I just bundled the info under a `page_load` event. (sentry metrics seem to use underscore casing.)

Note: We need to ensure we stay on ua-parser@^1 and not upgrade to 2 as they have some license changes that may affect us due to MPL2. Works fine for now.


frontend/src/main.ts
```diff
@@ -62,5 +65,34 @@ const loc = localStorage?.getItem('locale') ?? navigator.language;
...
+      browserVersion: `${browser.name}:${browser.version}`,
+      os: os.name,
+      osVersion: `${os.name}:${os.version}`,
+      device: device.model,
+      deviceModel: `${device.vendor}:${device.model}`,
+      resolution: deviceRes,
+      effectiveResolution: effectiveDeviceRes,
+      userAgent: navigator.userAgent,
+      locale: loc,
+      theme: getPreferredTheme(),
```
MelissaAutumn, line 92: values: `light` or `dark`.

frontend/src/main.ts
```diff
@@ -62,5 +65,34 @@ const loc = localStorage?.getItem('locale') ?? navigator.language;
...
+  Sentry.metrics.increment('page_load', 1, {
+    tags: {
+      browser: browser.name,
+      browserVersion: `${browser.name}:${browser.version}`,
+      os: os.name,
+      osVersion: `${os.name}:${os.version}`,
+      device: device.model,
+      deviceModel: `${device.vendor}:${device.model}`,
+      resolution: deviceRes,
+      effectiveResolution: effectiveDeviceRes,
```
MelissaAutumn, line 89: Basically the size of the browser viewport

frontend/src/main.ts
```diff
@@ -62,5 +65,34 @@ const loc = localStorage?.getItem('locale') ?? navigator.language;
...
+
+  Sentry.metrics.increment('page_load', 1, {
+    tags: {
+      browser: browser.name,
+      browserVersion: `${browser.name}:${browser.version}`,
+      os: os.name,
+      osVersion: `${os.name}:${os.version}`,
+      device: device.model,
+      deviceModel: `${device.vendor}:${device.model}`,
+      resolution: deviceRes,
```
MelissaAutumn, line 88: Should be screen size iirc

devmount (approved):
Nice looking good! Thanks Mel

PR #566, opened 2024-07-18
Add slow api rate limiting

### LINKED ISSUE #466: Investigate rate limiting options
We should rate limit the booking url, especially for people hitting multiple different booking urls. Mostly to discourage booking abuse / booking url sniffing.

MelissaAutumn:
Fixes #466 

Rules:
* waiting-list/join - 2/minute
* schedule/public/availability - 10/minute
* schedule/public/availability/request - 10/minute

I think we could reduce the request one to like 2 or 3, but it's not really a concern right now. Mostly just preemptively avoiding spam. I didn't hook it up with redis because I don't think we have redis setup correctly on prod yet. Shouldn't be a problem for us right now though. 

It works off of remote ip address, I think that's probably okay.

### DIFF
--- a/backend/test/integration/test_waiting_list.py
+++ b/backend/test/integration/test_waiting_list.py
@@ -7,6 +7,7 @@ from appointment.database import models
 from unittest.mock import patch
 
 from appointment.exceptions.validation import APIRateLimitExceeded
+from appointment.routes import waiting_list
 from appointment.routes.waiting_list import WaitingListAction
 from appointment.tasks.emails import send_confirm_email, send_invite_account_email
 from defines import auth_headers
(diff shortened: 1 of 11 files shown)


backend/test/integration/test_waiting_list.py
```diff
@@ -49,6 +48,34 @@ def test_already_in_list(self, with_db, with_client, make_waiting_list):
...
+        assert response.json() is True
+
+        # Oh so sneakily signing up a third!
+        email = 'hello2@example.org'
+        response = with_client.post('/waiting-list/join', json={'email': email})
+
+        # Ensure the response was rate limited
+        assert response.status_code == 429, response.json()
+        data = response.json()
+        assert data['detail']['id'] == APIRateLimitExceeded.id_code
```
devmount, line 77: Why does the `@limiter.limit("2/minute")` for this not work here? It returns 200 in tests instead of expected 429?

devmount (commented):
Thank you for this! One question below. I added German translation. Hardcoding the rate limits is fine for now, but we might want to make that configurable in the future e.g. via env maybe?

backend/test/integration/test_waiting_list.py
MelissaAutumn: that's not right. I'll look into this next week.

MelissaAutumn:
This is due to the rate limiter's cache not being cleared between requests. But this also brings up another issue so this PR is on hold.

PR #567, opened 2024-07-18
Switch page load collection to backend.

### LINKED ISSUE #588: Should drop meeting link in "Location" field as well as in description
**Describe the bug**
Currently, a Zoom link associated with an appointment is dropped into the description field. We should make it so that the Zoom link is also attached in the location field.

MelissaAutumn:
Fixes #588

I've kept it a separate endpoint in case folks want to block it (We'd rather they not block it, but that's up to them imo.) It requires sentry to be enabled on front and backend.

View:
https://thunderbird.sentry.io/metrics/?interval=30m&project=4505428124827648&statsPeriod=7d

### DIFF
--- a/backend/requirements.txt
+++ b/backend/requirements.txt
@@ -20,7 +20,7 @@ python-dotenv==1.0.1
 python-multipart==0.0.9
 PyJWT==2.6.0
 pydantic[email]==2.7.4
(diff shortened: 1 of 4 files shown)


devmount (approved):
I like that, thank you Mel!

PR #577, opened 2024-07-23
Components elements TypeScript

devmount:
## Description of the Change

This change adds types for all Vue components in `elements/*` and `views/admin/*`.

- A bug was fixed where data tables showed the paginated instead of the total number of entries
- A bug was fixed where data table top and bottom areas had a wrong background color

## Applicable Issues

Implements #413

### DIFF
--- a/frontend/src/components/CalendarMiniMonth.vue
+++ b/frontend/src/components/CalendarMiniMonth.vue
@@ -53,7 +53,6 @@ import {
   IconChevronLeft,
   IconChevronRight,
 } from '@tabler/icons-vue';
-import { appointmentState } from '@/definitions';
 import { dayjsKey } from "@/keys";
 
 // component constants
@@ -121,10 +120,7 @@ const eventsByDate = (d) => {
     return props.placeholder
       ? events.value[key].filter((e) => dj(e.start).isAfter(dj()))
       : events.value[key].filter(
-        (e) => (dj(e.start).add(e.duration, 'minutes').isAfter(dj()) && e.status === appointmentState.pending)
-            || (e.attendee && e.status === appointmentState.pending)
-            || e.remote
-            || e.preview,
+        (e) => (dj(e.start).add(e.duration, 'minutes').isAfter(dj())) || e.remote || e.preview,
       );
   }
   return [];
--- a/frontend/src/elements/AppointmentGridItem.vue
+++ b/frontend/src/elements/AppointmentGridItem.vue
@@ -1,3 +1,36 @@
+<script setup lang="ts">
+import { bookingStatus } from '@/definitions';
+import { inject, computed } from 'vue';
+import { keyByValue, timeFormat } from '@/utils';
+import { useI18n } from 'vue-i18n';
+import { Appointment } from '@/models';
+
+// icons
+import {
+  IconBulb,
+  IconCalendar,
+  IconClock,
+} from '@tabler/icons-vue';
+import { dayjsKey, paintBackgroundKey } from "@/keys";
+
+// component constants
+const { t } = useI18n();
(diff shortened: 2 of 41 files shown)


frontend/src/components/CalendarMiniMonth.vue
devmount, line 1: I threw away `appointmentState` since this was a relic from the past.

frontend/src/elements/AppointmentGridItem.vue
```diff
@@ -7,7 +40,7 @@
     }"
     :style="{ borderColor: appointment.calendar_color }"
     @mouseover="el => !isPast ? paintBackground(el, appointment.calendar_color, '22') : null"
-    @mouseout="el => !isPast ? paintBackground(el, appointment.calendar_color, _, true) : null"
+    @mouseout="el => !isPast ? paintBackground(el, appointment.calendar_color, '', true) : null"
```
devmount, line 43: Idk how to skip typed parameters, so I gave a (irrelevant) default argument here.

devmount (commented):
This is the second part for typing all components in the elements directory. I already started to set `lang="ts"` on some view components, I will properly type them in a future PR. Just wanted to already use some types without keeping deprecated stuff.

This needs to be tested with FTUE! Or is there a simple way to retrigger the FTUE dialog? I thought there was some flag somewhere 😅

frontend/src/views/admin/InviteCodePanelView.vue
```diff
@@ -231,19 +181,19 @@ const generateInvites = async () => {
...
   const { data, error } = response;
 
   if (error.value) {
-    const errorObj = data.value?.detail;
+    const errorObj = (data.value as Exception)?.detail;
 
     if (!errorObj) {
       pageError.value = t('error.somethingWentWrong');
     } else if (errorObj instanceof Array) {
       pageError.value = errorObj.map((err) => err.msg).join('\n');
```
devmount, line 194: @MelissaAutumn I couldn't figure out which case that is? When will `data.value` hold an array of error objects with `msg` property?

frontend/src/elements/AppointmentGridItem.vue
MelissaAutumn: undefined might work? That's the value if a parameter is skipped by default in js at least.

frontend/src/models.ts
```diff
@@ -157,43 +167,76 @@ export type Subscriber = {
...
+export type Exception = {
+  status_code?: number;
+  detail?: { msg: string }[]|ExceptionDetail;
+  headers?: any[];
+};
+export type ExceptionDetail = {
+  id?: string;
+  message?: string;
+  status?: number;
+}
```
MelissaAutumn, line 200: Since Exception references ExceptionDetail, can you move this above Exception. Thanks!

frontend/src/views/admin/InviteCodePanelView.vue
MelissaAutumn: I believe it's when there's multiple pydantic validation errors at once. I don't remember why this could come up.

MelissaAutumn (approved):
Couple things but looks good! I haven't done a run through but it looks good.

frontend/src/views/admin/InviteCodePanelView.vue
devmount: Ah I see, alright. Let's keep it that way for now.

frontend/src/models.ts
devmount: Of course!

frontend/src/elements/AppointmentGridItem.vue
devmount: Perfect, thanks

PR #582, opened 2024-07-25
Fix aws provider issue

jdbass:
## Description of the Change
Added "terragrunt init -upgrade" ahead of "terragrunt output" commands

## Benefits

Will force Tofu to upgrade the community modules before pulling output.  This should resolve the issue of intermittent provider version errors


MelissaAutumn (approved):
Thanks!

PR #585, opened 2024-07-29
FTUE Tracking

### LINKED ISSUE #573: Track FTUE progress
Add a sentry metric that tracks the users progress through the FTUE.

We want to track a particular user's journey through the FTUE, and check where they may abandon it.

Perhaps: Generate a UUID at FTUE load, and each time the user hits 'Continue' submit this UUID and a number representing the panel they're currently on?

MelissaAutumn:
Fixes #573 

This uses posthog to track ftue level and name. As well as if the user has finished the ftue (ftue_max_level === 1)

This also sets it directly on the "person", which we can create "cohorts" of and then create insights. (like users who haven't completed the FTUE, or what step they left-off on.)

Example/ https://us.posthog.com/project/81473/person/b1e273e627c450309a7b97dc416f0562279f9e36b89b9b1af09ac90a69ca461e


backend/src/appointment/routes/metrics.py
```diff
@@ -0,0 +1,55 @@
...
+def ftue_step(
+    request: Request,
+    data: schemas.FTUEStepIn,
+    posthog: Posthog | None = Depends(get_posthog),
+    subscriber: Subscriber | None = Depends(get_subscriber),
+):
+    posthog.set(distinct_id=subscriber.unique_hash, properties={
+        'apmt.ftue.step.name': data.step_name,
+        'apmt.ftue.step.level': data.step_level
+    })
```
MelissaAutumn, line 55: We could also create an event per click, but I figured storing this on the user would be more useful.

backend/src/appointment/routes/metrics.py
```diff
@@ -0,0 +1,55 @@
...
+        'referrer': request.headers.get('Referer'),
+        'screen': data.resolution,
+        'locale': data.locale,
+        'theme': data.theme,
+    }
+    posthog.set(distinct_id=subscriber.unique_hash, properties={
+        'apmt.user.locale': data.locale,
+        'apmt.user.theme': data.theme,
+        'apmt.user.screen': data.resolution,
+        'apmt.user.ftue_max_level': subscriber.ftue_level
```
MelissaAutumn, line 38: You can only query top-level properties, so sending dicts of data is a little useless afaik

frontend/src/App.vue
```diff
@@ -164,8 +169,47 @@ const getDbData = async () => {
...
 provide('refresh', getDbData);
 // provide refresh functions for components
 provide(refreshKey, getDbData);
+
+onMounted(async () => {
+  const usePosthog = inject(usePosthogKey);
+  if (isAuthenticated.value) {
+    if (usePosthog) {
+      const profile = useUserStore();
+      posthog.identify(profile.data.uniqueHash);
```
MelissaAutumn, line 209: This sets ties frontend session to the backend session.

frontend/src/main.ts
```diff
@@ -61,6 +66,10 @@ if (useSentry) {
   });
 }
 
+if (usePosthog) {
+  posthog.init(import.meta.env.VITE_POSTHOG_PROJECT_KEY, { api_host: import.meta.env.VITE_POSTHOG_HOST, person_profiles: 'identified_only' });
```
MelissaAutumn, line 70: We only track users that are logged in.

frontend/src/main.ts
```diff
@@ -76,41 +85,5 @@ const loc = localStorage?.getItem('locale') ?? navigator.language;
 app.use(i18ninstance);
 useDayJS(app, loc);
 
-if (useSentry) {
```
MelissaAutumn, line 79: We're not using sentry metrics anymore

backend/src/appointment/secrets.py
```diff
@@ -73,3 +73,7 @@ def normalize_secrets():
         os.environ['FXA_CALLBACK'] = secrets.get('callback_url')
         os.environ['FXA_ALLOW_LIST'] = secrets.get('allow_list')
         os.environ['APP_ADMIN_ALLOW_LIST'] = secrets.get('admin_list')
+        # Need to stuff these somewhere
+        os.environ['POSTHOG_PROJECT_KEY'] = secrets.get('posthog_project_key')
+        os.environ['POSTHOT_HOST'] = secrets.get('posthog_host')
```
MelissaAutumn, line 78: Stuffed the secrets into the fxa entry. We need a better way of injecting secrets...

MelissaAutumn:
I'm still investigating a way to remove user data when they delete their account. Seems like we'll have to use their private api which isn't fully covered in the python sdk (bummer.)
https://posthog.com/docs/api

backend/src/appointment/routes/metrics.py
```diff
@@ -0,0 +1,93 @@
...
+        'url': url,
+        'referrer': request.headers.get('Referer'),
+        'screen': data.resolution,
+        'locale': data.locale,
+        'theme': data.theme,
+        '$current_url': current_url,
+    }
+
+    if not subscriber:
+        sauce = '-'.join([request.client.host, data.browser_version, data.os_version])
```
MelissaAutumn, line 49: ...I thought I renamed this..one second.

frontend/src/stores/ftue-store.ts
```diff
@@ -68,10 +68,15 @@ export const useFTUEStore = defineStore('FTUE', () => {
...
-  const nextStep = async () => {
+  const nextStep = async (call: Fetch) => {
     if (hasNextStep.value) {
       clearMessages();
       data.value.step = stepList[data.value.step].next;
+
+      call('metrics/ftue-step').post({
+        step_level: data.value.step,
+        step_name: stepList[data.value.step].title.replace('ftue.steps.', ''),
+      });
```
devmount, line 79: What if for some reason that call fails? Do we need some error/exception handling here?

devmount (approved):
Looking good, just one question!

frontend/src/stores/ftue-store.ts
MelissaAutumn: If the api returns a bad response we will simply ignore the results. useFetch doesn't seem to list any exceptions used, and neither does the fetch api afaik.

PR #587, opened 2024-07-30
Qalendar fixes

### LINKED ISSUE #544: Week view is squished to the top
**Describe the bug**
The week view doesn't extend to its container and is squished to the top

**To Reproduce**
I'm not quite sure this is consistently reproducible. Sometimes it worked for me, other times not. Possible steps to reproduce the behavior:

1. Go to schedule view
2. Switch from month to week view
3. Go to calendar view
4. Switch from month to week view

**Expected behavior**
image

**Actual behavior**
image

devmount:
## Description of the Change

- List of pending events on the calendar page now only shows future pending events (which reduces the list to a reasonable length, but let me know if this not wanted)
- First day of week is now locale aware (Monday for German, Sunday for English). In a future iteration this should be a user setting keeping these as defaults.
- I tried to fix #544 and spent quite some time debugging Qalendar for this. It turns out, that it's an issue with computed / reactive properties and Qalendar not updating their UI accordingly.

## Benefits

Consistent height of events after a change of the day boundary.

## Applicable Issues

#544

### DIFF
--- a/frontend/src/components/CalendarQalendar.vue
+++ b/frontend/src/components/CalendarQalendar.vue
@@ -1,6 +1,6 @@
 <script setup>
 import {
-  ref, computed, inject, toRefs, watch,
+  ref, computed, inject, toRefs, watch, onMounted
 } from 'vue';
 import { Qalendar } from 'qalendar';
 import 'qalendar/dist/style.css';
@@ -274,7 +274,7 @@ const calendarEvents = computed(() => {
 });
 
 /**
- * Calculate the start and end times, and then space them our by 2 hours for style!
+ * Calculate the start and end times, and then space them out by 2 hours for style!
  * @type {ComputedRef<TimeNumeric>}
  */
 const dayBoundary = computed(() => {
@@ -332,7 +332,10 @@ const config = ref({
     length: timeSlotDuration.value, // Accepts [15, 30, 60]
     height: timeSlotHeight.value, // pixel height of each length
   },
-  dayBoundaries: dayBoundary,
+  dayBoundaries: {
+    start: dayBoundary.value.start,
+    end: dayBoundary.value.end
+  },
   eventDialog: {
     // We roll our own
     isDisabled: true,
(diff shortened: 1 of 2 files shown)


frontend/src/components/CalendarQalendar.vue
```diff
@@ -312,19 +315,17 @@ const config = ref({
     length: timeSlotDuration.value, // Accepts [15, 30, 60]
     height: timeSlotHeight.value, // pixel height of each length
   },
-  dayBoundaries: dayBoundary,
+  dayBoundaries: {
+    start: dayBoundary.value.start,
+    end: dayBoundary.value.end
+  },
```
devmount, line 321: At the point of initializing this config object, dayBoundary.value holds the default values 0, 24 making Qalendar span the container element over that range. Updating the boundary later in the watcher causes Qalendar to recalculate the displayed time rows, but it doesn't update the row height.

frontend/src/components/CalendarQalendar.vue
```diff
@@ -343,6 +344,7 @@ watch(dayBoundary, () => {
     return;
   }
 
+  // TODO: This does update the boundary values, but does NOT update the container height!
   qalendarRef.value.time.DAY_START = dayBoundary.value.start * 100;
   qalendarRef.value.time.DAY_END = dayBoundary.value.end * 100;
```
devmount, line 349: Any ideas how to trigger a UI rerender of Qalendar at this point?

frontend/src/components/CalendarQalendar.vue
```diff
@@ -361,7 +363,6 @@ watch(route, () => {
 </script>
 <template>
   <div
-    class="w-full"
```
devmount, line 364: This was also set on the parent. Removing the duplicated class here.

devmount (commented):
@MelissaAutumn I found the cause for the squished / streched calendar week / day views, but still no solution, since it's kind of a Qalendar issue imo.

frontend/src/components/CalendarQalendar.vue
MelissaAutumn: We might need to adjust our fork, and then pull in our fork. But I'm not sure 🤔

MelissaAutumn (approved):
I rebased and will be merging this in to get the locale aware day of week. Thanks!

PR #592, opened 2024-07-30
Try to avoid filling attendee.timezone with null.

MelissaAutumn:
I can't re-create the problem I saw, but this was the main problem with it. I think it's possible that their tba/user schema was out of date so it was missing accessToken which is what userStore.exists checks explicitly for null. 

Maybe it would be worth versioning our tba/user and any other store data that keeps in local storage. But that's another ticket!


devmount (approved):
Thanks, yeah this is really weird. Versioning the store data might be a good long-term solution to this!

PR #594, opened 2024-07-31
Fix login if username and email differ

devmount:
## Description of the Change

Password based login doesn't work, if username and email differ for a user. This issue was introduced, when the `can-login` endpoint got a check for correct email:

image

This made logging in via username impossible, when the username isn't a valid email address. To reduce the confusion, I switched to email based login for the password login too.

## Benefits

Email is always the way to go now for both login modes.

## Applicable Issues

None.


MelissaAutumn (approved):
Thanks! Sorry for not testing this originally 😅

PR #599, opened 2024-08-01
Fix posthog init

### LINKED ISSUE #598: Only call metrics endpoint if posthog is enabled
**Describe the bug**
Just an addition to #585. The frontend throws an error if posthog is disabled because it still calls `'metrics/page-load'`.

**Expected behavior**
Check first if posthog config is set and enabled.

devmount:
## Description of the Change

This change makes sure we only call metrics if posthog is enabled.

## Benefits

Eliminates a console error.

## Applicable Issues

Closes #598


MelissaAutumn (approved):
D'oh, thanks!

PR #602, opened 2024-08-01
[Draft] Adjust post login

### LINKED ISSUE #601: Adjust post-login flow
Right now the user is shipped back from our fxa callback with the access token in the url. That's not ideal, we should ship them a one-time use token and the client should do a POST fetch to retrieve the real access token.

MelissaAutumn:
Fixes #601 

This also adds a posthog sanitize function which I'm still looking over. Tests have not been updated either.

### DIFF
--- a/frontend/src/router.ts
+++ b/frontend/src/router.ts
@@ -1,4 +1,4 @@
-import { defineAsyncComponent } from 'vue';
+import { defineAsyncComponent, inject } from 'vue';
 import { RouteRecordRaw, createRouter, createWebHistory } from 'vue-router';
 import BookingView from '@/views/BookingView.vue';
 import BookingConfirmationView from '@/views/BookingConfirmationView.vue';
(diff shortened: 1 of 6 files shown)


frontend/src/router.ts
```diff
@@ -155,7 +179,21 @@ const router = createRouter({
 });
 
 router.beforeEach((to, from) => {
-  if (!to.meta?.isPublic && !['setup', 'contact', undefined].includes(to.name)) {
+  const toMeta: ApmtRouteMeta = to?.meta ?? {};
+  const fromMeta: ApmtRouteMeta = from?.meta ?? {};
+  const usePosthog = inject(usePosthogKey);
```
devmount, line 184: I get the following console warning:

image

I'm not 100% sure if this is the cause, but this is the only place an inject was added...

devmount (approved):
Thank you! I got one warning, but despite that, login is working fine!

frontend/src/router.ts
devmount: Thank you! Have a nice weekend!

MelissaAutumn:
I've swapped the usePosthog inject to a composable and also to remove an annoying import error (not actual error) on my ide I define the instance there too.

PR #604, opened 2024-08-02
Calendar, contact, home, profile and settings page TypeScript

devmount:
## Description of the Change

This change adds types for the several pages and components, including calendar, contact, home, profile and settings pages.

- External connections store is now typed with explicitly supported providers
- Several fixes and improvements including removal of deprecated code

## Applicable Issues

Implements #413

### DIFF
--- a/frontend/src/components/SettingsGeneral.vue
+++ b/frontend/src/components/SettingsGeneral.vue
@@ -5,7 +5,6 @@ import { useI18n } from 'vue-i18n';
 import { useUserStore } from '@/stores/user-store';
 import { dayjsKey, callKey } from "@/keys";
 import { SubscriberResponse } from "@/models";
-// import SwitchToggle from '@/elements/SwitchToggle';
 
 // component constants
 const user = useUserStore();
--- a/frontend/src/components/SettingsGeneral.vue
+++ b/frontend/src/components/SettingsGeneral.vue
@@ -1,3 +1,87 @@
+<script setup lang="ts">
+import { ColorSchemes } from '@/definitions';
+import { ref, reactive, inject, watch } from 'vue';
+import { useI18n } from 'vue-i18n';
+import { useUserStore } from '@/stores/user-store';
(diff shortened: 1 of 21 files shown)


frontend/src/components/SettingsGeneral.vue
```diff
@@ -1,3 +1,87 @@
+<script setup lang="ts">
+import { ColorSchemes } from '@/definitions';
+import { ref, reactive, inject, watch } from 'vue';
+import { useI18n } from 'vue-i18n';
+import { useUserStore } from '@/stores/user-store';
+import { dayjsKey, callKey } from "@/keys";
+import { SubscriberResponse } from "@/models";
+// import SwitchToggle from '@/elements/SwitchToggle';
```
MelissaAutumn, line 8:
```suggestion

```

frontend/src/views/CalendarView.vue
```diff
@@ -124,9 +55,8 @@ const getRemoteEvents = async (from, to) => {
 
 /**
  * Retrieve new events if a user navigates to a different month
- * @param dateObj
  */
-const onDateChange = async (dateObj) => {
+const onDateChange = async (dateObj: Time<Dayjs>) => {
```
MelissaAutumn, line 59: iirc the Dayjs type doesn't properly mixin the timezone property, we might want to use the type I created for dayJsKey in keys.ts.

MelissaAutumn (approved):
Thanks, just a few suggestions!

frontend/src/views/CalendarView.vue
devmount: Thank you, I tried to do that, but unfortunately it creates issues with dj() itself, because it only accepts string, number, Date or Dayjs.

image

devmount: Resolving this for now, but I'll keep it in mind if we run into tz issues at this place.

PR #609, opened 2024-08-06
Additional metrics

MelissaAutumn:
Adds the following metrics to help guide development resources and priority:

* apmt.copy - Copy to clipboard with a generic uid and label of the button
* apmt.calendar.add - Add a calendar with provider
* apmt.calendar.edit - Edit a calendar with provider
* apmt.calendar.delete - Delete a calendar with provider
* apmt.calendar.connect - Connect a calendar with provider
* apmt.calendar.disconnect - Disconnect a calendar with provider
* apmt.calendar.sync - Sync your calendars with old and new calendar counts
* apmt.account.refreshLink - Refresh your booking link
* apmt.account.download - Download your account data
* apmt.account.deleteAccount - Delete your account
* apmt.booking.confirm - Confirmed a booking request
* apmt.booking.deny - Denied a booking request
* apmt.booking.request - Someone has requested a booking, comes with a autoConfirmed param
* apmt.schedule.created - Schedule has been created
* apmt.schedule.updated - Schedule has been updated
--

I need to double check posthog's charting abilities to see if I can get the rest of the requested captures without adding more capture calls. 

Most of the metrics are to track provider popularity, track how many times a user copies their booking link, or if we need refresh their calendars in the background instead of just on-demand.

### DIFF
--- a/frontend/src/components/ScheduleCreation.vue
+++ b/frontend/src/components/ScheduleCreation.vue
@@ -392,18 +392,22 @@ import SecondaryButton from '@/elements/SecondaryButton';
 
 // icons
 import { IconChevronDown, IconInfoCircle } from '@tabler/icons-vue';
+
 import AlertBox from '@/elements/AlertBox';
 import SwitchToggle from '@/elements/SwitchToggle';
 import ToolTip from '@/elements/ToolTip.vue';
 import { useCalendarStore } from '@/stores/calendar-store';
 import { useExternalConnectionsStore } from '@/stores/external-connections-store';
+
 import SnackishBar from '@/elements/SnackishBar.vue';
 import { dayjsKey } from '@/keys';
+import { useScheduleStore } from '@/stores/schedule-store';
 
 // component constants
 const user = useUserStore();
 const calendarStore = useCalendarStore();
 const externalConnectionStore = useExternalConnectionsStore();
+const scheduleStore = useScheduleStore();
 const { t } = useI18n();
 const dj = inject(dayjsKey);
 const call = inject('call');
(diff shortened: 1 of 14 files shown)


devmount (approved):
Looking good so far!

frontend/src/components/ScheduleCreation.vue
```diff
@@ -656,19 +660,22 @@ const saveSchedule = async (withConfirmation = true) => {
   delete obj.id;
 
   // save schedule data
-  const { data, error } = props.schedule
-    ? await call(`schedule/${props.schedule.id}`).put(obj).json()
-    : await call('schedule/').post(obj).json();
+  const response = props.schedule
+    ? await scheduleStore.updateSchedule(call, props.schedule.id, obj)
+    : await scheduleStore.createSchedule(call, obj);
```
MelissaAutumn, line 665: Unifying this from FTUE

frontend/src/views/SettingsView.vue
```diff
@@ -20,9 +20,8 @@ const { t } = useI18n({ useScope: 'global' });
 const route = useRoute();
 const router = useRouter();
 const sections = enumToObject(SettingsSections);
-const routeView = route.params.view as string;
-
-const activeView = computed<number>(() => (routeView && sections[routeView] ? sections[routeView] : SettingsSections.General));
+// Note: Use direct variables in computed, otherwise it won't be updated if transformed (like by typing)
+const activeView = computed<number>(() => (route.params.view && sections[route.params.view as string] ? sections[route.params.view as string] : SettingsSections.General));
```
MelissaAutumn, line 24: Noticed after my merge that the settings menu was broken. Since VueJS router seems to mistype route.params.view (or is it typescripts fault?) I see why we cast it as a string into a variable, but by doing that it breaks reactivity. (Unless we wrap it in a ref, which meh)

MelissaAutumn:
After an annoying amount of hacks I think I've fully sanitized the outgoing data to posthog.

frontend/src/views/SettingsView.vue
devmount: Yes this is so annoying. Didn't know this was breaking reactivity 😱 I guess the problem is, that the default route param type includes string[] as well. We might want to take a look at https://router.vuejs.org/guide/advanced/typed-routes.html, but they say its experimental?? I mean, surely we're not the only ones with that issue 😅

PR #613, opened 2024-08-08
Schedule page and several definitions TypeScript

devmount:
## Description of the Change

This change adds types for the schedule page, several definitions and related components.

- DayJsType was extended by necessary plugins

## Applicable Issues

Implements #413


frontend/src/App.vue
```diff
@@ -253,6 +253,7 @@ onMounted(async () => {
         properties.$set.$initial_pathname = REMOVED_PROPERTY;
 
         // Remove initial person url
+        // TODO: Cannot find name '$initial_person_info'. ts(2304)
         if ($initial_person_info?.u) {
           $initial_person_info.u = REMOVED_PROPERTY;
         }
```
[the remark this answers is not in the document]
devmount: Resolved now, thanks.

devmount:
Appointment store tests are working again. My bad. Forgot to modify an imported definition.

frontend/test/stores/appointment-store.test.js
```diff
@@ -35,17 +35,17 @@ const restHandlers = [
...
       ],
     },
   ])),
 ];
 
 const server = setupServer(...restHandlers);
-/* server.events.on('request:start', ({ request }) => {
-  console.log('Outgoing:', request.method, request.url);
-}); */
+// server.events.on('request:start', ({ request }) => {
```
aaspinwall, line 46: If we're not using this commented code, I suggest we remove it

frontend/src/views/ProfileView.vue
```diff
@@ -49,7 +49,7 @@ const editProfile = async () => {
     <div class="text-4xl font-light">{{ user.data.name }}</div>
     <div class="flex items-center gap-4">
       <div class="rounded-full border border-gray-500 px-2 text-xs uppercase text-gray-500">
-        {{ keyByValue(subscriberLevels, user.data.level) }}
+        {{ keyByValue(SubscriberLevels, user.data.level, true) }}
```
aaspinwall, line 52: Although this is not part of the current PR, I noticed:
`keyByValue` is using an optional boolean parameter. It's hard to understand what this is doing when reading the implementation. 

This can be fixed with an object parameter with the options you want to pass. In this case having a third parameter as `{ isEnum: boolean }` would be easier to read.

Then you can have:
`keyByValue(SubscriberLevels, user.data.level, { isEnum: true })`

frontend/src/views/ScheduleView.vue
```diff
@@ -1,109 +1,89 @@
...
   const inclusiveTo = dj(to).add(1, 'day').format('YYYY-MM-DD');
 
   calendarEvents.value = [];
   await Promise.all(connectedCalendars.value.map(async (calendar) => {
-    const { data } = await call(`rmt/cal/${calendar.id}/${from}/${inclusiveTo}`).get().json();
+    const { data }: RemoteEventListResponse = await call(`rmt/cal/${calendar.id}/${from}/${inclusiveTo}`).get().json();
     if (Array.isArray(data.value)) {
-      calendarEvents.value.push(...data.value.map((e) => ({ ...e, duration: dj(e.end).diff(dj(e.start), 'minutes') })));
+      calendarEvents.value.push(
+        ...data.value.map((e) => ({
```
aaspinwall, line 57: I would suggest changing `e` for something that tells what it is. I see it's typed as `RemoteEvent`, maybe `event` would be nice

aaspinwall (approved):
Nothing blocking, just a few comments for readability

frontend/src/views/ProfileView.vue
devmount: Thank you for this suggestion. For your proposed solution I would need to change the function signature though, otherwise TS will complain about an argument of type '{ isEnum: boolean; }' not being assignable to parameter of type 'boolean'. I'll take a note to type this kind of function with an own signature and improve that after the heavy TS work is done. For now, your IDE will reveal the boolean argument name when typing or hovering that function 😇

PR #623, opened 2024-08-29
update readme with deploy instructions

malini:
## Description of the Change

Added instructions on how to deploy.

## Benefits

We don't have explicit documentation here, so we should document it within the codebase


devmount (approved):
Thanks Malini!

PR #627, opened 2024-09-03
Sitemap

malini:
## Description of the Change

add a sitemap.txt so we can point the Google Console to the url we upload it to, so we can get listed in google

## Benefits
SEO


MelissaAutumn (approved):

MelissaAutumn (changes requested):
actually I need to double check paths

MelissaAutumn (approved):

Issue #297, opened 2024-03-11
Determine optimal new user experience


malini:
**Is your feature request related to a problem? Please describe.**
After signing in, we currently dump the user onto https://stage.appointment.day/calendar/month view, without any guidance on what to do next.

**Describe the solution you'd like**
We should determine what experience we want to give first time users. A pop-up tutorial? A guide? etc.

Ideally, we should also add tracking stats like how many people finished the on-boarding and went on to use the product, etc.

laurelterlesky:
FTUE extra-small breakpoint 375px (< 599 px) is ready for dev:
https://www.figma.com/design/yBYCA41d0cHf50nOt4Uz4O/TBA_2024_Q3_FTUE-responsive?node-id=2095-4472&t=thRt81zT82DMvjw5-1

cc: @solangevalverde 

FTUE Dark Mode is ready for dev:
https://www.figma.com/design/hbXmyeGmHepX61zrqrHSaD/TBA_2024_Q2_FTUE?node-id=4503-10396&t=kUtBypt5jzptOKXL-1

cc: @micahilbery

PR #636, opened 2024-09-05
FTUE Dark-mode

MelissaAutumn:
Related: #297 

The majority of work here was re-organizing the pro services colours. A lot of the components are using an older colour spec. There's some obvious differences from the design due to this, so matched the best I could.

I've also moved bubble-select to the tbpro subdirectory, since it's a pro component. 

cc @laurelterlesky 

### Screenshots

Screen Shot 2024-09-05 at 13 49 55-fullpage
Screen Shot 2024-09-05 at 13 50 00-fullpage
Screen Shot 2024-09-05 at 13 50 12-fullpage
Screen Shot 2024-09-05 at 13 51 34-fullpage
Screen Shot 2024-09-05 at 13 51 47-fullpage
Screen Shot 2024-09-05 at 13 53 17-fullpage
Screen Shot 2024-09-05 at 13 54 00-fullpage


docker-compose.yml
```diff
@@ -19,7 +19,7 @@ services:
     build: ./frontend
     volumes:
       - './frontend:/app'
-      - 'node_modules:/app/node_modules'
+      - '/app/node_modules'
```
MelissaAutumn, line 22: I went back to anon volumes. You can do `docker-compose up --build -V` to nuke the built node_modules when deps update.

frontend/index.html
```diff
@@ -1,16 +1,14 @@
 <!DOCTYPE html>
-<html lang="">
+<html lang="" data-tbpro-service="apmt">
```
MelissaAutumn, line 2: Used in the css to select the correct service colours. We could also use another solution but I had the `.dark` class on the brain. Alternatively you can use "send" as the value to get send's colours.

laurelterlesky:
Looks great Mel! A few little things I spotted - not sure if you're only looking at colour atm and what I'm listing isn't of focus:

- All FTUE screens are missing the faded footer above the support link (footer component)
- Create profile screen: missing required (red) stars next to Full Name, Username URL
- Connect Calendars page: the check mark selected state should have a primary teal bkg with a black check
-  Connect Calendars page: Select all btn placement should be lower centered on the link between the header calendar selection area
- Set up your appointment availability: missing required (red star) next to select days
- Connect video: Zoom logo needs to be white in dark mode
- Read to start booking page: bring the copy higher, close to the icons (should be about 24px between bottom of icons to text)

(I can't see active states with the input fields, but in case, they should have primary teal for the outline)

laurelterlesky:
Mel's keen eye caught one more item - the glow under the modal is missing on all screens. 

and I'm also wondering - is it possible to show the empty calendar interface / dashboard under a scrim during FTUE modals?

MelissaAutumn:
Fixed underglow and divider, thanks for the spots. I'll fix those up too.
Screen Shot 2024-09-05 at 14 40 49-fullpage


I'll look into adding a page behind the overlay/scrim, it's a little tricky though.

laurelterlesky:
Glow up! Thanks - looks beautiful!

MelissaAutumn:
Added fake dashboard behind the scrim/overlay. That page was built to require data so it's not as easy as making the FTUE a component modal. But I've faked the data and inserted it into the page. It seems to work just fine.
image


Also the red stars is there, it only shows up if there's no data in the input. I can adjust that if you'd like!
image

And here's the active ring: (ignore 1password)
image

laurelterlesky:
Is the active state green outline A3ECE3? (primary green (dark mode) looks like light mode primary atm

And which red are you using for the required state outline?  Is it we can use the dark mode text-critical colour for the outline: EF4444

laurelterlesky:
and the dashboard behind the scrim looks AMAZING! Magic wizardy :)

MelissaAutumn:
Active outline is tailwind messery. I'll try to decouple it, but it might break the old-style elements. The sooner we re-design the entire app, the sooner I can remove a lot of css-hacks 😄 

And for the error outline it's the text/icon critical `EF4444`.

MelissaAutumn:
Additionally there doesn't seem to be a way to adjust the checkbox in css without faking the checkmark. That's...an odd oversight. I can look into that in the future if that's alright!

frontend/index.html
MelissaAutumn: cc @aaspinwall for the future where send has a ui and I've actually split this off into its own repo 😅

laurelterlesky:
tailwind misery. No worries - I don't want these little things to block this.

devmount (approved):
Looks awesome! Great work Mel!

Just some things I observed (might be out of scope here though):

1. Google still has the earlier appointment logo when connecting an account:
    image
2. The input field for the schedule title "My schedule" looks a bit out of place without border radius and the placeholder without left padding:
    image
3. The "select all" buttons in calendar selection broke out to the bottom for some reason:
    image
    We might want to overwrite that inherited top attribute:
    image

docker-compose.yml
devmount: Worked perfectly fine, thanks for the hint!

MelissaAutumn:
> Looks awesome! Great work Mel!
> 
> Just some things I observed (might be out of scope here though):
> 
>     1. Google still has the earlier appointment logo when connecting an account:
>        image
> 
>     2. The input field for the schedule title "My schedule" looks a bit out of place without border radius and the placeholder without left padding:
>        image
> 
>     3. The "select all" buttons in calendar selection broke out to the bottom for some reason:
>        image
>        We might want to overwrite that inherited top attribute:
>        image

Thanks for the logo catch! I'll update that, and into the lack of padding.

As for the placement of the Select all button, that's actually how it is designed. 😅

devmount:
Oh haha, but in your screenshots above the Select all button is positioned more on top 🤔

MelissaAutumn:
Ah yeah it somehow reverted its position, laurel caught it and commented on that button.

PR #650, opened 2024-09-11
Schedule creation step 4, Timezone step 1

### LINKED ISSUE #578: What Timezone Am I In?
**Describe the bug**
When setting a schedule, I assume the timezone it is displaying is the one that I have set in settings - but, I'm not sure! My timezone in the settings is Central Time (US), I am currently in Mountain Time (US). I have no idea what timezone I'm looking at in the scheduler.

** Proposed solution **
It should display the timezone you are scheduling in, with an option to change it right there. Maybe small, looking like "CDT (edit)" - Where the "edit" is a blue link you can click, could be "change" instead.

Screenshot 2024-07-24 at 8 04 47 AM

devmount:
## Description of the Change

- Step 4 "Booking Settings" was added to the Schedule creation form
- Tbpro TextInput was extended to support prefix text that is not part of the actual input
- The slug can now be customized, randomized and saved along with the existing schedule data
- Step 1 now includes timezone information

image

Sorry, I needed to upgrade Vue to 3.5 (to use templateRefs), but than I dropped that idea but kept the newer Vue 😅 

## Applicable Issues

Implements #460 
Closes #578


frontend/src/components/ScheduleCreation.vue
```diff
@@ -314,7 +323,12 @@ const saveSchedule = async (withConfirmation = true) => {
   emit('created');
   // Update our reference schedule!
   referenceSchedule.value = { ...scheduleInput.value };
-  resetSchedule(false);
+  revertForm(false);
+};
+
+// Update slug with a random 8 character string
+const refreshSlug = () => {
+  slugInput.value = (Math.random() + 1).toString(36).substring(4);
```
MelissaAutumn, line 331: We can probably substring https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID

frontend/src/components/ScheduleCreation.vue
```diff
@@ -673,18 +691,73 @@ watch(
...
+              :active="schedule.booking_confirmation"
+              :label="t('label.bookingConfirmation')"
+              :disabled="!scheduleInput.active"
+              @changed="toggleBookingConfirmation"
+              no-legend
+            />
+            <div class="whitespace-pre-line rounded-lg bg-white p-4 text-xs text-gray-500 dark:bg-gray-800">
+              <div>
+                {{ t('text.yourQuickLinkIs', { url: user.myLink }) }}<br />
+                <i18n-t keypath="text.toUpdateYourUsername.text" for="text.toUpdateYourUsername.link">
```
MelissaAutumn, line 747: I reccommend using the slot-based method which I discovered recently
https://vue-i18n.intlify.dev/guide/advanced/component.html#slots-syntax-usage

It's slightly more readable.

frontend/src/tbpro/elements/TextInput.vue
```diff
@@ -25,21 +29,28 @@ interface Props {
...
   required?: boolean;
   disabled?: boolean;
 }
-withDefaults(defineProps<Props>(), {
+const props = withDefaults(defineProps<Props>(), {
   type: 'text',
   help: null,
   remoteError: null,
   placeholder: '',
+  prefix: '',
```
MelissaAutumn, line 41: Can we set this to null? (Placeholder should also be null but that's my fault.)

frontend/src/tbpro/elements/TextInput.vue
```diff
@@ -61,20 +72,23 @@ const onChange = () => {
...
-      :disabled="disabled"
-      :placeholder="placeholder"
-      :required="required"
-      @invalid="onInvalid"
-      @change="onChange"
-      ref="inputRef"
-    />
+    <div class="tbpro-input">
+      <span v-if="prefix" ref="inputPrefix" class="prefix">{{ prefix }}</span>
+      <input
```
MelissaAutumn, line 77: Can we class the input? I believe I ran into a bug (in another component) where non-class/id selectors would not be scoped for child components. While this doesn't have any child components with inputs, I'd rather keep it explicit.

MelissaAutumn (approved):
Looks good other than a few points.

MelissaAutumn (commented):
Adjusting approval until backend is in 😅

frontend/src/components/ScheduleCreation.vue
devmount: That's better, thanks!

frontend/src/components/ScheduleCreation.vue
devmount: Ah nice! Changed it.

frontend/src/tbpro/elements/TextInput.vue
devmount: Of course. Haha sorry, sometimes I just duplicate properties as they are when adding new ones :D

frontend/src/tbpro/elements/TextInput.vue
devmount: Of course, done.

devmount (commented):
@MelissaAutumn This PR is ready now. Never mind about additional backend stuff, it was not needed. I could simply use the slug field when updating the schedule 🤷🏻‍♂️  A classic case of "Ich seh den Wald vor lauter Bäumen nicht" (I can't see the forest because of the many trees) 😂

PR #657, opened 2024-09-12
Migrate schedule creation inputs and buttons to tbpro components

### LINKED ISSUE #460: "No Confirmation Needed" or "Confirmation Needed"

devmount:
## Description of the Change

This change migrates text inputs, selects, primary buttons and secondary buttons to use tbpro components.

## Benefits

One step further to the new design.

## Applicable Issues

Implements #460


MelissaAutumn:
Looks good, but missed one 😄 

image

MelissaAutumn (approved):

devmount:
@MelissaAutumn Yeah, this is kind of a special field and the tbpro text-input does not really fit in here... 😅 

image

But I can extend it to be "invisible" (without borders or background).

MelissaAutumn:
That would probably work foe now

devmount:
I removed the background for now:
image

PR #658, opened 2024-09-13
Profanity filter

### LINKED ISSUE #574: Filter profanity/etc from public user names/urls
We do not filter user's shortUrls and we should check them against a list of known profanity/etc, so we do not allow harmful language.

Determine if Mozilla maintains a profanity filter somewhere we can leverage, or find a suitable alternative (maybe https://github.com/coffee-and-fun/google-profanity-words ?

When users select their username/shortUrls/etc, we should run their input through the filter before accepting it.

devmount:
## Description of the Change

- Cleaned up and removed all deprecated inject/provide keys
- Researched and tried several profanity filters and ended up using 2Toad/Profanity
- Implemented frontend validation for schedule creation, account settings and FTUE
- The TextInput tbpro component now supports custom error messages (in addition to form validation errors) 

Note that only full words are detected. So "p ass word" is a match and "password" is none.

## Benefits

Cleaner codebase and hopefully a little less offensive language.

image
image
image
profanity1
profanity2

## Applicable Issues

Closes #574

### DIFF
--- a/frontend/src/components/SettingsAccount.vue
+++ b/frontend/src/components/SettingsAccount.vue
@@ -5,7 +5,7 @@ import {
 import { useI18n } from 'vue-i18n';
 import { useRouter } from 'vue-router';
 import { useUserStore } from '@/stores/user-store';
-import { callKey } from '@/keys';
+import { callKey, hasProfanityKey } from '@/keys';
 import { StringListResponse, SubscriberResponse, BlobResponse, BooleanResponse } from '@/models';
 import CautionButton from '@/elements/CautionButton.vue';
 import ConfirmationModal from '@/components/ConfirmationModal.vue';
@@ -27,6 +27,7 @@ import { usePosthog, posthog } from '@/composables/posthog';
 // component constants
 const { t } = useI18n({ useScope: 'global' });
 const call = inject(callKey);
+const hasProfanity = inject(hasProfanityKey);
 const router = useRouter();
 const user = useUserStore();
 const schedule = useScheduleStore();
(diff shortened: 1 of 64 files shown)


devmount:
@malini Of course. I did that here: https://github.com/thunderbird/appointment/issues/574#issuecomment-2348616831

frontend/src/components/SettingsAccount.vue
```diff
@@ -214,11 +241,17 @@ const actuallyDeleteAccount = async () => {
...
-          v-model="activeDisplayName"
-          type="text"
-          class="w-full rounded-md"
-        />
+        <div class="w-full">
+          <input
+            v-model="activeDisplayName"
+            type="text"
+            class="w-full rounded-md"
+            :class="{ '!border-red-500': errorDisplayName }"
```
MelissaAutumn, line 249: add :required="true"

frontend/src/components/SettingsAccount.vue
```diff
@@ -66,34 +67,60 @@ const refreshData = async () => Promise.all([
...
  * Check if the username has been changed, and open a modal to warn the user their short link is going to change
  * If it didn't change, then just update the user immediately.
  */
 const updateUserCheckForConfirmation = async () => {
+  errorUsername.value = null;
+  errorDisplayName.value = null;
+
+  // Validate username
+  if (activeUsername.value === '') {
+    errorUsername.value = t('validation.fieldIsRequired', { field: t('label.username') });
```
MelissaAutumn, line 109: I feel like this should already trigger with the required attribute, since we use html form validation.

MelissaAutumn (approved):
Looks good. I tend to lean on the html form element since the browser tends to handle errors better than homebrew methods. So I'd suggest refactoring the one comment I have but good to merge after!

frontend/src/components/SettingsAccount.vue
devmount: Is the display name actually required? I thought it was optional?

MelissaAutumn: Ah misread the field. Meant to add it under activeUsername

devmount: Ah alright, that makes sense, thank you!

devmount: The problem here is, that it triggers the modal and does request the backend, giving a less accurate error message:

image

frontend/src/components/SettingsAccount.vue
devmount: At this place, we don't 😅 We just open the modal for confirmation of the username change...

devmount: @MelissaAutumn I can change that to use form validation next week if you like!

MelissaAutumn: Sounds good to me, thank you!

devmount: I added form validation at this place now.

devmount:
Added some final fixes and frontend code linting.

PR #667, opened 2024-09-16
Redesign: Switch toggle input

devmount:
## Description of the Change

This change styles the `<switch-toggle>` component according to the [latest design changes](https://www.figma.com/design/rDjrWQsgejQUyACjjXRwVa/TBA_2024_Q3_BookingConfirmationToggle?node-id=4046-524&node-type=frame&t=cMv21ItaNmVBnBIc-0). All Tailwindcss classes were removed and migrated to local css scope while using defined CSS custom properties for colours.

It is currently only implemented in the schedule creation component.

image

## Benefits

Next TBPro component for the use in all services.


frontend/src/assets/styles/effects.css
devmount, line 1: Now that I think about it, that could also live in variables.css

MelissaAutumn: variables works for now!

frontend/src/assets/styles/colours.css
```diff
@@ -80,6 +80,16 @@ html {
   --colour-ti-critical: #DC2626;
   --colour-ti-success: #1D783B;
 
+  /*
+   * Form elements
+   * Applied frequently for inputs, switches and interactive elements.
+   */
+  --colour-form-base: #E4E4E7;
```
MelissaAutumn, line 87: I believe this is actually element specific. I would maybe move these to the SwitchToggle. 

But I'll cc @laurelterlesky for comment.

frontend/src/tbpro/elements/SwitchToggle.vue
```diff
@@ -0,0 +1,107 @@
...
+
+// component constants
+const { t } = useI18n();
+
+// component emits
+const emit = defineEmits(['changed']);
+
+// component properties
+interface Props {
+  active: boolean; // initial toggle state
```
MelissaAutumn, line 13: Can you add name so we can use form elements? Thanks! 

(I know we don't use them right now except for login and contact, but it might come in handy when developing traditional forms)

frontend/src/tbpro/elements/SwitchToggle.vue
```diff
@@ -0,0 +1,107 @@
...
+  line-height: 1rem;
+  color: var(--colour-ti-secondary);
+}
+
+.toggle {
+  background: var(--colour-form-base);
+  width: 2rem;
+  height: 1rem;
+  flex-shrink: 0;
+  border-radius: 9999px;
```
MelissaAutumn, line 77: 100% also works!

MelissaAutumn (commented):
Just awaiting comment by @laurelterlesky

frontend/src/tbpro/elements/SwitchToggle.vue
devmount: Well, 100% used to create weird elliptical shapes. Which is the reason, why Tailwind still uses 9999px. But might be a thing of the past now 😇 
https://github.com/tailwindlabs/tailwindcss/discussions/3373

laurelterlesky:
I just looked in the FTUE on staging for this - but I don't see the revised toggle. Where should I look? (Happy to just have a screen grab too)

frontend/src/tbpro/elements/SwitchToggle.vue
devmount: Of course, good thinking!

laurelterlesky:
Wonderful - looks great!

MelissaAutumn (approved):
Good to go after adjusting changes.

Issue #487, opened 2024-06-13
Switch UUIDType to SQLAlchemy UUID


MelissaAutumn:
From:
https://sqlalchemy-utils.readthedocs.io/en/latest/data_types.html#module-sqlalchemy_utils.types.uuid

To:
https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.Uuid

Currently our database has it as a Binary(16) field.

MelissaAutumn:
We should make sure the existing data carries over to the new/updated field.

MelissaAutumn:
This has been fixed as of #1075

MelissaAutumn:
There was a sentry error logged after the containers ran, but I can't reproduce it. It's possible there was a db cache that wasn't cleared. 

Seems harmless https://thunderbird.sentry.io/issues/6712908595/?environment=stage&project=4505428124827648&query=is%3Aunresolved&referrer=issue-stream&stream_index=1

PR #668, opened 2024-09-16
feat: Switched UUIDType to SQLAlchemy UUID

Mr-Sunglasses:
## Description of the Change

Switched UUIDType to SQLAlchemy UUID

## Benefits

Add Native SQLAlchemy UUID Support which can help in offer  potential performance gains. 

## Applicable Issues

fix: #487

### DIFF
--- a/backend/src/appointment/database/models.py
+++ b/backend/src/appointment/database/models.py
@@ -12,7 +12,8 @@ import zoneinfo
 from functools import cached_property
 
 from sqlalchemy import Column, ForeignKey, Integer, String, DateTime, Enum, Boolean, JSON, Date, Time
-from sqlalchemy_utils import StringEncryptedType, ChoiceType, UUIDType
+from sqlalchemy.types import UUID
+from sqlalchemy_utils import StringEncryptedType, ChoiceType
 from sqlalchemy_utils.types.encrypted.encrypted_type import AesEngine
 from sqlalchemy.orm import relationship, as_declarative, declared_attr, Mapped
 from sqlalchemy.sql import func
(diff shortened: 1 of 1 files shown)


backend/src/appointment/database/models.py
```diff
@@ -203,7 +204,7 @@ class Appointment(Base):
     __tablename__ = 'appointments'
 
     id = Column(Integer, primary_key=True, index=True)
-    uuid = Column(UUIDType(native=False), default=uuid.uuid4, index=True, unique=True)
+    uuid = Column(UUID(as_uuid=True), default=uuid.uuid4, index=True, unique=True)
```
devmount, line 207: Does this need any migration?

devmount (commented):
Thanks again for helping us, @Mr-Sunglasses 🙌🏻 One question below.

backend/src/appointment/database/models.py
Mr-Sunglasses: No schema migration is needed. The underlying database column type (Binary(16)) remains the same, It's just that `sqlalchemy_utils.UUIDType` is changed to `sqlalchemy.types.UUID`

devmount: I see. I think the requirement in #487 was to migrate this to 32 character hexadecimal strings using [sqlalchemy.Uuid](https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.Uuid)

CC @MelissaAutumn

Mr-Sunglasses: Ah, I misunderstood the issue, let me migrate to `32 character hexadecimal strings` using [sqlalchemy.Uuid](https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.Uuid)

cc. @devmount @MelissaAutumn

Mr-Sunglasses: > No schema migration is needed. The underlying database column type (Binary(16)) remains the same, It's just that `sqlalchemy_utils.UUIDType` is changed to `sqlalchemy.types.UUID`

So I was wrong here, This change will store UUIDs as 32-character hexadecimal strings in the database, and Yes, this change will require a database migration. 

Again i'm sorry for all this confusion, my bad 😓 .

cc. @devmount

MelissaAutumn:
Thanks for the PR! Unfortunately this is missing the data migration to transfer existing uuids, and in general it's too big of a change for now.

backend/src/appointment/database/models.py
devmount: All good. Thanks anyway for your work on this!

PR #670, opened 2024-09-16
Perform server-side validation for bookings (Fixes #669)

### LINKED ISSUE #669: Backend validation for time slots

MelissaAutumn:
Fixes #669 

Doing a sweep of backend validation issues. 

For tests we should always declare the timezone as UTC time. The frontend only ever sends utc time so it's not a problem. To make the code easier to maintain we'll fail any non-UTC time coming in through the booking route.

### DIFF
--- a/backend/test/integration/test_schedule.py
+++ b/backend/test/integration/test_schedule.py
@@ -1,8 +1,7 @@
 import zoneinfo
-from datetime import date, time, datetime, timedelta
+from datetime import date, time, datetime, timedelta, timezone
 from unittest.mock import patch
 
-import pytest
 from freezegun import freeze_time
 
 from appointment.tasks import emails as email_tasks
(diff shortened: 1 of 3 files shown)


backend/test/integration/test_schedule.py
```diff
@@ -497,10 +502,7 @@ def bust_cached_events(self, all_calendars=False):
         assert data.get('detail').get('id') == validation.SlotAlreadyTakenException.id_code
 
         # Okay change up the time
-        start_time = time(11)
-
-        slot_availability['slot']['start'] = datetime.combine(start_date, start_time).isoformat()
-
+        slot_availability['slot']['start'] = datetime.combine(start_date, time(11), tzinfo=timezone.utc).isoformat()
```
MelissaAutumn, line 505: Changing start_time re-evals start_datetime for the staticmethod. Oops.

devmount (approved):
Tested and works. Thanks!

PR #683, opened 2024-09-24
Schedule creation UI enhancement

### LINKED ISSUE #460: "No Confirmation Needed" or "Confirmation Needed"

devmount:
## Description of the Change

This change implements most of the remaining tasks to implement [the current design](https://www.figma.com/design/rDjrWQsgejQUyACjjXRwVa/TBA_2024_Q3_BookingConfirmationToggle?node-id=1-6&node-type=canvas&t=WwFZ4pgr9bGR79Rr-0) for the schedule creation section:

- Input labels
- Spacing
- Snackish bar
- Light / dark mode

## Benefits

image

a

## Applicable Issues

Closes #460


frontend/src/components/ScheduleCreation.vue
```diff
@@ -660,12 +651,12 @@ watch(
               @change="toggleZoomLinkCreation"
               class="size-5 rounded-md"
             />
```
MelissaAutumn, line 653: This can be a separate PR if you want, but we should normalize these checkboxes with the SyncCard's checkboxes. (And then pull those checkboxes into SyncCard 😅 )

MelissaAutumn (approved):
Looks good, thanks!

frontend/src/components/ScheduleCreation.vue
devmount: Ah that is a great suggestion! I forgot we already had styled checkboxes! I did it right away.

devmount:
@MelissaAutumn If you like you can test again, I added the checkbox component now.

MelissaAutumn:
Looks like I can still edit the weekday selections and the slug input while the schedule is disabled. I'll make a ticket so it doesn't block the merge of this though.

devmount:
> Looks like I can still edit the weekday selections and the slug input while the schedule is disabled. I'll make a ticket so it doesn't block the merge of this though.

Good catch! Yes please create a ticket, I'll look into that

PR #694, opened 2024-09-26
Fix weekday selection and quick link to get disabled

### LINKED ISSUE #686: Some fields are editable while schedule is deactivated
The quick link field, and weekday selection are still editable while the schedule is disabled. They should display the "no" cursor and not be interactable.

devmount:
## Description of the Change

- Add the ability to be disabled for BubbleSelect
- Fix the quick link field to get disabled on schedule deactivation

image

image

## Context

I also added the changes to a corresponding PR in services-ui:
https://github.com/thunderbird/services-ui/pull/4

## Applicable Issues

Closes #686

### DIFF
--- a/frontend/src/components/ScheduleCreation.vue
+++ b/frontend/src/components/ScheduleCreation.vue
@@ -504,10 +504,10 @@ watch(
               {{ t("label.availableDays") }}
             </div>
             <bubble-select
-              class="bubble-select"
               :options="scheduleDayOptions"
               v-model="scheduleInput.weekdays"
               :required="true"
+              :disabled="!scheduleInput.active"
             />
           </div>
           <div>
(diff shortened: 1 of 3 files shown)


frontend/src/tbpro/elements/TextInput.vue
```diff
@@ -167,6 +167,7 @@ const onChange = () => {
     overflow: hidden;
     max-width: 40%;
     text-wrap: nowrap;
+    z-index: 1;
```
devmount, line 170: For some weird reason the prefix element gets hidden behind the input as soon as that gets disabled 🤷🏻

frontend/src/components/ScheduleCreation.vue
```diff
@@ -872,10 +873,6 @@ watch(
   display: block;
 }
 
-.bubble-select {
-  gap: .25rem;
-}
-
```
devmount, line 878: This is handled by the BubbleSelect component itself now and hence not needed anymore.

MelissaAutumn (approved):

PR #697, opened 2024-09-26
Fix booking link toggle in schedule creation

### LINKED ISSUE #695: After removing one google account and adding another my schedule is "turned off"
**Describe the bug**
I initially added one of my Google Accounts, which seemed to work.
I then removed that account and added another one, enabling new calendars.
Now following my invite link tells me:

> The schedule has been turned off. Please contact the schedule owner for more information.

**To Reproduce**

1. Add a Google account
2. Remove that Google Account, add a new one
3. Go to my appointment booking link

**Expected behavior**
It should enable my booking.

**Actual behavior**
It doesn't.

**Additional context**

On the [dashboard](https://appointment.day/dashboard) I see this in the web console being logged:

```
[Vue warn]: Error in https://vuejs.org/error-reference/#runtime-1: "TypeError: can't access property "booking_confirmation", te.schedule is null"
```

devmount:
## Description of the Change

- Booking confirmation is now properly handled via schedule input object. This fixes a bug where a direct call on the retrieved schedule object happend on null.
- The SwitchToggle component now handles state changes from the parent. This sometimes happens during initialization when waiting for async data.

## Applicable Issues

Fixes #695

### DIFF
--- a/frontend/src/tbpro/elements/SwitchToggle.vue
+++ b/frontend/src/tbpro/elements/SwitchToggle.vue
@@ -1,5 +1,5 @@
 <script setup lang="ts">
-import { ref, onMounted } from 'vue';
+import { ref, onMounted, watch } from 'vue';
 import { useI18n } from 'vue-i18n';
 
 // component constants
@@ -29,6 +29,15 @@ const toggleState = () => {
     emit('changed', state.value);
   }
 };
+
+// Update state if parent changes active prop
+// e.g. after initializing async data
+watch(
+  () => props.active,
+  () => {
+    state.value = props.active;
+  }
+);
 </script>
 
 <template>
--- a/frontend/src/tbpro/elements/SwitchToggle.vue
+++ b/frontend/src/tbpro/elements/SwitchToggle.vue
@@ -16,7 +16,11 @@ interface Props {
   label?: string; // input label
   noLegend?: boolean; // hide "on" and "off" labels
 }
-const props = defineProps<Props>();
+const props = withDefaults(defineProps<Props>(), {
+  disabled: false,
+  label: null,
+  noLegend: true,
+});
 
 // current state
 const state = ref(false);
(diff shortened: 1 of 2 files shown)


frontend/src/tbpro/elements/SwitchToggle.vue
```diff
@@ -29,6 +29,15 @@ const toggleState = () => {
...
 };
+
+// Update state if parent changes active prop
+// e.g. after initializing async data
+watch(
+  () => props.active,
+  () => {
+    state.value = props.active;
+  }
+);
```
MelissaAutumn, line 40: I think you might be able to {active} = toRef(props) instead of this.

MelissaAutumn (approved):
Looks good, one suggestion though!

frontend/src/tbpro/elements/SwitchToggle.vue
devmount: Actually I tried `const state = toRef(props.active);` but it won't properly initialize state on page load... 😅 Or I'm just to stupid to properly implement this 😂 Maybe I oversaw sth?

MelissaAutumn: Oh I think you need to toRef props itself and destructing assign active.

so `{ active } = toRef(props);`

but i can take a look on tuesday!

devmount: Ah, you mean `toRefs()` (with an 's')? OR actually `toRef()`? I tried the former too but it complained about being read-only when updating that ref 😂

MelissaAutumn: This works for now 😄

devmount: haha alright! thanks

PR #702, opened 2024-10-01
Filter out declined for Google Events

### LINKED ISSUE #696: Declined events show up in calendar
**Describe the bug**
If I decline an invite to an event, the event is added to my calendar with its attendance set to "Declined". However, unless I fully remove it from my calendar, it still shows up in my dashboard, and as a "Busy" period in my quick link.

**To Reproduce**

1. Connect your Google Calendar (might happen with CalDAV too but I haven't tested it)
2. Send yourself an event invite, from another account to your Google account
3. Decline the event from your Google account

**Expected behavior**

The event doesn't show up in my dashboard (or if it shows, it's with a visual distinction that shows it won't affect my availability) and on my availability as shown by my quick link.

**Actual behavior**

The event shows up like events which attendance isn't "Declined".

**Screenshots**

screenshot_declined_event

Event in my Google calendar (the "me" with an avatar is the Google account that's synced up in Appointment)

image

Event shown in my Appointment dashboard

image

Availability gap via my quick link

- OS: Windows 11
- Browser Version: Firefox 130.0.1 (64-bit)

MelissaAutumn:
Fixes #696 

The fix is only applied to the google calendar connector because we don't actually of a method of determining which attendee is the calendar owner. (As they could technically have no email associated with a caldav calendar...) This will hopefully be cleared up with the freebusy api.

### DIFF
--- a/backend/src/appointment/controller/calendar.py
+++ b/backend/src/appointment/controller/calendar.py
@@ -185,8 +185,17 @@ class GoogleConnector(BaseConnector):
             if status == 'cancelled' or transparency == 'transparent':
                 continue
 
-            # Mark tentative events
+            # Grab the attendee list for marking tentative events / filtering out declined events
             attendees = event.get('attendees') or []
+            declined = any(
+                (attendee.get('self') and attendee.get('responseStatus') == 'declined') for attendee in attendees
+            )
+
+            # Don't show declined events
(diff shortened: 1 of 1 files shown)


devmount (approved):
Yes, this works until we use the freebusy API. Thank you!

PR #706, opened 2024-10-02
Swap the :checked state to a v-model that applies on change event.

### LINKED ISSUE #705: Generate Zoom Meeting checkbox does not restore state on page load
When you load the page the current state of the "Generate zoom meeting" checkbox does not load into the form.

MelissaAutumn:
Fixes #705 

Originally we used state because v-model wasn't working. Now it works, so uhh. I moved it back.


devmount (approved):
Thank you, `v-model` is indeed better here! Sorry that might have been me when switching to tbpro components 😬

Issue #120, opened 2023-07-20
Built-in Authentication/Plug-in Authentication


ryanleesipes:
We are going to open source Appointment, eventually. We will need some kind of authentication method - or/and the ability to plug in alternative authentication methods so that self-hosters can have an account system easily enough.

MelissaAutumn:
Reviving this ancient ticket. For password auth, a user will be created if there's no existing user in the database.

After that registration will be locked and users can be added via our basic admin panel.

MelissaAutumn:
This was fixed in https://github.com/thunderbird/appointment/pull/716

PR #716, opened 2024-10-08
Password Login Fixes

MelissaAutumn:
Part of #120 

This allows the password flow to create a new subscriber on first-login. But only first-login. (Well when there's no subscribers.) 

Input wanted: Do we need to put this behaviour behind a env flag for extra security? 

I discovered a fun bug with schedule creation/updates that causes it to fail because times are in UTC and 14:00 as a start-time is bigger than 0:00 as an end time. So we now collect timezones for schedules, and store them. (That'll be handy later!) We don't do anything else with them right now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,10 @@ Check out the project's respective readmes:
 
 This project uses [Fluent](https://projectfluent.org/) for localization. Files are located in their respective `l10n/<locale>/*.ftl`.
 
+### Self-hosting
+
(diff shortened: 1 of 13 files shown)


malini:
+1 for hiding this behind an env flag! Best to isolate this behaviour from prod as much as possible

MelissaAutumn:
> +1 for hiding this behind an env flag! Best to isolate this behaviour from prod as much as possible

Sounds good, it's only active with password auth and not fxa as that will always create new users.

MelissaAutumn:
And added the env gate.

devmount (approved):
Tested this and looks like it works without any issues! Thank you!

PR #718, opened 2024-10-10
CalDAV autodiscovery

### LINKED ISSUE #446: CalDAV auto discovery
We use to have this, and now we don't. We should re-add it.

MelissaAutumn:
Fixes #446 

I started off just re-enabling the old behaviour but found it didn't really work in context of the FTUE. So I've added the ability to save CalDAV principal servers as an external connection. That required a lot of additional work, whoops!

Note: This does not affect the "Add a single caldav calendar" functionality that exists on the calendars setting page. This simply treats a caldav principal server connection like we would google oauth. 

### Changes:
* Added a python Tools method to fix-up malformed caldav urls
* Added a python Tools method to perform a dns SRV lookup for a caldav entry
* Implemented sync_calendars for caldav connector
* Renamed ftueStep.GooglePermissions to ftueStep.CalendarProvider
* Add a switch button during the FTUE calendar provider step
* Add modal text
* Refactor LoginView::handleFormError to be generic and now lives in utils
* Allow force fetching of external connections
* Add CalDAV as an external connection option
* Refactor GoogleOauthProvider to be more in-line with how CalDAVProvider works
* Add shortcuts for decrypt/encrypt in Python utils
* Cache dns lookup for the remainder of the ttl

### Screenshots
image
image
image
image


devmount (changes requested):
First: Great work, Melissa, as usual! 👏🏻 The setup is easy and all my Nexcloud CalDAV calendars were found 🎉 

One thing to mention though: We might need a check if the CalDAV calendar URL is already there. E.g. I had on of 5 calendars from one principal already active and now I have the same calendar twice under calendars.

Also found some minor things with migrations (again, sorry 🙈 ), props and unsused imports. Thanks again for this!

MelissaAutumn:
I might have messed up a rebase, but we'll sort it out.

MelissaAutumn:
> 
> One thing to mention though: We might need a check if the CalDAV calendar URL is already there. E.g. I had on of 5 calendars from one principal already active and now I have the same calendar twice under calendars.
> 

It should lookup by calendar url, it's possible that the url differs by a single character causing the mismatch. I think this is fine especially since the majority of users won't ever add another calendar server.

MelissaAutumn:
Rebased the rebase...and that seems to have fixed it!

devmount (approved):
Awesome, thanks!

PR #727, opened 2024-10-30
Disable quick link update on inactive schedules

### LINKED ISSUE #686: Some fields are editable while schedule is deactivated

devmount:
## Description of the Change

The button for updating the quick link slug is now grayed out and inactive, as soon as the schedule becomes inactive.

image

## Applicable Issues

Closes #686

### DIFF
--- a/frontend/src/components/ScheduleCreation.vue
+++ b/frontend/src/components/ScheduleCreation.vue
@@ -716,7 +716,14 @@ watch(
               >
                 {{ t("label.quickLink") }}
               </text-input>
-              <refresh-icon class="mt-7 cursor-pointer text-teal-600" @click.prevent="refreshSlug" />
+              <refresh-icon
+                class="mt-7"
+                :class="{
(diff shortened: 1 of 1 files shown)


MelissaAutumn (approved):
Yay an appointment pr! 

Looks good! I would have also accepted a disabled state on the refresh icon, but I guess it is indeed just an icon. 🤔

devmount:
@MelissaAutumn Haha yes, I'm back after vacation 😇 

Good point, I can make that an actual button if you like?

MelissaAutumn:
That would be good if it's not too much more work!

devmount:
I found: We can simply use our LinkButton component with icons! Even supports getting disabled out of the box. If you want you can take a final look at it. Else I'll merge it in.

MelissaAutumn:
Looks good!

PR #729, opened 2024-10-30
Schedule page stores

### LINKED ISSUE #648: Hook up the schedule page to stores
The dashboard/schedule page is slow to load because it doesn't store its data inside a...store. It should pull schedules from the scheduleStore (while also refreshing it on page load), and remoteEvents from the calendarStore. We'll need to be careful about only storing the user's remote events in the calendar store and not other people's schedules, so we shouldn't touch the BookingView.

devmount:
## Description of the Change

This PR aims to improve performance of the dashboard (Schedule configuration page). The already existing schedule store got hooked up on the dashboard page. Still misses:

- [x] Adding remote event retrieval to the calendar store
- [x] Hook up stored events for the dashboard view

## Benefits

The dashboard now retrieves events per month, only if that month is navigated to the first time. The events will be remembered from this point making calendar navigation much faster and more convenient, when returning to views/months already visited.

## Applicable Issues

Closes #648

### DIFF
--- a/frontend/src/stores/calendar-store.ts
+++ b/frontend/src/stores/calendar-store.ts
@@ -1,9 +1,13 @@
-import { Calendar, CalendarListResponse, Fetch } from '@/models';
+import { Calendar, CalendarListResponse, Fetch, RemoteEvent, RemoteEventListResponse } from '@/models';
 import { defineStore } from 'pinia';
-import { ref, computed } from 'vue';
+import { ref, computed, inject } from 'vue';
+import { dayjsKey } from '@/keys';
+import { Dayjs } from 'dayjs';
 
 // eslint-disable-next-line import/prefer-default-export
 export const useCalendarStore = defineStore('calendars', () => {
+  const dj = inject(dayjsKey);
+
   // State
   const isLoaded = ref(false);
 
@@ -11,8 +15,12 @@ export const useCalendarStore = defineStore('calendars', () => {
   const calendars = ref<Calendar[]>([]);
   const unconnectedCalendars = computed((): Calendar[] => calendars.value.filter((cal) => !cal.connected));
   const connectedCalendars = computed((): Calendar[] => calendars.value.filter((cal) => cal.connected));
-
   const hasConnectedCalendars = computed(() => connectedCalendars.value.length > 0);
+
+  // List of remote events. Retrieved in batches per month.
+  const remoteEvents = ref<RemoteEvent[]>([]);
+  // List of month batches already called.
+  const remoteMonthsRetrieved = ref<string[]>([]);
 
   const connectGoogleCalendar = async (call: Fetch, email: string) => {
     const urlFriendlyEmail = encodeURIComponent(email);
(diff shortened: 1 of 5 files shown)


frontend/src/views/ScheduleView.vue
```diff
@@ -117,6 +81,7 @@ onMounted(async () => {
   // Don't actually load anything during the FTUE
   if (route.name === 'setup') {
     // Setup a fake schedule so the schedule creation bar works correctly...
+    // TODO: move that to the schedule store as initial/default value
```
devmount, line 84: @MelissaAutumn I think it would be more convenient to have the initial schedule object in the store instead of here. If you agree, I can move it.

frontend/src/views/ScheduleView.vue
```diff
@@ -204,7 +167,7 @@ const dismiss = () => {
         :calendars="connectedCalendars"
         :schedule="firstSchedule"
         :active-date="activeDate"
-        @created="getFirstSchedule"
+        @created="scheduleStore.fetch(call, true)"
```
devmount, line 170: We're forcing a re-fetch when the schedule is first created.

frontend/src/stores/calendar-store.ts
```diff
@@ -44,6 +52,48 @@ export const useCalendarStore = defineStore('calendars', () => {
...
 
+  /**
+   * Get all cremote events from connected calendars for given time span
+   * @param call preconfigured API fetch function
+   * @param activeDate Dayjs object defining the current month
+   * @param force force a refetch
+   */
+  const getRemoteEvents = async (call: Fetch, activeDate: Dayjs, force = false) => {
+    // Get month identifier to remember this month's events are already retrieved
+    const month = activeDate.format('YYYY-MM');
```
devmount, line 63: I just kept our decision to retrieve events on a monthly basis. But if you wish I can make that configurable as well.

frontend/src/views/ScheduleView.vue
MelissaAutumn: That sounds like a good idea!

frontend/src/stores/calendar-store.ts
MelissaAutumn: Do we have a define for that date format?

frontend/src/stores/calendar-store.ts
```diff
@@ -44,6 +52,48 @@ export const useCalendarStore = defineStore('calendars', () => {
...
+    const to = activeDate.endOf('month').add(1, 'day').format('YYYY-MM-DD');
+  
+    // If retrieval is forced, delete cache and start with zero events again
+    if (force) {
+      remoteMonthsRetrieved.value = [];
+    }
+    const calendarEvents = force ? [] : [...remoteEvents.value];
+
+    // Only retrieve remote events if we don't have this month already cached
+    if (!remoteMonthsRetrieved.value.includes(month)) {
```
MelissaAutumn, line 76: Can we flip this so that it's like:
```suggestion
    if (remoteMonthsRetrieved.value.includes(month)) {
        return;
    }
    ... existing code but here now ...
```

As I don't see why we would need to re-push an already cached month, but let me know if I've misread this function!

frontend/src/stores/schedule-store.ts
```diff
@@ -43,18 +44,18 @@ export const useScheduleStore = defineStore('schedules', () => {
   };
 
   /**
-   * Restore default state, empty and unload calendars
+   * Restore default state, empty and unload schedules
```
MelissaAutumn, line 47: 😅 Thanks!

MelissaAutumn (commented):
Few questions

frontend/src/stores/calendar-store.ts
devmount: Whoops, you're totally right 😅

frontend/src/stores/calendar-store.ts
devmount: No, but now we have. Thanks for this suggestion!

frontend/src/views/ScheduleView.vue
devmount: Done.

devmount:
Thank you for your feedback! Implemented all suggestions now.

MelissaAutumn (approved):
LGTM!

PR #734, opened 2024-11-05
Make all test time objects timezone aware

### LINKED ISSUE #732: Tests failing due to timeshift?
It's now Standard time, and it gets dark early now woo?, yay?, etc...

Anywho this seems to break a test which shouldn't be broken afaik. But also maybe this test was always sort of broken. 
https://github.com/thunderbird/appointment/actions/runs/11689313125/job/32551813941

MelissaAutumn:
Fixes #732 

Time objects were naive (not timezone aware) and I think that was skewing the tests during Daylight savings. Either way the values are now actually correct, so we'll see if this breaks again in about 4 or so months 😅


devmount (approved):
Thanks for this! I hope we don't have more places with timezone naive dates 😅

PR #737, opened 2024-11-05
Convert schedule's time-range based on the date it was updated. 

### LINKED ISSUE #736: Schedule's time is not converted correctly in dashboard
The schedule's start and end time are not converted correctly from UTC. Due to the wonders of daylight savings we calculate the utc offset of those times based on the schedule's last update time. 

So if someone saves their schedule with a time of 9-5 in PDT (-7) it currently turns it into 8-4 in PST (-8). Instead we should use the date of `time_updated` to calculate the UTC offset to apply to the time range. Then the time range will always be 9-5 (in this example.) 

Here's an example of how it's done:
https://github.com/thunderbird/appointment/blob/main/backend/src/appointment/database/models.py#L305

MelissaAutumn:
Fixes #736

Time is confusing...Details are in the ticket.

### DIFF
--- a/frontend/src/components/ScheduleCreation.vue
+++ b/frontend/src/components/ScheduleCreation.vue
@@ -122,14 +122,8 @@ onMounted(() => {
   if (props.schedule) {
     scheduleInput.value = { ...props.schedule };
     // calculate utc back to user timezone
-    scheduleInput.value.start_time = dj(`${dj().format(dateFormat)}T${scheduleInput.value.start_time}:00`)
-      .utc(true)
-      .tz(user.data.timezone ?? dj.tz.guess())
-      .format('HH:mm');
-    scheduleInput.value.end_time = dj(`${dj().format(dateFormat)}T${scheduleInput.value.end_time}:00`)
-      .utc(true)
-      .tz(user.data.timezone ?? dj.tz.guess())
-      .format('HH:mm');
+    scheduleInput.value.start_time = scheduleStore.timeToFrontendTime(scheduleInput.value.start_time, scheduleInput.value.time_updated);
+    scheduleInput.value.end_time = scheduleStore.timeToFrontendTime(scheduleInput.value.end_time, scheduleInput.value.time_updated);
 
     // Adjust the default calendar if the one attached is not connected.
     const { calendar_id: calendarId } = scheduleInput.value;
--- a/frontend/src/stores/schedule-store.ts
+++ b/frontend/src/stores/schedule-store.ts
@@ -1,6 +1,8 @@
 import { i18n } from '@/composables/i18n';
 import { defineStore } from 'pinia';
-import { ref, computed, inject, Ref } from 'vue';
+import {
+  ref, computed, inject, Ref,
+} from 'vue';
 import { useUserStore } from '@/stores/user-store';
 import {
   DateFormatStrings,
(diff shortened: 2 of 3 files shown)


frontend/src/stores/schedule-store.ts
```diff
@@ -191,13 +194,14 @@ export const useScheduleStore = defineStore('schedules', () => {
...
+   * BaseTime is needed for daylight savings awareness. If your time is coming from the backend's schedule route
+   * then ensure you pass in schedule.time_updated as the baseTime!
    */
-  const timeToFrontendTime = (time) => {
+  const timeToFrontendTime = (time: string, baseTime: string) => {
     const dateFormat = DateFormatStrings.QalendarFullDay;
     const user = useUserStore();
 
-    return dj(`${dj().format(dateFormat)}T${time}:00`)
+    return dj(`${dj(baseTime).format(dateFormat)}T${time}:00`)
```
devmount, line 204: How does this change anything, if `time` is added separately here? Sorry, I guess I too dumb to see it atm 😂

MelissaAutumn: Sorry I should have called it baseDate, but it inits the dj instance as baseTime's datetime object. And then formats it out to a string, but part of the string format is the different time...from the time variable.

I hope that made sense 😅

frontend/src/components/ScheduleCreation.vue
```diff
@@ -122,14 +122,8 @@ onMounted(() => {
...
-    scheduleInput.value.start_time = dj(`${dj().format(dateFormat)}T${scheduleInput.value.start_time}:00`)
-      .utc(true)
-      .tz(user.data.timezone ?? dj.tz.guess())
-      .format('HH:mm');
-    scheduleInput.value.end_time = dj(`${dj().format(dateFormat)}T${scheduleInput.value.end_time}:00`)
-      .utc(true)
-      .tz(user.data.timezone ?? dj.tz.guess())
-      .format('HH:mm');
+    scheduleInput.value.start_time = scheduleStore.timeToFrontendTime(scheduleInput.value.start_time, scheduleInput.value.time_updated);
+    scheduleInput.value.end_time = scheduleStore.timeToFrontendTime(scheduleInput.value.end_time, scheduleInput.value.time_updated);
```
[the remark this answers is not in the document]
MelissaAutumn: I was going a little crazy trying to figure out why the store function wasn't working on the dashboard 😄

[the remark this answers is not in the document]
devmount: Haha yeah I feel that

frontend/src/stores/schedule-store.ts
devmount: Yes, I just thought it wouldn't make any difference since we're doing `.format('HH:mm');` in the end anyway 😅

devmount (approved):
Works so far on my end, I just couldn't properly provoke the daylight savings time issue 😇

frontend/src/stores/schedule-store.ts
MelissaAutumn: Daylight savings works differently in every single datetime library, and that keeps me up at night 🫠

devmount: Yeah. Somebody recently said, daylight saving creates more problems then it solves. I found that to be absolutely true. 😅

PR #740, opened 2024-11-14
increased container size in prod

jdbass:
Description of the Change

Increases backend container size in prod to 1vcpu/4GB while keeping stage on 0.5vcpu/1GB

## Benefits

More resources to support increased user base as we expand the beta


MelissaAutumn (approved):
looks good, thanks!

PR #744, opened 2024-11-16
Validate booking slot's weekday in schedule timezone. (Fixes #735)

### LINKED ISSUE #735: "The time slot you have selected could not be found" error
**Describe the bug**
SlotNotFound error was triggered on a seemingly valid booking calendar. We're not sure how this was triggered, and errors like these are [caught by FastAPI](https://github.com/thunderbird/appointment/blob/main/backend/src/appointment/routes/schedule.py#L202) and not surfaced to Sentry since it's not a fatal state.

error

I'm logging this here so we can keep track of this particular case, and eventually come back to think about how to track API exceptions. Sentry may or may not be the right place, but we're unable to diagnose the cause of this, so ideally we should have some kind of telemetry here.

MelissaAutumn:
Fixes #735 

Fixes a case where the date could overflow to a day of the week which the schedule has disabled. Which would fail booking slot validation. 

This is because days of the week are saved based on the user/schedule's timezone rather than UTC. Which is a little confusing and we should maybe look into fixing that, but this is a quick fix for now.

### DIFF
--- a/backend/src/appointment/routes/schedule.py
+++ b/backend/src/appointment/routes/schedule.py
@@ -1,3 +1,5 @@
+import zoneinfo
+
 import sentry_sdk
 from fastapi import APIRouter, Depends, BackgroundTasks, Request
 import logging
@@ -48,6 +50,7 @@ router = APIRouter()
 limiter = Limiter(key_func=get_remote_address)
 
 
+
 def is_this_a_valid_booking_time(schedule: models.Schedule, booking_slot: schemas.SlotBase) -> bool:
     """Checks for timezone correctness, weekday and start/end time validity."""
     # For now lets only accept utc bookings as that's what our frontend supplies us.
@@ -62,7 +65,9 @@ def is_this_a_valid_booking_time(schedule: models.Schedule, booking_slot: schema
         return False
 
     # Is the time requested on a day of the week they have disabled?
-    iso_weekday = int(booking_slot.start.strftime('%u'))
+    # This should be based on the schedule timezone
+    start_date_tz = booking_slot.start.astimezone(zoneinfo.ZoneInfo(schedule.timezone))
+    iso_weekday = int(start_date_tz.strftime('%u'))
     if iso_weekday not in schedule.weekdays:
         return False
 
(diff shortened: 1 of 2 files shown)


backend/src/appointment/routes/schedule.py
```diff
@@ -73,10 +78,9 @@ def is_this_a_valid_booking_time(schedule: models.Schedule, booking_slot: schema
 
     booking_slot_end = booking_slot.start + timedelta(minutes=schedule.slot_duration)
 
-    if (
-        booking_slot.start < datetime.combine(today, schedule.start_time, tzinfo=timezone.utc)
-        or booking_slot_end > datetime.combine(today, schedule.end_time, tzinfo=timezone.utc) + timedelta(days=add_day)
-    ):
+    if booking_slot.start < datetime.combine(
+        today, schedule.start_time, tzinfo=timezone.utc
+    ) or booking_slot_end > datetime.combine(today, schedule.end_time, tzinfo=timezone.utc) + timedelta(days=add_day):
```
devmount, line 83: Can we move this into separate vars to make it a little bit more readable? E.g.:

```suggestion
    too_early = booking_slot.start < datetime.combine(today, schedule.start_time, tzinfo=timezone.utc)
    too_late = booking_slot_end > (
      datetime.combine(today, schedule.end_time, tzinfo=timezone.utc) + timedelta(days=add_day)
    )
    if too_early or too_late:
```

backend/test/unit/test_utils.py
```diff
@@ -83,14 +83,41 @@ def test_bug_735(self, make_schedule):
...
-            is_valid = is_this_a_valid_booking_time(
-                schedule,
-                s_a.slot
-            )
+            is_valid = is_this_a_valid_booking_time(schedule, s_a.slot)
+
+        assert is_valid is True
+
+    def test_bug_735_case_2(self, make_schedule):
+        # Request data submitted from bug and anonymized.
```
devmount, line 96: Just a quick overview what this bug is about would be nice (since we don't have a descriptive name here), like:

```suggestion
        # Bug where a "Time slot not found" error was triggered incorrectly
        # Request data submitted from bug and anonymized.
```

devmount (approved):
Thank you for fixing this and providing a test 👏🏻  Just two minor suggestions.

PR #752, opened 2024-11-18
Small fix for auth unit test

rwood-moz:
Small fix in a couple of asserts in the auth unit test.


MelissaAutumn (approved):
This should indeed be .id == .id, so thanks! 

I guess we could also override __eq__ on our base model and return that .id == .id comparison, but best to be explicit like you've done here.

devmount (commented):
Thank you, Rob!

PR #764, opened 2024-11-22
Fallback to schedule owner timezone

### LINKED ISSUE #763: Can't book on calendar
**Describe the bug**
Have sent invite, recipient unable to book time on my calendar.

Had @devmount review logs. Appears there is no timezone associated with my schedule or it can't be retrieved by the time of the booking request.

https://thunderbird.sentry.io/issues/6080449546/?project=4505428124827648

Image

devmount:
## Description of the Change

This change retrieves the schedule owner's timezone if the timezone of the schedule itself is None.

## Benefits

This handles a case were schedules, that were created earlier than the schedule.timezone field exists, can still be booked.

## Applicable Issues

Fixes #763

### DIFF
--- a/backend/src/appointment/routes/schedule.py
+++ b/backend/src/appointment/routes/schedule.py
@@ -66,7 +66,8 @@ def is_this_a_valid_booking_time(schedule: models.Schedule, booking_slot: schema
 
     # Is the time requested on a day of the week they have disabled?
     # This should be based on the schedule timezone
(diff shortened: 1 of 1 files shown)


malini:
from Matrix discussion, we should instead update our db with the owner's tz if the schedule's timezone is null and fix the data consistency caused by [this change](https://github.com/thunderbird/appointment/commit/8094904cbf72bce00bfe62cd432037719d451bb3#diff-1d810ba22035665e8ce68c193c11abbec1fdc754452ef1f1304f9b0592604c02R69), rather than introducing a check in code for bad data.

PR #768, opened 2024-11-25
Hold invites

### LINKED ISSUE #459: Send a "HOLD:" meeting invite to scheduling party until the meeting is confirmed.
Had someone schedule time on my calendar recently through Appointment and they noted that it would have been nice to have an invite sent to them to hold that time open until I confirmed the meeting. I agree. You could be waiting for confirmation and fill that slot without thinking and then end up with a conflict.

Some examples: We could simply make the title for the event "HOLD: Appointment - Ryan Sipes and (other party)" or an alternative could be "AWAITING CONFIRMATION: Appointment - Ryan Sipes and (other party)"

The invite could be updated with a new title (removing "HOLD:") once the event is confirmed.

devmount:
## Description of the Change

This change implements the [previously discussed](https://github.com/thunderbird/appointment/issues/459#issuecomment-2492915805) hold event flow.

This still needs heavy testing for both: Google and CalDAV.

## Applicable Issues

Fixes #459

### DIFF
--- a/backend/src/appointment/controller/apis/google_client.py
+++ b/backend/src/appointment/controller/apis/google_client.py
@@ -103,6 +103,8 @@ class GoogleClient:
         # Limit the fields we request
         fields = ','.join(
             (
+                'items/id',
+                'items/iCalUID',
                 'items/status',
                 'items/summary',
                 'items/description',
--- a/backend/src/appointment/controller/apis/google_client.py
+++ b/backend/src/appointment/controller/apis/google_client.py
@@ -6,7 +6,7 @@ from googleapiclient.errors import HttpError
 from ...database import repo
 from ...database.models import CalendarProvider
 from ...database.schemas import CalendarConnection
-from ...exceptions.calendar import EventNotCreatedException
+from ...exceptions.calendar import EventNotCreatedException, EventNotDeletedException
 from ...exceptions.google_api import GoogleScopeChanged, GoogleInvalidCredentials
 
 
@@ -153,7 +153,15 @@ class GoogleClient:
         return response
 
     def delete_event(self, calendar_id, event_id, token):
-        pass
+        response = None
+        with build('calendar', 'v3', credentials=token, cache_discovery=False) as service:
+            try:
+                response = service.events().delete(calendarId=calendar_id, eventId=event_id).execute()
+            except HttpError as e:
+                logging.warning(f'[google_client.delete_event] Request Error: {e.status_code}/{e.error_details}')
+                raise EventNotDeletedException()
+
+        return response
 
     def sync_calendars(self, db, subscriber_id: int, token):
         # Grab all the Google calendars
--- a/backend/src/appointment/controller/apis/google_client.py
+++ b/backend/src/appointment/controller/apis/google_client.py
@@ -141,13 +141,13 @@ class GoogleClient:
 
         return items
 
-    def create_event(self, calendar_id, body, token):
+    def save_event(self, calendar_id, body, token):
         response = None
         with build('calendar', 'v3', credentials=token, cache_discovery=False) as service:
             try:
                 response = service.events().import_(calendarId=calendar_id, body=body).execute()
             except HttpError as e:
-                logging.warning(f'[google_client.create_event] Request Error: {e.status_code}/{e.error_details}')
+                logging.warning(f'[google_client.save_event] Request Error: {e.status_code}/{e.error_details}')
                 raise EventNotCreatedException()
 
         return response
--- a/backend/src/appointment/controller/calendar.py
+++ b/backend/src/appointment/controller/calendar.py
@@ -504,6 +504,7 @@ class Tools:
         cal = Calendar()
         cal.add('prodid', '-//Thunderbird Appointment//tba.dk//')
         cal.add('version', '2.0')
+        cal.add('method', 'REQUEST')
         org = vCalAddress('MAILTO:' + organizer.preferred_email)
         org.params['cn'] = vText(organizer.preferred_email)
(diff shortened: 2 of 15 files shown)


backend/src/appointment/controller/apis/google_client.py
```diff
@@ -141,19 +141,27 @@ def list_events(self, calendar_id, time_min, time_max, token):
 
         return items
 
-    def create_event(self, calendar_id, body, token):
+    def save_event(self, calendar_id, body, token):
```
devmount, line 144: I renamed this cause we're now using it for creating and updating events.

backend/src/appointment/controller/apis/google_client.py
```diff
@@ -141,19 +141,27 @@ def list_events(self, calendar_id, time_min, time_max, token):
...
                 raise EventNotCreatedException()
 
         return response
 
     def delete_event(self, calendar_id, event_id, token):
-        pass
+        response = None
+        with build('calendar', 'v3', credentials=token, cache_discovery=False) as service:
+            try:
+                response = service.events().delete(calendarId=calendar_id, eventId=event_id).execute()
```
devmount, line 159: For some reason, this throws:

> [google_client.delete_event] Request Error: 404/[{'domain': 'global', 'reason': 'notFound', 'message': 'Not Found'}

Still investigating.

backend/src/appointment/controller/calendar.py
```diff
@@ -501,7 +516,7 @@ def create_vevent(
             slot.start.replace(tzinfo=timezone.utc) + timedelta(minutes=slot.duration),
         )
         event.add('dtstamp', datetime.now(UTC))
-        event.add('status', 'CONFIRMED')
+        event.add('status', 'NEEDS-ACTION' if on_hold else 'CONFIRMED')
```
devmount, line 519: This needs to be tested directly in Thunderbird.

backend/src/appointment/controller/calendar.py
```diff
@@ -840,3 +882,9 @@ def fix_caldav_urls(url: str) -> str:
             url = 'https://api.googlecontent.com/caldav/v2/'
 
         return url
+
+    @staticmethod
+    def default_event_title(slot: schemas.Slot, owner: schemas.Subscriber, prefix='') -> str:
```
devmount, line 887: TODO: add some comments

backend/src/appointment/database/repo/slot.py
```diff
@@ -114,11 +114,3 @@ def delete(db: Session, slot_id: int):
...
     db.commit()
     return db_slot
-
-
-def is_available(db: Session, slot_id: int):
-    """check if slot is still available for booking"""
-    slot = get(db, slot_id)
-    if slot.schedule:
-        return slot and slot.booking_status == models.BookingStatus.requested
-    return False
```
devmount, line 124: This isn't needed anymore since slots are now always unavailable as soon as they're booked.

devmount (commented):
@MelissaAutumn This is still wip, but feel free to already take a look at it. Except for the remote event deletion, everythings already working fine on Google and CalDAV on my end.

backend/src/appointment/controller/apis/google_client.py
MelissaAutumn: That's very odd. We have the permissions to do so, so it shouldn't be that.

backend/src/appointment/controller/calendar.py
devmount: Whoops, this was supposed to be `TENTATIVE`, but we can keep `CONFIRMED` as well. I don't think this is needed for the prompt to update. According to RFC 5546 and [Philipp](https://github.com/thunderbird/appointment/issues/459#issuecomment-2491213535), there needs to be a METHOD:REQUEST. I'll add that too.

backend/src/appointment/controller/apis/google_client.py
MelissaAutumn: Will do!

MelissaAutumn:
Fixed the issue. It turns out Google needs their internal id field to delete events, and not the one we create via their import method. So once we save an event it returns the event object back to us, so I just store it on appointment.external_id (new field!). CalDAV doesn't need to use this field, so it falls back to uuid. 

Also discovered that it may fail if the user deletes their event before they reject the request. I'm not sure if we want to handle that any differently thank a "link is invalid" error. Additionally I discovered if you switch the schedule's calendar, delete and update requests were being attempted on the schedule's calendar and not the appointment's calendar (the one it's actually on.) so I patched that up. 

Feel free to give it a try again to see if it all works on your end now!

backend/src/appointment/controller/apis/google_client.py
devmount: Tested, works now 🎉

devmount (commented):
Excellent work @MelissaAutumn 👏🏻  Google event deletion on denial works flawlessly now on my end too!

> Also discovered that it may fail if the user deletes their event before they reject the request. I'm not sure if we want to handle that any differently thank a "link is invalid" error.

I think it's fine to keep that error for now and see, if we want to maybe make that more specific in the future if we need it.

> Additionally I discovered if you switch the schedule's calendar, delete and update requests were being attempted on the schedule's calendar and not the appointment's calendar (the one it's actually on.) so I patched that up.

Oh, thank you! Didn't think of that particular case. Maybe we should also inform the user about open bookings when switching calendars on the schedule page or disabling them in the settings? Just as idea to improve the UX at that point, but out of scope here.

Imo we can merge this now.

MelissaAutumn (approved):

PR #769, opened 2024-11-25
Prefill schedule timezone

### LINKED ISSUE #763: Can't book on calendar

MelissaAutumn:
Fixes #763 

Also I had to move normalize_secrets as for some reason in my vm a starlette package was trying to import our secrets.py file. 🤔 Utils probably works for that.

But this change adds a data migration that will set the timezone of the schedule to the subscriber's timezone. If the subscriber doesn't have a timezone (rare, and I think we've corrected these cases already) it sets them to UTC, and then sets the schedule to UTC.


devmount (approved):
Thank you, this works!

PR #774, opened 2024-11-27
Replace the content-type on ics files

### LINKED ISSUE #773: Add missing itip support
Thunderbird currently doesn't pick up our ics files. It seems like we're missing a method in the email attachment header.
https://www.rfc-editor.org/rfc/rfc2446#section-3.2

We recently added method REQUEST to our ics files, so let's do the same for the email attachment header. 
Adjusting "Content-Type: text/calendar" -> "Content-Type: text/calendar; charset="UTF-8"; method=REQUEST"

MelissaAutumn:
Should fix #773 

I can't really test it in thunderbird, but the header shows up in mailpit. 

I'm not sure if there's a better way to fix this header, the python api seems a little...light on implementation details outside of "Hey add an attachment to an email and send it!" 🤔

### DIFF
--- a/backend/src/appointment/controller/mailer.py
+++ b/backend/src/appointment/controller/mailer.py
@@ -94,14 +94,24 @@ class Mailer:
 
         # add attachment(s) as multimedia parts
         for a in self._attachments():
-            # Attach it to the html payload
-            message.get_payload()[1].add_related(
-                a.data,
-                a.mime_main,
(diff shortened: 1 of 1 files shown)


devmount (approved):
Wow, that's nice! Works on my end:

image

There is an option at the top right in Mailpit to save the email as raw. I just dragged that file into Thunderbird et voila!

PR #776, opened 2024-11-28
Declined booking event handling

### LINKED ISSUE #775: Send an updated ics with on declined booking emails
Currently the declined booking doesn't include an updated ics file. We should generate one with the method=CANCEL (in ics and email attachment header) to notify the user's calendar.
https://www.rfc-editor.org/rfc/rfc2446#section-3.2.5

devmount:
## Description of the Change

This change adds an attachment to the email that informs the user, that their booking got rejected. This enables the user to simply click delete and update their calendar to remove the hold event.

image

## Applicable Issues

Closes #775


MelissaAutumn (approved):
Thanks!

PR #784, opened 2024-12-04
Add data-testid attributes

rwood-moz:
Adding some `data-testid` attributes to some UI DOM elements that may be used by future E2E tests.

### DIFF
--- a/frontend/src/views/SettingsView.vue
+++ b/frontend/src/views/SettingsView.vue
@@ -94,7 +94,7 @@ onMounted(() => {
         "
         :class="{ '!bg-teal-500 !text-white': view === activeView }"
         @click="show(key)"
-        :data-testid=key
+        :data-testid="'settings-' + key + '-settings-btn'"
       >
         <span>{{ t('label.' + key) }}</span>
         <icon-chevron-right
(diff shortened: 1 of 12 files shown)


MelissaAutumn (approved):
Thank you!

Is there a general convention for these like "input-div-wrapper" and "input-div-wrapper-input" is something I'm seeing here? Or is it mostly just pick and choose?

frontend/src/views/SettingsView.vue
```diff
@@ -94,6 +94,7 @@ onMounted(() => {
         "
         :class="{ '!bg-teal-500 !text-white': view === activeView }"
         @click="show(key)"
+        :data-testid=key
```
devmount, line 97: Is there a reason to not using quotes here? I think we always use them even on single word attributes.

```suggestion
        :data-testid="key"
```

devmount: Oh and the term `key` might be a bit too general as an identifier? But I'm really no expert here, if that is sufficient that's totally fine.

devmount (approved):
Looking good! Just wondering the same thing like Melissa, if we should follow a certain convention for the test id naming.

Thank you for taking care of this!!

rwood-moz:
Appreciate your feedback @MelissaAutumn and @devmount! PR is updated and ready for re-review when you have some time, thanks!

devmount (approved):
Awesome, thanks for the changes! Can be merged imo

PR #796, opened 2024-12-17
Add owner email as reply-to header

### LINKED ISSUE #765: Replying to Appointment Invites currently fills in 'appointment-support@thunderbird.net'
**Describe the bug**
When someone receives emails pertaining to bookings, the sender is no-reply@appointment.day, but if you try to reply, it goes to appointment-support@thunderbird.net

I don't think we should do that behaviour, we have seen a booking user try to reply to an "Invitation sent from Thunderbird Appointment" subject email, with a message intended for the Appointment user's email.

Instead, I think we should either have replies go to no-reply@ or the Appointment user's email.

I'm in favour of the latter, because I think that's what people may think might happen, and that is what Calendly does, so there is some level of user expectation that they should be able to directly reply to an invite and that email will go to the booking calendar's owner.

devmount:
## Description of the Change

This PR changes the default _reply-to_ (configured system or support email address) for the invitation mail (x has accepted your booking) and the new booking mail (x has just booked...).

## Benefits

The configuration is now like this:

- invitation mail (_... has accepted your booking_)
  - To: Bookee
  - Reply-To: Owner
- new booking mail (_... has just booked..._)
  - To: Owner
  - Reply-To: Bookee

## Applicable Issues

Closes #765

### DIFF
--- a/backend/src/appointment/controller/mailer.py
+++ b/backend/src/appointment/controller/mailer.py
@@ -223,6 +223,7 @@ class InvitationMail(BaseBookingMail):
             'plain': l10n('invite-mail-plain'),
         }
         super().__init__(*args, **default_kwargs, **kwargs)
+        self.reply_to = self.email
 
     def html(self):
         return get_template('invite.jinja2').render(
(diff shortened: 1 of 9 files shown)


MelissaAutumn:
I think we'll need to replace this text as well:

image

Maybe remove the "reply to this email" part for the "booking confirmation email" and add a "Need to get in touch with Full Name <full_name@example.org>? Simply reply to this email."

MelissaAutumn:
Looks good! I've just split it into two lines since it looks a little cleaner. 
image

Might want to double check the german localization though!

MelissaAutumn (approved):

devmount:
That's nice, thank you!

PR #798, opened 2024-12-18
Check for slot status before confirming or rejecting a booking request

### LINKED ISSUE #791: You can confirm a booking as many times as you want
Currently if you keep clicking "Confirm" in the appointment confirmation email you can send as many invite emails as you want. We should probably limit this to one time. 

This does not work for declining requests from your email.

No extra appointments are created, it just sends the email.

devmount:
## Description of the Change
This PR adds a check for slot status before confirming or rejecting a booking request.

## Benefits
Prevents booking requests from being answered multiple times, each time sending a confirmation/rejection email to the bookee.

## Applicable Issues
Closes #791

### DIFF
--- a/backend/test/integration/test_schedule.py
+++ b/backend/test/integration/test_schedule.py
@@ -905,8 +905,7 @@ class TestDecideScheduleAvailabilitySlot:
             headers=auth_headers,
         )
 
-        with with_db() as db:
-            assert response.status_code == 404, response.text
+        assert response.status_code == 404, response.text
 
 
     def test_deny(
@@ -983,5 +982,4 @@ class TestDecideScheduleAvailabilitySlot:
             headers=auth_headers,
         )
 
-        with with_db() as db:
-            assert response.status_code == 404, response.text
+        assert response.status_code == 404, response.text
--- a/backend/test/integration/test_schedule.py
+++ b/backend/test/integration/test_schedule.py
@@ -698,6 +698,7 @@ class TestRequestScheduleAvailability:
                 headers=auth_headers,
             )
             assert response.status_code == 404, response.text
+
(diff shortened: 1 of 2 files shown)


MelissaAutumn (approved):
Thanks! Can you add a test case too?

devmount (commented):
@MelissaAutumn Thank you. Of course, sorry, always forgetting about the tests 🙈 I extended two existing test cases.

backend/test/integration/test_schedule.py
```diff
@@ -893,6 +898,17 @@ def test_confirm(
...
 
+        # Now try to confirm the same event again
+        response = with_client.put(
+            '/schedule/public/availability/booking',
+            json=availability,
+            headers=auth_headers,
+        )
+
+        with with_db() as db:
+            assert response.status_code == 404, response.text
```
MelissaAutumn, line 909: I don't think this requires the with_db() as db line.

backend/test/integration/test_schedule.py
```diff
@@ -959,3 +975,13 @@ def test_deny(
...
+
+        # Now try to deny the same event again
+        response = with_client.put(
+            '/schedule/public/availability/booking',
+            json=availability,
+            headers=auth_headers,
+        )
+
+        with with_db() as db:
+            assert response.status_code == 404, response.text
```
MelissaAutumn, line 987: ditto

MelissaAutumn (approved):
Remove the with_db as db: line and it's good to merge!

backend/test/integration/test_schedule.py
devmount: That is undoubtedly true.

PR #800, opened 2024-12-18
Fix the selected calendar on the Dashboard schedule component. (Fixes #758)

### LINKED ISSUE #758: Dashboard shows the wrong calendar selected
Dashboard's schedule options will display the wrong calendar (always the first one in the list.) Behind the scenes it's still the calendar you selected and saved, so it's just a frontend error.

MelissaAutumn:
Turns out props.calendars is an array not an object!

Closes #758


devmount (approved):
Good catch, thank you!

PR #803, opened 2024-12-19
Add subscriber language and localize emails

### LINKED ISSUE #766: Appointment Confirmation Email not in my selected language
**Describe the bug**
Confirmation Email was in German. None of the other ones have been. I assume it was because the one who was booking is German?

**Expected behavior**
In my selected language, English.

**Actual behavior**
The appointment confirmation was in German.

**Screenshots**
Scheduler's email name and email removed.
Image

devmount:
## Description of the Change

This PR adds a new field `language` for the subscribers table and extends our l10n middleware for using a custom locale.
The tests were adjusted and extended accordingly.

## Benefits

Subscribers should see emails sent to them in their configured language now.

## Applicable Issues

Closes #766

### DIFF
--- a/backend/src/appointment/controller/mailer.py
+++ b/backend/src/appointment/controller/mailer.py
@@ -271,7 +271,8 @@ class ConfirmationMail(BaseBookingMail):
         self.confirmUrl = confirm_url
         self.denyUrl = deny_url
         self.schedule_name = schedule_name
-        default_kwargs = {'subject': l10n('confirm-mail-subject', {'name': name}, kwargs['lang'])}
+        lang = kwargs['lang'] if 'lang' in kwargs else None
+        default_kwargs = {'subject': l10n('confirm-mail-subject', {'name': name}, lang)}
         super().__init__(name=name, email=email, date=date, duration=duration, *args, **default_kwargs, **kwargs)
 
 
@@ -351,7 +352,8 @@ class NewBookingMail(BaseBookingMail):
            Reply-To: Bookee
         """
         self.schedule_name = schedule_name
-        default_kwargs = {'subject': l10n('new-booking-subject', {'name': name}, kwargs['lang'])}
+        lang = kwargs['lang'] if 'lang' in kwargs else None
+        default_kwargs = {'subject': l10n('new-booking-subject', {'name': name}, lang)}
         super(NewBookingMail, self).__init__(name=name, email=email, date=date, duration=duration, *args, **default_kwargs, **kwargs)
         self.reply_to = email
 
--- a/backend/src/appointment/controller/mailer.py
+++ b/backend/src/appointment/controller/mailer.py
@@ -56,7 +56,8 @@ class Mailer:
         html: str = '',
         plain: str = '',
         attachments: list[Attachment] = [],
-        method: str = 'REQUEST'
+        method: str = 'REQUEST',
+        lang: str = None
     ):
         self.sender = sender
         self.to = to
@@ -66,6 +67,7 @@ class Mailer:
         self.body_plain = plain
         self.attachments = attachments
         self.method = method
+        self.lang = lang
 
     def html(self):
         """provide email body as html per default"""
@@ -164,7 +166,9 @@ class Mailer:
 
 class BaseBookingMail(Mailer):
     def __init__(self, name, email, date, duration, *args, **kwargs):
-        """Base class for emails with name, email, and event information"""
+        """Base class for emails with name, email, and event information
+           Can have a different locale by providing a lang argument
+        """
         self.name = name
         self.email = email
         self.date = date
@@ -174,8 +178,8 @@ class BaseBookingMail(Mailer):
         super().__init__(*args, **kwargs)
 
         # Localize date and time range format
-        self.time_format = l10n('time-format')
-        self.date_format = l10n('date-format')
+        self.time_format = l10n('time-format', lang=self.lang)
+        self.date_format = l10n('date-format', lang=self.lang)
 
         # If value is key then there's no localization available, set a default.
         if self.time_format == 'time-format':
@@ -217,7 +221,10 @@ class BaseBookingMail(Mailer):
 
 class InvitationMail(BaseBookingMail):
     def __init__(self, *args, **kwargs):
-        """init Mailer with invitation specific defaults"""
+        """Init Mailer with invitation/booking-accepted specific defaults
+           To: Bookee
+           Reply-To: Event owner
+        """
         default_kwargs = {
             'subject': l10n('invite-mail-subject'),
             'plain': l10n('invite-mail-plain'),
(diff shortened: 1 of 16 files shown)


backend/src/appointment/controller/mailer.py
```diff
@@ -364,13 +384,17 @@ def html(self):
...
+           To: Support
+           Reply-To: Requestee
+        """
         self.requestee_name = requestee_name
         self.requestee_email = requestee_email
         self.topic = topic
         self.details = details
         default_kwargs = {'subject': l10n('support-mail-subject', {'topic': topic})}
         super(SupportRequestMail, self).__init__(os.getenv('SUPPORT_EMAIL', 'help@tb.net'), *args, **default_kwargs, **kwargs)
+        self.reply_to = requestee_email
```
devmount, line 397: This changed line is off-topic, sorry. But I thought it would be very convenient, to be able to just reply to a support request. Let me know if this is not wanted.

backend/src/appointment/l10n.py
```diff
@@ -1,10 +1,18 @@
 from typing import Union, Dict, Any
 
 from starlette_context import context, errors
+from .middleware.l10n import get_fluent
 
 
-def l10n(msg_id: str, args: Union[Dict[str, Any], None] = None) -> str:
-    """Helper function to automatically call fluent.format_value from context"""
+def l10n(msg_id: str, args: Union[Dict[str, Any], None] = None, lang: str = None) -> str:
```
devmount, line 7: The l10n helper now accepts a third parameter for a custom language, independent from the request.

backend/src/appointment/middleware/l10n.py
```diff
@@ -5,6 +5,21 @@
...
+    # Make sure our fallback locale is always in locales
+    if FALLBACK_LOCALE not in locales:
+        locales.append(FALLBACK_LOCALE)
+
+    base_url = 'src/appointment/l10n'
+
+    loader = FluentResourceLoader(f'{base_url}/{{locale}}')
+    fluent = FluentLocalization(locales, ['main.ftl', 'email.ftl', 'fields.ftl'], loader)
+
+    return fluent.format_value
```
devmount, line 20: I thought it's better to let this live directly in the module, but let me know if you prefer it to be a `@staticmethod`

backend/src/appointment/templates/email/includes/base.jinja2
```diff
@@ -8,7 +8,7 @@
 {% set colour_tbpro_apmt_primary = '#008080' %}
 {% set colour_tbpro_apmt_primary_hover = '#066769' %}
 {% set colour_tbpro_apmt_secondary = '#81D4B5' %}
-<html lang="{{ l10n('locale') }}" style="
+<html lang="{{ l10n('locale', lang=lang if lang else None) }}" style="
```
devmount, line 11: Since we need to translate the base (and footer) templates too, but not all emails require a custom locale, I did implement it like this. If you have a better idea, please tell!

backend/src/appointment/controller/mailer.py
```diff
@@ -258,11 +265,14 @@ def text(self):
...
     def __init__(self, confirm_url, deny_url, name, email, date, duration, schedule_name, *args, **kwargs):
-        """init Mailer with confirmation specific defaults"""
+        """Init Mailer with action-required:confirm/deny specific defaults
+           To: Event owner
+        """
         self.confirmUrl = confirm_url
         self.denyUrl = deny_url
         self.schedule_name = schedule_name
-        default_kwargs = {'subject': l10n('confirm-mail-subject', {'name': name})}
+        lang = kwargs['lang'] if 'lang' in kwargs else None
```
MelissaAutumn, line 274: Do we need to do this? kwargs is passed through to the constructor (and BaseBookingMail's constructor.) and it doesn't seem like it's explicitly used in the constructor elsewhere.

backend/src/appointment/middleware/l10n.py
MelissaAutumn: This is fine, it took me a minute to remember how the context system is setup 😅.

backend/src/appointment/migrations/versions/2024_12_18_1226-b398005a40e7_add_language_to_subscriber_table.py
```diff
@@ -0,0 +1,37 @@
...
+depends_on = None
+
+
+def upgrade() -> None:
+    # Add language column to subscribers table
+    op.add_column('subscribers', sa.Column('language', models.encrypted_type(sa.String), index=True))
+
+    # Prefill new column with default value
+    session = Session(op.get_bind())
+    subscribers: list[models.Subscriber] = session.query(models.Subscriber).all()
```
MelissaAutumn, line 28: You'll have to confirm but having a default value and nullable=False should remove the need for this.

MelissaAutumn (changes requested):
Some small changes requested. But a big thank you for getting this done!

backend/src/appointment/middleware/l10n.py
devmount: yeah, me too 😅

backend/src/appointment/migrations/versions/2024_12_18_1226-b398005a40e7_add_language_to_subscriber_table.py
```diff
@@ -0,0 +1,37 @@
...
+# revision identifiers, used by Alembic.
+revision = 'b398005a40e7'
+down_revision = 'd791a3f0e478'
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # Add language column to subscribers table
+    op.add_column('subscribers', sa.Column('language', models.encrypted_type(sa.String), index=True))
```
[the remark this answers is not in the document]
devmount: Of course.

backend/src/appointment/migrations/versions/2024_12_18_1226-b398005a40e7_add_language_to_subscriber_table.py
devmount: I thought so too, in fact I first tried it with default values. But it didn't work, the language column would always stay empty... I thought maybe it's because of the encrypted type? Maybe you have another idea?

backend/src/appointment/controller/mailer.py
devmount: I thought it would be a good idea to not require lang to be in the kwargs, but yeah, I see it makes sense that it's there. I will update it.

backend/src/appointment/migrations/versions/2024_12_18_1226-b398005a40e7_add_language_to_subscriber_table.py
devmount: I switched to db defaults, you can test the branch again, if you like

MelissaAutumn:
I appear to have broken every test. Oops.

devmount:
Nice catch with the server_default, didn't think of that! But what's up with the tests now 😂

MelissaAutumn:
Will roll back to default= as server_default expects an encrypted string because we encrypt the field. We also can't encrypt it at the model definition time because the env / secret key isn't available then 🤔

MelissaAutumn (approved):
Perfect, thanks for your hard work here!

PR #805, opened 2024-12-19
Adjust CALDAV.test_connection to return a boolean or raise an exception with a message. (Fixes #802)

### LINKED ISSUE #802: Provide caldav connection test errors to the end-user
Currently we just catch the error and return false on this function. But we might want to have the function return a Str | None. With a Str being an error and None meaning the connection was sucessful.

Once we have that we have RemoteCalendarConnectionError hold an error message for the frontend to use and show. 

https://github.com/thunderbird/appointment/blob/731c6d7fe4493bb8b076855518125915a3928a0e/backend/src/appointment/controller/calendar.py#L336

MelissaAutumn:
This hopefully gives the user a little more feedback on what went wrong with their connection attempt.

## Screenshots

image
image

### DIFF
--- a/backend/src/appointment/l10n/en/main.ftl
+++ b/backend/src/appointment/l10n/en/main.ftl
@@ -19,6 +19,7 @@ health-bad = Health BAD
 ## General Exceptions
 
 unknown-error = An unknown error occurred. Please try again later.
+unknown-error-short = an unknown error
 
 appointment-not-found = The appointment could not be found.
 calendar-not-found = The calendar could not be found.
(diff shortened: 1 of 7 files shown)


frontend/src/tbpro/elements/NoticeBar.vue
```diff
@@ -51,6 +51,7 @@ const isError = props.type === 'error';
   font-size: 0.8125rem;
   font-weight: 700;
   line-height: 1;
+  white-space: pre;
```
MelissaAutumn, line 54: For our sweet newline character in the localization message

backend/src/appointment/l10n/en/main.ftl
```diff
@@ -45,7 +46,14 @@ not-in-allow-list = Your email is not in the allow list.
 
 schedule-not-active = The schedule has been turned off. Please contact the schedule owner for more information.
 
-remote-calendar-connection-error = The remote calendar could not be reached. Please verify your connection information and try again.
+remote-calendar-connection-error = The remote calendar could not be reached due to {$reason}.
+
+        Please verify your connection information and try again.
```
devmount, line 51: Is this intended to be indented?

frontend/src/tbpro/elements/NoticeBar.vue
devmount: Ah, it is intended. Thanks.

devmount (approved):
Thank you! I added the German translation.

PR #806, opened 2024-12-19
Replace HttpExceptions with RedirectResponses with login err codes. (Fixes #801)

### LINKED ISSUE #801: Provide a better error message for switching emails during fxa flow
Email mismatch is a uninformative message. We should change it to display:
"Error: You must login to <email> on the Mozilla Accounts login form."

or something like that.

MelissaAutumn:
Instead of displaying raw json or text, we now redirect them back to the login page with an error code. That error code is used to look up a proper error message.

This removes the confusing "Email mismatch" message and a few others.

## Screenshots
image
image

cc @malini for copy check.

### DIFF
--- a/backend/src/appointment/routes/auth.py
+++ b/backend/src/appointment/routes/auth.py
@@ -1,7 +1,6 @@
 import json
 import os
 import secrets
-import uuid
 from datetime import timedelta, datetime, UTC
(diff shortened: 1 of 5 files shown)


devmount (approved):
That's nice, thanks! I added German translation.

PR #807, opened 2024-12-20
Migrate class-based config to ConfigDict

### LINKED ISSUE #783: Resolve pytest warnings
We currently are being hit with a bunch of pydantic warnings per test. These are simply deprecation warnings like:

>warning : /Users/melissaautumn/Dev/appointment/backend/.venv/lib/python3.12/site-packages/pydantic/_internal/_config.py:284: PydanticDeprecatedSince20: Support for class-based `config` is deprecated, use ConfigDict instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.7/migration/
  warnings.warn(DEPRECATION_MESSAGE, DeprecationWarning)

And should be fairly easy to resolve. Once fixed, confirm that pytest displays 0 warnings in the summary.

devmount:
## Description of the Change

Migrating our current class-based model config to [ConfigDict](https://docs.pydantic.dev/2.10/api/config/#pydantic.config.ConfigDict). Also use `json_schema_extra`. Also use `.model_dump()` instead of `.dict()`.

## Benefits

No more deprecation warnings. Hopefully.

## Applicable Issues

Closes #783

### DIFF
--- a/backend/src/appointment/database/schemas.py
+++ b/backend/src/appointment/database/schemas.py
@@ -317,7 +317,7 @@ class Subscriber(SubscriberAuth):
     id: int
     calendars: list[Calendar] = []
     slots: list[Slot] = []
-    ftue_level: Optional[int] = Field(gte=0)
+    ftue_level: Optional[int] = Field(json_schema_extra={gte: 0})
 
 
 class SubscriberMeOut(SubscriberBase):
--- a/backend/src/appointment/database/schemas.py
+++ b/backend/src/appointment/database/schemas.py
@@ -9,7 +9,7 @@ from uuid import UUID
 from datetime import datetime, date, time, timezone, timedelta
 from typing import Annotated, Optional, Self
 
-from pydantic import BaseModel, Field, EmailStr, model_validator
+from pydantic import BaseModel, ConfigDict, Field, EmailStr, model_validator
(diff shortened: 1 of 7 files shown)


backend/src/appointment/database/schemas.py
```diff
@@ -152,15 +149,16 @@ class AvailabilityBase(BaseModel):
...
     id: int
     time_created: datetime | None = None
     time_updated: datetime | None = None
 
-    class Config:
-        from_attributes = True
-
 
 class ScheduleBase(BaseModel):
+    model_config = ConfigDict(json_encoders = { time: lambda t: t.strftime('%H:%M') })
```
MelissaAutumn, line 160: Seems like json_encoders will be deprecated in the near future
https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.json_schema_extra

but we can punt that to a later issue.

MelissaAutumn (approved):
Thank you!

backend/src/appointment/database/schemas.py
devmount: Good catch. I saw that but didn't find a quick way to migrate it, so let's do that later 😇

PR #808, opened 2024-12-20
Fix incorrect endtime in schedule validation message (Fixes #778)

### LINKED ISSUE #778: Wrong end time suggested in availability error
**Describe the bug**
When you set the availability end time so that the available time is shorter than the appointment duration settings, the resulting end time error suggests an incorrect end time.

**To Reproduce**
Dashboard => Scheduling Details
Verify booking duration is set to the default of `30 minutes`.
Dashboard => Set General Availability
Set start time 9:00 AM, and end time 9:05AM
Click `Save` button

**Expected behavior**
Error to appear indicating invalid end time, with correct end time suggestion (i.e. at least 9:30 AM).

**Actual behavior**
Error appears stating end time should be `at least 04:30 AM`.

**Screenshots**
Image

- OS: Mac OS 15.1.1
- Firefox 133.0 (aarch64)

**Additional context**

MelissaAutumn:
#778 

This should fix the incorrect endtime displayed in the schedule time validation error message. The backend was incorrectly returning it as a non-UTC time which then the frontend used to convert it from UTC to the schedule's timezone. (So double conversion.)

This should also fix an issue with daylight savings and this error message by always saving the current date as the start_date, as we don't actually expose that value to the end-user. (And it's used for daylight savings conversion.)

And finally I've adjusted the schedule creation test to use a timezone, and to use pacific time-converted values for the time so it's actually accurate. 

## Screenshots
image
image

cc @rwood-moz

### DIFF
--- a/backend/src/appointment/database/models.py
+++ b/backend/src/appointment/database/models.py
@@ -282,8 +282,8 @@ class Schedule(Base):
     details: str = Column(encrypted_type(String))
     start_date: datetime.date = Column(encrypted_type(Date), index=True)
     end_date: datetime.date = Column(encrypted_type(Date), index=True)
-    start_time: datetime.time = Column(encrypted_type(Time), index=True)
-    end_time: datetime.time = Column(encrypted_type(Time), index=True)
+    start_time: datetime.time = Column(encrypted_type(Time), index=True)  # in utc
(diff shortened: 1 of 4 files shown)


devmount (approved):
Awesome, thank you!

PR #809, opened 2024-12-20
Add a process to generate html from local md files and create a docs route for zoom

MelissaAutumn:
Added a new command based off the download-legal command that generates html pages from local docs. Currently we only need this for zoom because of their verification requirements but we can use it for other pages in the future.

The markdown pages are separated by language/locale right now, I don't think we need a German version of this page just yet though. We should wait until we confirm copy is okay, and Zoom accepts it.

This has to be on appointment.day as per the [guidelines](https://developers.zoom.us/docs/build-flow/prep-app-for-prod/#link--support) of the zoom verification process.

### Screenshots
Screen Shot 2024-12-20 at 12 42 20-fullpage

cc @malini

### DIFF
--- a/backend/src/appointment/commands/generate_documentation_pages.py
+++ b/backend/src/appointment/commands/generate_documentation_pages.py
@@ -1,0 +1,41 @@
+import os
+
+import markupsafe
+import requests
+import markdown
+
+
+def open_or_get(path: str):
+    if path.startswith('http'):
+        return requests.get(path).text
+
+    # Otherwise it's a path
+    with open(path, 'r') as fh:
+        return fh.read()
+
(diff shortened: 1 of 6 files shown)


malini (approved):
lgtm, we can discuss other Privacy Policy languages in the new year, gearing up for mvp

malini:
filed https://github.com/thunderbird/appointment/issues/810 for following up on privacy policy

MelissaAutumn:
> filed #810 for following up on privacy policy

Thanks! I think that comment was a copy-and-paste error from the download-legal but this is a good convo. This is just documentation for Zoom.

PR #812, opened 2025-01-02
Fix browserstack session name

rwood-moz:
Currently when the production sanity test is ran via github actions, the same browserstack build name is used each time (every test run is added to the same browserstack Automate build). This change updates the test configuration so that we will have a separate browserstack build for each time the test runs. See comment below for new screenshots from BrowserStack.


rwood-moz:
When run from a local machine the BrowserStack build name now includes the date/time:

Screenshot 2025-01-02 at 2 01 46 PM

jdbass (approved):
lgtm

PR #814, opened 2025-01-02
Expand the production sanity test

rwood-moz:
Add the following scenarios to the production sanity playwright E2E test:

- able to access the booking page via short link
- able to access the booking page via long link
- able to request a new booking

Here is a link to the new [updated test running via GitHub actions](https://github.com/thunderbird/appointment/actions/runs/12682709711) on this branch.

And here is the corresponding [recorded video of the test running in BrowserStack](https://automate.browserstack.com/dashboard/v2/builds/5a013832e5e5a6887bd38c4c95a293087035b9aa/sessions/a82f4bd488bd554f38515e4b949676952b98695c) (if you need BrowserStack credentials just let me know and I can set you up!).

### DIFF
--- a/test/e2e/pages/booking-page.ts
+++ b/test/e2e/pages/booking-page.ts
@@ -1,0 +1,166 @@
+import { expect } from '@playwright/test';
+import { type Page, type Locator } from '@playwright/test';
+import { APPT_PROD_MY_SHARE_LINK, APPT_PROD_SHORT_SHARE_LINK_PREFIX, APPT_PROD_LONG_SHARE_LINK_PREFIX } from '../const/constants';
+import { DashboardPage } from './dashboard-page';
+
+export class BookingPage {
+  readonly page: Page;
+  readonly titleText: Locator;
+  readonly invitingText: Locator;
+  readonly confirmBtn: Locator;
+  readonly bookingCalendar: Locator;
+  readonly calendarHeader: Locator;
+  readonly nextMonthArrow: Locator;
+  readonly availableBookingSlot: Locator;
+  readonly bookSelectionNameInput: Locator;
+  readonly bookSelectionEmailInput: Locator;
+  readonly bookSelectionBookBtn: Locator;
+  readonly requestSentTitleText: Locator;
+  readonly requestSentAvailabilityText: Locator;
+  readonly requestSentBookingSlot: Locator;
+  readonly requestSentCloseBtn: Locator;
+
+  constructor(page: Page) {
+    this.page = page;
+    this.titleText = this.page.getByTestId('booking-view-title-text');
+    this.invitingText = this.page.getByTestId('booking-view-inviting-you-text');
+    this.bookingCalendar = this.page.getByTestId('booking-view-calendar-div');
+    this.confirmBtn = this.page.getByTestId('booking-view-confirm-selection-button');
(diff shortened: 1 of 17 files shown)


test/e2e/pages/booking-page.ts
```diff
@@ -0,0 +1,166 @@
...
+  /**
+   * With the booking page week view already displayed, select the first available booking slot.
+   * If there is no slot available on the current week, this methond will skip to the next week
+   * and look for slots there. If no slots are avaible on the next week either, then an error
+   * will be raised.
+   * @param userDisplayName String containing the display name of the Appointment user
+   * @returns String containing the reference text for the time slot that was requested
+   * as retrieved from the DOM ie. 'event-2025-01-08 09:30'.
+   */
+  async selectAvailableBookingSlot(userDisplayName: string): Promise<string> {
```
devmount, line 89: Thank you for those very descriptive method comments 👍🏻

devmount (commented):
Just some feedback on the way 😇 Great work, thank you so much!

rwood-moz:
> Just some feedback on the way 😇 Great work, thank you so much!

Thanks for the feedback! Updated accordingly.

test/e2e/tests/book-appointment.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+    // this text also appears twice, once in the pop-up and once in underlying page
+    await expect(bookingPage.requestSentAvailabilityText.first()).toBeVisible();
+    await expect(bookingPage.requestSentAvailabilityText.nth(1)).toBeVisible();
+    const expectedText: string = `${PROD_DISPLAY_NAME}'s Availability`;
+    expect(bookingPage.requestSentAvailabilityText.first()).toContainText(expectedText);
+    expect(bookingPage.requestSentAvailabilityText.nth(1)).toContainText(expectedText);
+
+    // booking request sent dialog should display the correct time slot that was requested
+    // our requested time slot is stored in this format, as example: 'event-2025-01-14 14:30'
+    // the dialog reports the slot in this format, as example: 'Tuesday, January 14, 2025 02:30 PM'
```
malini, line 74: do you know what code generates the time displayed in the dialog? can that logic be turned into a util and used instead of writing the getTimeFromSlotString custom function?

rwood-moz: Hmmm interesting idea, thanks! Another thing that might be easier or more consistent with test automation would be to add a `data-testid` attribute to the available time slots and it could be formatted like

`data-testid='Tuesday, January 14, 2025 02:30 PM'` then when the test selects an available time slot it could grab the data-testid attribute value for the selected time slot; and it would already be formatted for what we need.

rwood-moz: Looks like the code that generates the event time in the request sent dialog is actually generated right in the frontend itself:

https://github.com/thunderbird/appointment/blob/main/frontend/src/components/BookingModal.vue#L54

`const time = computed(() => dj(props.event.start).format(`dddd, MMMM D, YYYY ${timeFormat()}`));`

So I don't think we can go that route.

rwood-moz: @malini Would you mind approving this as it is? I don't see any easy way to get rid of the custom `getTimeFromSLotString` function unfortunately.

malini (approved):

PR #817, opened 2025-01-07
Don't name inline attachments. (Fixes #816)

### LINKED ISSUE #816: Fix inline images showing up as attachments in emails
Currently we attach inline images like any other image. This work well as long as the cid is referenced in the email, most clients will hide the image from the attachment list. (https://github.com/thunderbird/appointment/blob/main/backend/src/appointment/controller/mailer.py#L116)

However it seems k9mail/tfa looks for `Content-Disposition: inline` which seems good and correct. So we should probably mark inline images as such.

MelissaAutumn:
Fixes #816 

Previously Condition-Disposition was being set to attachment because we named the file. But we want it as an inline file, so don't name the file. 

image

### DIFF
--- a/backend/src/appointment/controller/mailer.py
+++ b/backend/src/appointment/controller/mailer.py
@@ -118,7 +118,6 @@ class Mailer:
                     a.mime_main,
                     a.mime_sub,
                     cid=f'<{a.filename}>',
-                    filename=a.filename,
                 )
 
         return message


devmount (approved):
Such an easy fix, thank you!! Tested in Mailpit, TfA and Thunderbird ✅

PR #818, opened 2025-01-07
Add a localstorage option to change the start of the week (Fixes #589)

### LINKED ISSUE #589: Make the start of the week configurable
Following https://github.com/thunderbird/appointment/pull/587, the first day of the week is now set based on locale. Ideally , it should be configurable by the user, but we should explore options here on how to do that.

* For now, it can be another setting in our Settings page, where we allow the user to set the start of their week on their calendar using a dropdown.
* initiate a SoW doc for Design to get involved so we can improve this experience

MelissaAutumn:
Fixes #589 

I was originally going to show every day of the week, but Qalendar only supports sunday and monday 😢

Quick and dirty setting, as I think @devmount is handling pinia store for settings soon. 

### Screenshot
image

### DIFF
--- a/frontend/src/components/SettingsGeneral.vue
+++ b/frontend/src/components/SettingsGeneral.vue
@@ -1,11 +1,11 @@
 <script setup lang="ts">
 import { ColorSchemes } from '@/definitions';
 import {
-  ref, reactive, inject, watch,
+  ref, reactive, inject, watch, computed,
 } from 'vue';
 import { useI18n } from 'vue-i18n';
 import { useUserStore } from '@/stores/user-store';
-import { dayjsKey, callKey } from '@/keys';
+import { dayjsKey, callKey, isoWeekdaysKey } from '@/keys';
 import { SubscriberResponse } from '@/models';
 
 // component constants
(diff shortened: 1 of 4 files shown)


frontend/src/components/SettingsGeneral.vue
```diff
@@ -1,18 +1,22 @@
...
 
 // component constants
 const user = useUserStore();
 const { t, locale, availableLocales } = useI18n({ useScope: 'global' });
 const call = inject(callKey);
 const dj = inject(dayjsKey);
+const isoWeekdays = inject(isoWeekdaysKey);
+
+const startOfTheWeek = ref(localStorage?.getItem('startOfTheWeek') ?? isoWeekdays[0].long.toLowerCase());
+const availableStartOfTheWeekOptions = computed(() => isoWeekdays.filter((day) => day.long === 'Monday' || day.long === 'Sunday'));
```
devmount, line 19: This won't work for other languages than English. E.g. when viewing the settings page in German, this array would be empty.

devmount (changes requested):
Thank you for this! Works fine for English, but unfortunately doesn't work for other languages... 😅 

I added the German translation.

PR #824, opened 2025-01-15
Fix RemoteCalendarConnectionError init

rwood-moz:
When a `RemoteCalendarConnectionError()` is raised and no reason is provided the result is a TypeErrror:

```
>               raise RemoteCalendarConnectionError()
E               TypeError: RemoteCalendarConnectionError.__init__() missing 1 required positional argument: 'reason'
```

The fix is to have the RemoteCalendarConnectionError() reason parameter to have a default value of None.


devmount (approved):
Good catch, thank you!

PR #828, opened 2025-01-20
User store extended settings store

### LINKED ISSUE #804: Save user language setting to db
As soon as https://github.com/thunderbird/appointment/issues/766 is fixed, we should connect the Frontend setting for the UI language with the new subscriber language field.
This way, a user can change their language settings and it is persisted between different devices.

**Additional context**
We should implement a Pinia store for the (general) settings.

devmount:
## Description of the Change
This change

- Extends the subscriber model by all currently existing general settings (language, timezone, color scheme and time mode)
- Extends the user store to handle those settings and the UI to save the settings on change to the db
- Cleans up a lot of todos and deprecated code
- Extends user store tests for all general settings

## Benefits

Settings are now saved and restored consistently over multiple devices.

## Applicable Issues

Closes #804

### DIFF
--- a/frontend/src/models.ts
+++ b/frontend/src/models.ts
@@ -220,7 +220,7 @@ export type User = {
   preferredEmail: string;
   level: number;
   name: string;
-  timezone: string;
+  settings: UserConfig;
   username: string;
   signedUrl: string;
   avatarUrl: string;
@@ -228,6 +228,17 @@ export type User = {
   scheduleLinks: string[];
   isSetup: boolean,
   uniqueHash: string;
+};
+
+/**
+ * User settings to customize the application
+ * Used to store language, theme, time format and such.
+ */
+export type UserConfig = {
+  language: string,
+  colorScheme: string,
+  timeFormat: number,
+  timezone: string;
 };
 
 /**
@@ -241,15 +252,18 @@ export type UserActivity = {
 export type Subscriber = {
   id?: number;
   username: string;
-  name: string;
-  email: string;
-  preferred_email: string;
-  level: number;
-  timezone: string;
-  avatar_url: string;
-  is_setup: boolean;
-  unique_hash: string;
-  schedule_links: string[];
+  name?: string;
+  email?: string;
+  preferred_email?: string;
+  level?: number;
+  language?: string;
+  timezone?: string;
+  color_scheme?: string;
+  time_mode?: number;
+  avatar_url?: string;
+  is_setup?: boolean;
+  unique_hash?: string;
+  schedule_links?: string[];
   secondary_email?: string;
   invite?: Invite;
   time_created?: string;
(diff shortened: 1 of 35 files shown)


frontend/src/models.ts
```diff
@@ -220,7 +220,7 @@ export type User = {
   preferredEmail: string;
   level: number;
   name: string;
-  timezone: string;
+  settings: UserConfig;
```
devmount, line 223: I decided for a separate type here to keep things organized.

frontend/src/stores/user-store.ts
```diff
@@ -25,6 +35,50 @@ const initialUserObject = {
...
+    const dj = inject(dayjsKey);
+    const detectedTimeFormat = Number(dj('2022-05-24 20:00:00').format('LT').split(':')[0]) > 12 ? 24 : 12;    
+
+    data.value.settings = {
+      language: i18n.locale.value,
+      colorScheme: ColorSchemes.System,
+      timeFormat: detectedTimeFormat,
+      timezone: dj.tz.guess(),
+    };
+  }
```
devmount, line 60: This handles the new settings location for existing users already having a local copy of the user store

devmount (commented):
Some notes on the changes.

frontend/src/stores/user-store.ts
```diff
@@ -25,6 +35,50 @@ const initialUserObject = {
...
+   * @param fetch preconfigured function to perform API calls
+   */
+  const init = (fetch: Fetch) => {
+    call.value = fetch;
+  }
+
+  // Init user config if not already available
+  if (!data.value?.settings) {
+    const dj = inject(dayjsKey);
+    const detectedTimeFormat = Number(dj('2022-05-24 20:00:00').format('LT').split(':')[0]) > 12 ? 24 : 12;
```
MelissaAutumn, line 52: I think using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/hourCycle will work better?

MelissaAutumn (approved):
Looks good, just some small nits about colour which I realize is a bit silly. So no worries on changing that.

MelissaAutumn:
Looks like backend tests are broken right now

devmount:
Yes, I'm on it

frontend/src/stores/user-store.ts
devmount: The hourCycle property turned out to not being meant to work for detection. It is just a property already coded in the locale, e.g. the "h23" in "fr-FR-u-hc-h23". The function actually providing time format information for locales is https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getHourCycles but thats not yet working in Firefox, unfortunately.

PR #830, opened 2025-01-21
Enable code coverage analysis locally and in CI

rwood-moz:
Enable code coverage analysis (for the backend tests) both when running the tests locally as well as in CI. See the updated README file for more information.

Thanks @MelissaAutumn as you already had the groundwork for this in the repo! I am just updating the README and enabling the code coverage tool to run in CI.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -137,19 +137,30 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.12'
+          python-version: '3.13'
           cache: 'pip'
 
       - name: Install dependencies
         run: |
           cd ./backend
           python -m pip install --upgrade pip
-          python -m pip install .'[test]'
+          python -m pip install -e .'[test]'
 
       - name: Test with pytest
+        working-directory: ./backend
         run: |
-          cd ./backend && python -m pytest
-  
+          pwd
+          current_dir=$(pwd)
+          echo $current_dir
+          coverage run --data-file $current_dir/.coverage -m pytest --disable-warnings -s
+
+      - name: Generate code coverage report
+        working-directory: ./backend
+        continue-on-error: true
+        run: |
+          coverage report
+
+
   validate-frontend:
     needs: detect-changes
     runs-on: ubuntu-latest
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -149,9 +149,7 @@ jobs:
       - name: Test with pytest
         working-directory: ./backend
         run: |
-          pwd
           current_dir=$(pwd)
-          echo $current_dir
           coverage run --data-file $current_dir/.coverage -m pytest --disable-warnings -s
 
       - name: Generate code coverage report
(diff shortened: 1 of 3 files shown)


rwood-moz:
Got the code coverage report working both locally and in CI now \o/

Here's a link to a GHA validate-backend CI run and you can see the code coverage report:
https://github.com/thunderbird/appointment/actions/runs/12933592396/job/36072579769

Note: I am setting the `continue-on-error: true` flag for the generate code coverage report step because if the report fails or code coverage is under the threshold (70%) I don't want that to fail the entire workflow and block deployments.

Screenshot 2025-01-23 at 11 32 11 AM

.github/workflows/validate.yml
```diff
@@ -137,19 +137,30 @@ jobs:
...
+          python -m pip install -e .'[test]'
 
       - name: Test with pytest
+        working-directory: ./backend
         run: |
-          cd ./backend && python -m pytest
-  
+          pwd
+          current_dir=$(pwd)
+          echo $current_dir
```
MelissaAutumn, line 154: We can probably removd the debug code now right? 😄

backend/README.md
```diff
@@ -43,15 +43,65 @@ This project is deployed with Mozilla Accounts (known as fxa in the code.) Since
...
+cd backend && python -m pytest test/integration/test_schedule.py
+```
+
+To run the tests with the pytest warnings turned off:
+
+```bash
+cd backend && python -m pytest --disable-warnings
+```
+
+If you are debugging tests and have print staements inside them, run with this option so the output appears:
```
MelissaAutumn, line 73: typo!

backend/README.md
```diff
@@ -43,15 +43,65 @@ This project is deployed with Mozilla Accounts (known as fxa in the code.) Since
...
+
+```bash
+cd backend && python -m pytest -s
+```
+
```
MelissaAutumn, line 78: Thanks! I'm spoiled by pycharms test runner launcher

MelissaAutumn (approved):
Looks good! Some minor things but nothing thats not a 2 second fix.

Do we have a way to view the coveeage level from the pr?

.github/workflows/validate.yml
```diff
@@ -137,19 +137,30 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.12'
+          python-version: '3.13'
```
devmount, line 140: Nice! This should fix some issues we had due to different python versions 👍🏻

devmount (approved):
Awesome work, thank you Rob!

rwood-moz:
> Looks good! Some minor things but nothing thats not a 2 second fix.
> 
> Do we have a way to view the coveeage level from the pr?

There are some third-party github actions that allow just that - publish your code coverage results to a PR comment, i.e.:

https://github.com/marketplace/actions/python-coverage

However I'm not sure if we want to use third-party GHA that aren't certified by GitHub because of security; also it requires a separate github token also. Maybe something to consider in the future though!

PR #836, opened 2025-01-27
Fix backend lint errors

rwood-moz:
Backend linting (via `ruff check`) currently reports: `Found 100 errors`; update the code (and ruff config where appropriate) to address these lint errors.

### DIFF
--- a/backend/src/appointment/routes/waiting_list.py
+++ b/backend/src/appointment/routes/waiting_list.py
@@ -136,8 +136,9 @@ def invite_waiting_list_users(
 
     for id in data.id_list:
         # Look the user up!
-        waiting_list_user: models.WaitingList|None = \
+        waiting_list_user: models.WaitingList|None = (
             db.query(models.WaitingList).filter(models.WaitingList.id == id).first()
+        )
         # If the user doesn't exist, or if they're already invited ignore them
         if not waiting_list_user or waiting_list_user.invite:
             continue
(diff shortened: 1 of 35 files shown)


rwood-moz:
With this PR:

```
(venvtest) rwood@Roberts-MacBook-Pro backend % pwd
/Users/rwood/dev/appointment/backend
(venvtest) rwood@Roberts-MacBook-Pro backend % ruff check
All checks passed!
```

backend/src/appointment/routes/waiting_list.py
```diff
@@ -136,7 +136,8 @@ def invite_waiting_list_users(
 
     for id in data.id_list:
         # Look the user up!
-        waiting_list_user: models.WaitingList|None = db.query(models.WaitingList).filter(models.WaitingList.id == id).first()
+        waiting_list_user: models.WaitingList|None = \
```
MelissaAutumn, line 139: I think we can just wrap this with parenthesis instead of using \.

MelissaAutumn (approved):
Looks good at a glance!

backend/src/appointment/routes/waiting_list.py
devmount: (sneaked in here and edited Mels comment accordingliy 😇 )

rwood-moz: Oh wait I think you meant instead of using `\` ok so maybe this:

```
        waiting_list_user: models.WaitingList|None = (
            db.query(models.WaitingList).filter(models.WaitingList.id == id).first()
        )
```

rwood-moz: haha thank you both, updated

PR #844, opened 2025-01-29
Enable linting in CI, and add frontend build step in validate

rwood-moz:
This PR does a few things:

- Adds a frontend ESLINT `npm run lint` linting step in the GHA frontend validation 
- Adds a backend `ruff check` linting step in the GHA backend validation
- Adds a frontend build step (`npm run build`) during the frontend GHA validation as previously the frontend tests were run but no actual build was done, so build errors were not caught

**Note: Once this lands, the GHA validate job will start failing if there are linting errors. Therefore my other PRs must land first (which fix existing lint errors): #836  and #843.**

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -158,6 +158,10 @@ jobs:
         run: |
           coverage report
 
+      - name: Backend lint check
+        run: |
+          cd ./backend && ruff check
+
 
   validate-frontend:
     needs: detect-changes
@@ -177,7 +181,15 @@ jobs:
         run: |
           cd ./frontend
           npm install
-          
+
+      - name: Build frontend
+        run: |
+          cd ./frontend && npm run build
+
       - name: Test with vitest
         run: |
           cd ./frontend && npm run test -- --run
+
+      - name: Frontend lint check
+        run: |
+          cd ./frontend && npm run lint
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -146,6 +146,10 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install -e .'[test]'
 
+      - name: Backend lint check
+        run: |
+          cd ./backend && ruff check
+
       - name: Test with pytest
         working-directory: ./backend
         run: |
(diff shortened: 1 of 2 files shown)


rwood-moz:
I tested this out via GHA associated with this PR and I verified that:
- Backend and frontend linting checks run as part of GHA validate workflow
- If linting errors are found the GHA step/workflow will be marked as failed
- A frontend build is now done as part of the GHA validate workflow
- If the frontend build fails the GHA step/workflow will be marked as failed

If you want to have a closer look at these changes running in CI:
- Here's a validate-frontend GHA workflow where you can click into and see the build frontend and frontend linting steps running successfully
- Here's a [validate-backend GHA workflow](https://github.com/thunderbird/appointment/actions/runs/13036518701/job/36368248312)  where you can click into and see the backend linting step running successfully
- Here's a [validate-frontend workflow](https://github.com/thunderbird/appointment/actions/runs/13036518701/job/36368247488) where I intentionally made a temporary change in the source to break the frontend build, and you can see the frontent build step failed
- Here's a [validate-frontend workflow where linting errors](https://github.com/thunderbird/appointment/actions/runs/13036617661/job/36368592996) show up in the frontend linting, causing the step to fail

devmount (approved):
Lgtm! Just need to keep in mind to run the linter on my PRs when this landed 😂 
Thanks!

.github/workflows/validate.yml
```diff
@@ -177,7 +181,15 @@ jobs:
         run: |
           cd ./frontend
           npm install
-          
+
```
aaspinwall, line 184: I think linting should run before building. I think semantically the code should be formatted correctly in order for it to be built correctly. Let me know what you think

rwood-moz: Yes good point, I agree, will update, thanks!

devmount: Oh good point! Agreed! Also if I provide a PR and forgot to fix linter issues, it will be faster to report about that.

rwood-moz: Also did the same for `validate-backend`, now doing the linting first. [Latest is here](https://github.com/thunderbird/appointment/actions/runs/13077069446/job/36491757811?pr=844).

Issue #730, opened 2024-10-30
On first load of the General Availability page, the prev/next month buttons do not work properly


radishmouse:
**General Availability prev/next month arrow buttons do not work on first load**

Clicking the arrow to go to the next month skips a month. Clicking the arrow to to go the previous month does nothing.

This only happens on first load of the General Availability page. 

Note: Clicking the arrows to go from month to month act normally after the initial click.

**To Reproduce**

1. Visit a user's general availability calendar
2. Click the arrow to go to the next month or the previous month

**Expected behavior**
If it's currently October:
- clicking the arrow to go the next month should go to November.
- clicking the arrow to go the previous month should go to September.

**Actual behavior**
If it's currently October:
- clicking the arrow to go the next month goes to December.
- clicking the arrow to go the previous month does nothing

**Screenshots**

Reloading the page, then clicking the arrow to go to next month skips a month:
[Screencast_20241030_170002.webm](https://github.com/user-attachments/assets/cc631f96-7fe0-41ae-8cf8-2af466541276)

Reloading the page, then clicking the arrow to go to the previous month does nothing (until it is clicked a second time):
[Screencast_20241030_171220.webm](https://github.com/user-attachments/assets/b2e2384b-329e-4f49-90eb-b0dad8f0967f)

**System (please complete the following information):**

- OS: Linux
- Browser Version: 131.0.3 (64-bit)

devmount:
I just saw that [Qalendar is no longer maintained](https://github.com/tomosterlund/qalendar?tab=readme-ov-file#docs), so we'd have to find a workaround.

MelissaAutumn:
That's fun. I think there's some plans to redo the booking page. I've forked it a while ago with intention that we'd maintain / theme it as needed / fix any bugs we find though: https://github.com/thunderbird/qalendar

So if there's a low-effort fix we can apply that would solve this that would work, but if it's more than an hour, we should verify if we're planning on removing qalendar soon.

devmount:
Alright, I'll see if I can fix it. We should still verify the Qalendar removal imho.

devmount:
Closing this as there was no low-effort fix and it's no longer relevant as soon as we ditch the month view (or the calendar at all).

PR #846, opened 2025-01-31
Fix production sanity test end-of-month issues

rwood-moz:
The production sanity E2E test failed because of a couple of issues related to end-of-month:

- Since there were no slots left at the end of January the request booking page automatically displays February, even though the current month is still January (makes sense); however the test was expecting the request booking page header to be the current month of January. Update the test to accept either the current month or next month in the request booking page header.

- When the test signs into the host account to verify a hold appointment was created at the correct time slot, the current month (January) was displayed however the appointment was actually created for February since no slots were left in January. The test should have automatically skipped forward to February (since the hold appointment wasn't found in the current month) but it didn't (a bug); fix this.

### DIFF
--- a/test/e2e/tests/book-appointment.spec.ts
+++ b/test/e2e/tests/book-appointment.spec.ts
@@ -14,12 +14,35 @@ const verifyBookingPageLoaded = async () => {
   await expect(bookingPage.invitingText).toBeVisible();
   await expect(bookingPage.invitingText).toContainText(PROD_DISPLAY_NAME);
   await expect(bookingPage.bookingCalendar).toBeVisible();
-  // calendar header should contain current MMM YYYY
+
+  // verify calendar header
   const today: Date = new Date();
   const curMonth: string = today.toLocaleString('default', { month: 'short' });
   const curYear: string = String(today.getFullYear());
-  await expect(bookingPage.calendarHeader).toHaveText(`${curMonth} ${curYear}`);
-  // confirm button is disabled by default until a slot is selected
+
+  // by default you can only book slots 1-14 days in the future; if it's near the end of the
+  // month then there's a chance there are no slots availble to be booked; the booking request
+  // page always shows the month that has the first available time slot; so the month displayed
+  // may be the current month or one month in the future (and perhaps next year if January)
+
+  // the header may contain either the current month or the next month
+  today.setMonth(today.getMonth() + 1, 1);
+  const nextMonth = today.toLocaleString('default', { month: 'short' });
+  const monthRegex = new RegExp(String.raw`${curMonth}|${nextMonth}`);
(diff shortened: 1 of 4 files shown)


test/e2e/tests/book-appointment.spec.ts
```diff
@@ -14,11 +14,24 @@ const verifyBookingPageLoaded = async () => {
...
-  // calendar header should contain current MMM YYYY
+
+  // verify calendar header
   const today: Date = new Date();
   const curMonth: string = today.toLocaleString('default', { month: 'short' });
   const curYear: string = String(today.getFullYear());
-  await expect(bookingPage.calendarHeader).toHaveText(`${curMonth} ${curYear}`);
+
+  // expect correct year
+  await expect(bookingPage.calendarHeader).toContainText(curYear);
```
devmount, line 24: This wouldn't work at the end of the year, when the switch is from December to January, or would it?

devmount (commented):
I know this is still draft, just one question 😇

test/e2e/tests/book-appointment.spec.ts
devmount: Great, thanks!

rwood-moz:
In this PR I fixed the first issue (month value in the header could be this month or next month). Made the same for the year since the year might be the current year + 1 if current year is Dec and the displayed month is Jan.

About the second issue - scrolling to the next month on the host dashboard to verify the hold appt created:

Due to another issue in Appointment (issue #730) today (Jan 31) clicking the `>` forward arrow actually causes the calendar to skip ahead from Jan to Mar; and the test fails as it is looking for a hold appointment created in Feb. So I'll need to find a new solution for going ahead one month instead of clicking the next `>` month arrow. Perhaps click on the month button, or maybe browse to the `Bookings` tab and view a list of `all pending bookings` and ensure our requested booking shows up there.

rwood-moz:
This PR now also changes the way a requested appointment is verified. With this change instead of searching for the new HOLD appointment on the dashboard calendar, the test navigates to the bookings tab, views all future pending bookings, filters by the users, and then verifies a HOLD appointment is listed for the expected time slot.

rwood-moz:
Here's [a link to BrowserStack](https://automate.browserstack.com/dashboard/v2/builds/7f8c02497a0bfa69f5d465c77943f0d49ac93349/sessions/26592258e8f31b0c880283b4f378d5e1f130e08e) (including video playback) of running the updated test. If you need BrowserStack credentials just let me know and I can set you up. 👍

devmount (approved):
I don't see any issues here. Thanks for your work!

rwood-moz:
> I don't see any issues here. Thanks for your work!

Thanks for the review, appreciated!

Issue #850, opened 2025-02-04
Time setting seems to default to 24h


MelissaAutumn:
I'm guessing this is related to the store changes, but it seems like it at least breaks sanity tests. 

cc @rwood-moz @devmount

rwood-moz:
Example [test failure here](https://github.com/thunderbird/appointment/actions/runs/13141217335/job/36668907362).

```
Expected string: "Wednesday, February 5, 2025 11:30 AM"
Received string: "Wednesday, February 5, 2025 11:30"
```

devmount:
I guess I'm to blame here 😅  I implemented the general settings store in #828 which set 24h as default iirc

devmount:
@rwood-moz Yes, in my understanding it was like this. If you rather want h12 as default, we can implement that too.

rwood-moz:
@devmount No blame, just trying to figure out what happened :) For the default time format (12 or 24hr) what you and @MelissaAutumn prefer works for m! If it's 24 hr I can update the tests easily. :)

devmount:
@MelissaAutumn I think you approved the h24 default? Any opinion on this?

MelissaAutumn:
We should probably be consistent with default settings between services. Most of the team is 12 hour time based, but dunno about user stats. 

cc @malini @ryanjjung @radishmouse @aaspinwall for any input / suggestions.

malini:
I don't know either, honestly.

I'd default to 12hr for users and they can swap to 24hrs. My guess is that people used to 12hrs would be immediately confused by seeing 18:00 and might think it an error, but someone used to 24 hours would see 6PM and think "ugh, gotta switch to 24 hour mode".

devmount:
Alright, I'll switch this to 12 hr default.

rwood-moz:
I tried changing the Appointment test production account (that the test uses) back to 12hr format so that the test will start passing again - but in production now, for me it won't save settings changes. I changed the settings back to 12hr, but it never updates the dashboard, and if I sign out and back in the settings are back to 24 hour. Same if I change to dark mode and sign out, when I sign back in it's back to light mode.

@devmount Would you mind giving that a try, are you able to change and save settings in production?

devmount:
@rwood-moz Thanks, I'll take a look at that. Can you check your browser local storage if changing that setting changes the corresponding local storage entry?

rwood-moz:
After changing to `12hr` the local storage does show the change:

[...]

PR #852, opened 2025-02-05
Remove references to prod in env var names

rwood-moz:
In order to allow the E2E tests to run on environments other than production, the first step is to rename the environment variables to be generic (remove `prod` references) and also make a new `.env.prod.example` env vars file. This will clear the way to add more supported environments (i.e. stage).


rwood-moz:
Wait for issue #850 to be resolved before landing this (as the existing prod sanity test is currently failing, and I want to test this PR out with that resolved first).

ryanjjung (approved):
~prod~

PR #854, opened 2025-02-05
Calculate the cipher length and base64 length ahead of time. (Fixes #853)

### LINKED ISSUE #853: Encrypted fields have an issue with truncating.
Encrypted fields are encrypted and then converted to base64 as noted here: https://github.com/kvesteri/sqlalchemy-utils/blob/master/sqlalchemy_utils/types/encrypted/encrypted_type.py#L114

We don't take into account the length of the additions from the cipher or base64 encoding. We should do that, and then update the strings and do a data migration to nuke a known case with schedule's detail field: https://thunderbird.sentry.io/issues/6269890787/?project=4505428124827648&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&statsPeriod=14d&stream_index=13

MelissaAutumn:
Fixes #853

I also added a clean-up migration right before the field length change to fix an existing user's schedules.details field which is currently erroring out. I'm not sure how that happened because I got a mysql truncate error on submitting 250 characters, but maybe prod is setup in non-strict mode. (spooky if true)

I've tested lengths manually via the application, and added a test to automate some existing cases we have. 

The type length migration was generated via alembic's autogenerate command (with a small edit to env.py that I didn't commit.) and tweaked to fix the autogen errors. It's a scary migration, but the function is tested, and we have stage to potentially nuke if needed.

### DIFF
--- a/backend/src/appointment/database/models.py
+++ b/backend/src/appointment/database/models.py
@@ -96,9 +96,17 @@ class TimeMode(enum.Enum):
     h24 = 24
 
 
+def calculate_encrypted_length(length: int):
+    """Calculate the length of the string after it's been encrypted and encoded."""
+    cipher_length = length + 16 - (length % 16)  # Fixed block size of 16 for Aes
+    return int((cipher_length + 2) / 3) << 2  # Base64 with padding
+
+
 def encrypted_type(column_type, length: int = 255, **kwargs) -> StringEncryptedType:
     """Helper to reduce visual noise when creating model columns"""
-    return StringEncryptedType(column_type, secret, AesEngine, 'pkcs5', length=length, **kwargs)
+    return StringEncryptedType(
+        column_type, secret, AesEngine, 'pkcs5', length=calculate_encrypted_length(length), **kwargs
+    )
 
 
 @as_declarative()
@@ -441,6 +449,7 @@ class Invite(Base):
 
 class WaitingList(Base):
     """Holds a list of hopefully future-Appointment users"""
+
     __tablename__ = 'waiting_list'
 
     id = Column(Integer, primary_key=True, index=True)
--- a/backend/src/appointment/database/schemas.py
+++ b/backend/src/appointment/database/schemas.py
@@ -166,8 +166,8 @@ class ScheduleBase(BaseModel):
     slug: Optional[str] = None
     calendar_id: int
     location_type: LocationType | None = LocationType.inperson
-    location_url: str | None = None
-    details: str | None = None
+    location_url: str | None = Field(max_length=2048)
+    details: str | None = Field(max_length=250)
     start_date: date | None = None
     end_date: date | None = None
     start_time: time | None = None
--- a/backend/src/appointment/migrations/versions/2025_02_05_1859-16c0299eff23_update_enc_field_lengths.py
+++ b/backend/src/appointment/migrations/versions/2025_02_05_1859-16c0299eff23_update_enc_field_lengths.py
@@ -1,0 +1,411 @@
+"""update enc field lengths
+
+Revision ID: 16c0299eff23
+Revises: 645fd31f827d
+Create Date: 2025-02-05 18:59:31.214328
+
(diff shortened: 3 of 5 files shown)


backend/src/appointment/database/schemas.py
```diff
@@ -166,8 +166,8 @@ class ScheduleBase(BaseModel):
     slug: Optional[str] = None
     calendar_id: int
     location_type: LocationType | None = LocationType.inperson
-    location_url: str | None = None
-    details: str | None = None
+    location_url: str | None = Field(max_length=2048)
+    details: str | None = Field(max_length=250)
```
MelissaAutumn, line 170: We should fill these out for the rest of the schemas/fields 😄

backend/test/unit/test_models.py
```diff
@@ -1,6 +1,32 @@
...
+
+
+class TestMisc:
+    def test_calculate_encrypted_length(self, faker):
+        """Ensures an encrypted string is equal to our length calc method.
+        We have a list of clear_strs which are random bits of text with variable length.
+        We encrypt it (same method used for db fields) and run it through our calculate method,
+        and assert they are the same length"""
+        clear_strs = [
+            'a',
```
MelissaAutumn, line 15: text has a min length of 5, so here's 1 character instead.

backend/src/appointment/database/models.py
```diff
@@ -96,9 +96,17 @@ class TimeMode(enum.Enum):
...
+    cipher_length = length + 16 - (length % 16)  # Fixed block size of 16 for Aes
+    return int((cipher_length + 2) / 3) << 2  # Base64 with padding
+
+
 def encrypted_type(column_type, length: int = 255, **kwargs) -> StringEncryptedType:
     """Helper to reduce visual noise when creating model columns"""
-    return StringEncryptedType(column_type, secret, AesEngine, 'pkcs5', length=length, **kwargs)
+    return StringEncryptedType(
+        column_type, secret, AesEngine, 'pkcs5', length=calculate_encrypted_length(length), **kwargs
+    )
```
devmount, line 109: Somehow I always thought, that the StringEncryptedType would take care of the appropriate length automatically, but well, that was a little too naive it seems 😅

backend/src/appointment/migrations/versions/2025_02_05_1859-16c0299eff23_update_enc_field_lengths.py
devmount, line 1: That's what I call a migration 😂  Thanks for providing the downgrade for the fields too!

MelissaAutumn:
minor regression related to the field max length change, feel free to ignore that as i fix it.

devmount (approved):
Awesome work, Mel! Thanks for taking care of that! I wasn't aware that this was an issue.
Code looks fine, I can only test this tomorrow. But feel free to merge in if you want.

backend/src/appointment/database/models.py
MelissaAutumn: same 😅

backend/src/appointment/migrations/versions/2025_02_05_1859-16c0299eff23_update_enc_field_lengths.py
MelissaAutumn: That was all alembic thankfully!

MelissaAutumn:
I'll hold off until you test, more testing the better!

devmount:
Alright, I'll report back tomorrow!

devmount (approved):
Ok I ran these migrations and tested through the application. Didn't notice anything suspicious. Booking works. Confirmation works. I guess my local test data didn't run into any issues because no column limits were reached.

I only found a little regression caused by my previous PR about the email templates. will provide a fix asap.

I say: let's test this on stage!

MelissaAutumn:
I will merge tomorrow so I'm around when it deploys.

PR #857, opened 2025-02-06
Fix inline icons in invitation email

### LINKED ISSUE #856: Broken image in invitation email
**Describe the bug**
After #847 one of the icons isn't shown anymore.

**Expected behavior**
All inline images load properly.

**Screenshots**

Image

devmount:
## Description of the Change

This PR fixes indices of inline images in the invitation email. Maybe for the future it's better to work with dictionary keys instead of positional list keys for attachements.

## Benefits

Fixes a broken icon image.

## Applicable Issues

Closes #856

### DIFF
--- a/backend/src/appointment/controller/mailer.py
+++ b/backend/src/appointment/controller/mailer.py
@@ -251,8 +251,8 @@ class InvitationMail(BaseBookingMail):
             duration=self.duration,
             # Image cids
             tbpro_logo_cid=self._attachments()[0].filename,
-            calendar_icon_cid=self._attachments()[1].filename,
-            clock_icon_cid=self._attachments()[2].filename,
+            calendar_icon_cid=self._attachments()[2].filename,
+            clock_icon_cid=self._attachments()[3].filename,
         )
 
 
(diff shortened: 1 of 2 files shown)


backend/src/appointment/controller/mailer.py
```diff
@@ -251,8 +251,8 @@ def html(self):
             duration=self.duration,
             # Image cids
             tbpro_logo_cid=self._attachments()[0].filename,
-            calendar_icon_cid=self._attachments()[1].filename,
-            clock_icon_cid=self._attachments()[2].filename,
+            calendar_icon_cid=self._attachments()[2].filename,
+            clock_icon_cid=self._attachments()[3].filename,
```
MelissaAutumn, line 255: We should maybe switch to a dict, or change the BaseBookingMail so that inline images are separate from other attachments.

MelissaAutumn (approved):
Good idea, worth doing it now or should we punt it to a ticket?

backend/src/appointment/controller/mailer.py
devmount: I'm fine with both! Can you put that into a follow-up ticket? Thanks!

MelissaAutumn: https://github.com/thunderbird/appointment/issues/860 🎉

devmount: Thank you!

PR #868, opened 2025-02-12
Fallback to `VITE_DEFAULT_HOUR_FORMAT` if the browser doesn't provide a time format. (Fixes #867)

### LINKED ISSUE #867: Fix default time format if value isn't defined by browser
`.hour12` is an optional value of `Intl.DateTimeFormat().resolvedOptions()`. D'oh.

MelissaAutumn:
Should fix this right up and allow folks to override this on a self-hosted install.


rwood-moz (approved):
LGTM! Tried it out locally and confirmed in a new / private browser the dashboard and request booking (share link) page defaults to `12 hr`. Also successfully able to change both to `24 hr` and saved locally. Awesome!

PR #873, opened 2025-02-18
Add Zendesk Link

### LINKED ISSUE #872: Add Zendesk support link
Our existing email-based solution isn't reliable as we don't have any visibility on who's replying to questions. We have a Zendesk solution coming up, but for now we should simply link it up.

MelissaAutumn:
Fixes #872 

This also removes the reply-to header for support emails, and replaces it with a link to our contact form. (Which now redirects to zendesk.)

### DIFF
--- a/frontend/src/views/ContactView.vue
+++ b/frontend/src/views/ContactView.vue
@@ -1,93 +1,48 @@
 <script setup lang="ts">
-import { inject, ref } from 'vue';
-import { useI18n } from 'vue-i18n';
-import { useUserStore } from '@/stores/user-store';
-import { AlertSchemes } from '@/definitions';
-import { callKey } from '@/keys';
-import { BooleanResponse } from '@/models';
-import PrimaryButton from '@/elements/PrimaryButton.vue';
-import AlertBox from '@/elements/AlertBox.vue';
-import TextInput from '@/elements/TextInput.vue';
+import { onMounted } from 'vue';
+import LoadingSpinner from '@/elements/LoadingSpinner.vue';
 
-// icons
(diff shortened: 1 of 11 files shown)


frontend/src/views/ReportBugView.vue
```diff
@@ -2,8 +2,10 @@
 import { onMounted } from 'vue';
 import LoadingSpinner from '@/elements/LoadingSpinner.vue';
 
+const redirectUrl = import.meta.env?.VITE_REPORT_BUG_URL ?? '/';
+
 const redirect = async () => {
-  window.location = import.meta.env?.VITE_REPORT_BUG_URL ?? '/';
+  window.location.replace(redirectUrl);
```
MelissaAutumn, line 8: Switched to .replace so you're less likely to fight with your back button.

frontend/src/views/ContactView.vue
devmount, line 1: Do we still need this file at all? We could just solve the redirect in the router, or am I missing something?

devmount (approved):
Looking good, thank you! I added the German translation and one question concerning routing below.

frontend/src/views/ContactView.vue
MelissaAutumn: We could, but I don't want to stuff props inside the router file. It does make for a very sad SFC though.

devmount: Ok, alright 👍🏻

frontend/src/views/ReportBugView.vue
devmount: Convenient, thanks!

PR #876, opened 2025-02-19
Pin backend scale to 2 containers

ryanjjung:
This removes the environment name condition surrounding container scale.


MelissaAutumn (approved):
Thanks!

PR #890, opened 2025-02-26
Update the download data readme and make filenames consistent

rwood-moz:
While trying out the `download your data` feature I noticed the downloaded data readme is out of date, and the files are not named consistently (some are plural and some aren't). Update the readme and change the name of some of the data csv files so they are more consistent.

Here are the resulting csv files downloaded:

Screenshot 2025-02-26 at 4 45 15 PM

### DIFF
--- a/backend/src/appointment/l10n/en/main.ftl
+++ b/backend/src/appointment/l10n/en/main.ftl
@@ -145,7 +145,12 @@ account-data-readme = ============================================
                       The following files are included:
                        - appointments.csv : A list of Appointments from our database
                        - attendees.csv : A list of Appointment Slot Attendees from our database
+                       - availability.csv : Schedule availability data from our database
                        - calendars.csv : A list of Calendars from our database
-                       - slots.csv : A list of Appointment Slots from our database.
-                       - subscriber.csv : The personal information we store about you from our database.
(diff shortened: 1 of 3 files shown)


MelissaAutumn (approved):
awesome ty!

backend/src/appointment/l10n/en/main.ftl
devmount, line 1: An (unqualified) question regarding language: We often use the term _"... from our database"_ - isn't that kind of implied anyway, if I download my existing data from a service? I was just thinking if this might be misunderstood into the other direction, that this is only the data from our database, and there might be more lying around at other databases 😅

devmount (approved):
Thank you Rob for taking care of this!! I improved the German lang slightly and had one minor question.

backend/src/appointment/l10n/en/main.ftl
devmount: You're welcome! One moment, I'll quickly check that

devmount: Alright, removed it there too ✅

rwood-moz: Awesome, thanks 😃

PR #900, opened 2025-03-04
Fix some settings headings/strings

rwood-moz:
While developing more E2E tests I noticed some small inconsistencies and typos in some settings headers/string values. Capitalizing the `Settings` in the settings headers; and fixing the language selection entry for english (but in the DE locale).


devmount (approved):
Good work! Thank you for catching those inconsistencies 🙌🏻

PR #910, opened 2025-03-06
Run E2E tests in CI when E2E test source changes

rwood-moz:
Update the GHA validate workflow so that when the E2E test source itself changes, the E2E tests should be run as part of the PR checks.

Also a small backend README correction.


ryanjjung (approved):
🍰

PR #917, opened 2025-03-10
Fix url parsing issues

### LINKED ISSUE #904: Fix codeql issues
**Describe the bug**
In https://github.com/thunderbird/appointment/security/code-scanning, there are 4 sanitization issues we should address. 

Ignore the "Full server-side request forgery" issue, as it's not something we have to use user-given urls to set up their CalDAV.

devmount:
## Description of the Change

This PR makes the way we check for hostnames more precise.

## Applicable Issues

Fixes #904

### DIFF
--- a/backend/src/appointment/controller/calendar.py
+++ b/backend/src/appointment/controller/calendar.py
@@ -1012,8 +1012,13 @@ class Tools:
         """Fix up some common url issues with some caldav providers"""
         parsed_url = urlparse(url)
 
+        # If for some reason we don't have a hostname, there's nothing to fix
+        # and we just throw back that weird url
+        if not parsed_url.hostname:
+            return url
+
         # Handle any fastmail issues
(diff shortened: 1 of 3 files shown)


devmount:
@MelissaAutumn I'm not sure I agree with the code suggestions above, since imo a check for `parsed_url.hostname.endswith(...)` should handle all cases here, but might be just me 😅

MelissaAutumn:
Gosh that's a spammy alert. I guess the detection here is picking up that you can technically register notactuallyfastmail.com, and that would pass. 

I'd maybe throw is_valid_hostname into utils, and write a few test cases to ensure that it's working correctly with a couple cases (like notactuallyfastmail.com for example 😄)

devmount:
I'll do that, thanks!

devmount:
@MelissaAutumn Feel free to take a look again. I added the utility function and tests now

MelissaAutumn (approved):
ty!

PR #920, opened 2025-03-10
Timezone offset issue

### LINKED ISSUE #919: Timezone offset issues
It's that time of the year again for timezone issues.

MelissaAutumn:
Fixes #919 

This took a bit of thinking. But we can't actually calculate the time in UTC as the stored schedule start/end time is in UTC from potentially a different offset than the incoming booking start/end time.

So lets transform these all to local time and do the comparison. We have the intended start/end time's offset from when the schedule was last saved, so we can use that to transform 17 back to 9, and today's offset can be used to transform 16 to 9. 

I've also adjusted the test subscriber's timezone to default to America/Vancouver, and updated some tests.

### DIFF
--- a/backend/src/appointment/database/models.py
+++ b/backend/src/appointment/database/models.py
@@ -229,7 +229,7 @@ class Calendar(Base):
     connected = Column(Boolean, index=True, default=False)
     connected_at = Column(DateTime)
 
-    owner: Mapped[Subscriber] = relationship('Subscriber', back_populates='calendars')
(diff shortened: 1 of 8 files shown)


MelissaAutumn:
I'm going to merge this so we can check if stage is fixed. And follow up with a second PR if things are still broken.

rwood-moz:
> I'm going to merge this so we can check if stage is fixed. And follow up with a second PR if things are still broken.

Thanks Mel, I created #922 to run E2E tests on stage only after main merges (and not on PRs).

Issue #922, opened 2025-03-10
Run E2E tests after stage deployments (main merges) not on PRS


rwood-moz:
For some reason I was thinking that when Appointment PRs are created they are automatically deployed to stage when the PR branch is created/updated, and then the tests run after that. However that is not the case, **PRs are only deployed to stage after they are merged into main**. Therefore the current E2E tests running on PRs/branch pushes are invalid as they aren't actually running against the PR branch itself; they're running on whatever is currently deployed to the `stage` environment.

Thanks to @MelissaAutumn for catching this.

So for now turn off the E2E tests step running as part of PR validations; and instead **have the E2E tests run on the `stage` environment after merges to main are done (and after they are deployed to stage)**. So this will mean adding a run-e2e-tests step as part of the [deploy-staging GHA workflow](https://github.com/thunderbird/appointment/blob/main/.github/workflows/deploy-staging.yml).

PR #923, opened 2025-03-11
Run E2E tests on stage only after stage deployments

rwood-moz:
Only run the E2E tests on stage after stage deployments; no point in running them after PR/branch updates because the branches themselves aren't being deployed to stage until after they're merged to main anyway. More info in #922.

### DIFF
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -364,3 +364,46 @@ jobs:
             ecr_tag.zip
             frontend.zip
             iac.zip
+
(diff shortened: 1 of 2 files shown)


rwood-moz:
The  [GHA checks on this PR ](https://github.com/thunderbird/appointment/actions/runs/13789901724/job/38566765320?pr=923)verified that the E2E tests are no longer run as part of PR validation.

MelissaAutumn (approved):
thanks! I know sometimed the backend takes a while to start up, so we may need to tweak timing. But we can give this a shot!

Issue #914, opened 2025-03-07
Emails will fail to login if the casing does not match


rtanglao:
**Describe the bug**
An appointment waitlist person received the email and tried to sign in but was asked for an invite code.
**To Reproduce**

1. Signed up for the wait list on January 2, 2025 (see image attachment (1))
2. Received the invite email, clicked on "Sign Up"
3. Clicked on "Login" (instead of "Sign up for the beta")
4. Appointment Asked for an invite code

**Expected behavior**
Appointment doesn't ask for an invite code. Instead, it prompts them to create/login to a Mozilla Account and then once logged into their Mozilla

**Actual behavior**
Prompted for an invite code. 
Could this be a user error i.e. user should have clicked on "Sign up for the beta" perhaps?

**Screenshots**

* (1): ![Image](https://github.com/user-attachments/assets/550fb624-b85d-4196-9608-a36abb07381c)

**System (please complete the following information):**

- OS: [e.g. Windows 10]
- Browser Version [e.g. Firefox 114.0b8 (64-Bit)]

**Additional context**
https://mzla.zendesk.com/agent/tickets/11 and perhaps the same problem in https://mzla.zendesk.com/agent/tickets/12

MelissaAutumn:
Most likely user error, but we should add the email they signed up with in the actual email invite with a "sign in with this email"

MelissaAutumn:
Oh I am wrong, and a fool. It's actually the case sensitivity issue. Will provide a fix in a bit.

PR #926, opened 2025-03-11
Page the admin panel routes, and add a hard-delete route in admin panel. (Fixes #925)

### LINKED ISSUE #925: Admin panel subscriber list errors out
We need to start paging it

MelissaAutumn:
Bit of a big one. My simple solution to listing all of the subscribers finally 500s. Yay, we can replace it with a paginated list. 

I believe we should be okay if we pre-fetch each page, it's most likely just the size, so thats what I've done here. On page-load it will prefetch the first page to get a page count, and grab the rest, offsetted by 250ms. 

Should work I think?

Additionally this adds a hard-delete button so we can remove a user from appointment. They're required to be disabled/soft-deleted first. And it's behind a confirmation modal. This is so we can fix up some existing stuck users thanks to the email casing issue in #914


frontend/src/views/admin/SubscriberPanelView.vue
```diff
@@ -67,6 +71,13 @@ const filteredSubscribers = computed(() => subscribers.value.map((subscriber) =>
     value: subscriber.time_deleted ? 'Enable' : 'Disable',
     disabled: !subscriber.time_deleted && subscriber.email === user.data.email,
   },
+  hardDelete: {
+    type: TableDataType.Button,
+    buttonType: TableDataButtonType.Caution,
+    value: 'Hard Delete',
+    disabled: !subscriber.time_deleted || subscriber.email === user.data.email,
+    tooltip: 'Completely remove the user from the database. This is the same as deleting someones account!! You must disable the account first.',
```
MelissaAutumn, line 79: Oh also added tooltips as a possibility for buttons

devmount: Nice. I suggest we localize those too.

frontend/src/views/admin/SubscriberPanelView.vue
```diff
@@ -273,4 +319,15 @@ onMounted(async () => {
   <div v-else class="flex size-full min-h-[75vh] items-center justify-center">
     <loading-spinner/>
   </div>
+  <!-- Refresh link confirmation modal -->
+  <confirmation-modal
+    :open="hardDeleteModalOpen"
+    title="WARNING"
+    message="This will completely remove a user from Thunderbird Appointment, including all of their user data. Don't do this unless you're told to."
+    confirm-label="Delete"
```
devmount, line 327: Localize those would be nice too!

devmount (approved):
Awesome work! Just two comments for (maybe) missing localization.

frontend/src/views/admin/SubscriberPanelView.vue
MelissaAutumn: Ah yeah, they're internal only so wasn't a priority. There's a few things in the admin panel aren't localized. So didn't want to bother you with that part 😅 

I'll create a follow up ticket.

devmount: Perfect, thanks!

PR #927, opened 2025-03-11
Update deps

MelissaAutumn:
Works on my machine!


rwood-moz (approved):
Works on my machine too 👍

PR #930, opened 2025-03-12
Don't include relationships for the Subscriber admin item

MelissaAutumn:
I didn't get this error until now...odd. But this prevents it from recursively calling the relationships...oops.

This change pulls in some of schema.Subscriber keys without pulling in the relationships to calendars and slots that we don't need in this context.


rwood-moz (approved):
lgtm

PR #933, opened 2025-03-12
Disable/Enable Routes Should Use ID (Fixes #932)

### LINKED ISSUE #932: Enable/Disable routes need to use id

MelissaAutumn:
get_by_email now does searches on lowercase emails only, so we can't delete a mixed-case user lol!

This should have been using ids from the start anyway. 

Additionally I've added in a quick search textbox to help find emails faster (or invite codes)


devmount (approved):
Good thing to switch to ids, thank you!

PR #937, opened 2025-03-12
Escape regex in admin panel search filter

MelissaAutumn:
Not critical because it's not something folks can access without permission. 

Note this function is not baseline, however it's in Firefox and Safari, and will be shipping soon in Chrome-based browsers
https://chromium-review.googlesource.com/c/chromium/src/+/6305455


devmount (approved):
TIL, thanks for this improvement!

PR #942, opened 2025-03-14
Init fetch for stores

### LINKED ISSUE #918: Remove fetch parameter in each API calling store function
In the user store, we now have an init function to get rid of pulling the fetch function through every store method that does API calls. We should refactor the other stores as well with such an init function for convenience.

We can extend that init function's functionality in the future if needed.

Image

devmount:
## Description of the Change

Implements an `init(...)` function for stores to provide data required at runtime. Also add basic tests for each store where we didn't have one yet.

## Applicable Issues

Closes #918


MelissaAutumn (approved):
Looks good. 

If we want to streamline the init a bit we could create some factory functions to help smash init and use together

something like:

```
createAppointmentStore = (call: fetch) => {
  const store = useAppointmentStore();
  store.init(call);
  return store;
};
```

But this looks good as-is!

devmount:
Good thinking, thank you! I'll add that and merge after 👍🏻

PR #944, opened 2025-03-17
Fix Admin/Invites not returning the proper total_pages count (Fixes #943)

### LINKED ISSUE #943: Pagination not retrieving every page

MelissaAutumn:
Ah the important of testing.


devmount (approved):
Good catch, thanks for this!
The importance of testing indeed. Should we expand pagination tests to other models too? Or is this currently the only occurence?

PR #945, opened 2025-03-17
Pass the actual error back to the end user and remove some redundant url cleanup functions.

MelissaAutumn:
Just cleaning up some issues I found during some testing. We implemented dns lookup a while ago, and with this minor fix we no longer need to "clean up" any urls. 

image

### DIFF
--- a/backend/src/appointment/exceptions/validation.py
+++ b/backend/src/appointment/exceptions/validation.py
@@ -214,9 +214,6 @@ class RemoteCalendarConnectionError(APIException):
     status_code = 400
 
     def get_msg(self):
-        if self.reason:
-            return self.reason
-
         return l10n('remote-calendar-connection-error')
 
 


backend/src/appointment/exceptions/validation.py
```diff
@@ -214,6 +214,9 @@ class RemoteCalendarConnectionError(APIException):
     status_code = 400
 
     def get_msg(self):
+        if self.reason:
+            return self.reason
```
devmount, line 218: This would return the same string as title message and as description for the frontend alert. Imo this could just be left as is, when we're using the corresponding alert component.

backend/src/appointment/controller/calendar.py
devmount, line 1: Nice! Feels good to throw away code 😇 🧹

devmount (commented):
Thank you for this. We might want to just use our `<alert-box>` component to show the error messages with proper details here.

backend/src/appointment/exceptions/validation.py
MelissaAutumn: Thanks, looks like I just needed the frontend change 😄 

Adjusted!

PR #947, opened 2025-03-17
Add string representation to models, so we can at least get their ids.

MelissaAutumn:
To help debug issues in the future with random errors. Just ids for now. 

Instead of showing `<appointment.database.models.Subscriber object at $pointer_address>` it should show something like `Subscriber: 15`.

### DIFF
--- a/backend/src/appointment/database/models.py
+++ b/backend/src/appointment/database/models.py
@@ -214,6 +214,9 @@ class Subscriber(HasSoftDelete, Base):
         hash_instance.update(str(id).encode('utf-8'))
         return hash_instance.hexdigest()
 
+    def __str__(self):
+        return f'Subscriber: {self.id}'
+
 
 class Calendar(Base):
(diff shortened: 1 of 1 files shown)


backend/src/appointment/database/models.py
```diff
@@ -477,3 +501,6 @@ class WaitingList(Base):
     invite_id = Column(Integer, ForeignKey('invites.id'), nullable=True, index=True)
 
     invite: Mapped['Invite'] = relationship('Invite', back_populates='waiting_list', single_parent=True)
+
+    def __str__(self):
+        return f'WaitingList: {self.id}'
```
rwood-moz, line 506: Just a nit, add a space => `Waiting List:`

rwood-moz (approved):
Just a tiny nit/suggestion. Cool!

devmount (commented):
Good idea! We can extend those strings in the future if we need it. Thank you!

PR #951, opened 2025-03-18
Fix GHA E2E stage test job

### LINKED ISSUE #950: GHA E2E tests job not running after stage deployment
**Describe the bug**
The E2E tests should be running after a deployment to stage (i.e. after a merge to main) however the GHA e2e stage test job is always being skipped for some reason. It did run when I merged [my initial PR into main](https://github.com/thunderbird/appointment/actions/runs/13790659737) when implementing the tests on stage, but seems to not work on other merges to main/stage deployments.

**To Reproduce**
Merege a PR to main which causes the GHA `deploy-staging.yml` workflow to run. 

**Expected behavior**
Inside the workflow is the `e2e-tests-browserstack-stage:` job which should run after the deployment to stage.

**Actual behavior**
The `e2e-tests-browserstack-stage:` job is skipped.

**Additional context**
For example see [this action on a merge to main](https://github.com/thunderbird/appointment/actions/runs/13928947368).

The `e2e-tests-browserstack-stage` job has this condition to run:

```
    if: |
      needs.deploy-stage-backend.result == 'success' || needs.build-and-deploy-stage-frontend.result == 'success'
```

Which looks good to me, but doesn't seem to work. Instead try having the E2E job run after the create release job (just need the create release and no if condition); that will also ensure there's a bit more time after the stage deployment before the tests start.

rwood-moz:
## Description of the Change
Fix the `e2e-tests-browserstack-stage` job (within the `deploy-staging.yml` GHA workflow). Remove the IF condition and run the tests every time a release has been created.

## Applicable Issues
Fixes #950

### DIFF
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -367,11 +367,7 @@ jobs:
 
   e2e-tests-browserstack-stage:
     name: e2e-tests-browserstack-stage
-    needs:
-      - deploy-stage-backend
-      - build-and-deploy-stage-frontend
-    if: |
-      needs.deploy-stage-backend.result == 'success' || needs.build-and-deploy-stage-frontend.result == 'success'
+    needs: create-release
     runs-on: ubuntu-latest
     environment: staging
     env:


MelissaAutumn (approved):
Ah yeah that should do it

Issue #949, opened 2025-03-18
Create documentation for running with podman (WIP)


CloudArtAdmins:
**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is.
No problem. This is how I'd use this project, though, so I decided to make an issue before working on a pull request.

**Describe the solution you'd like**
Documents and files that lay out the container (systemd) files for running this in podman.

**Describe alternatives you've considered**
A clear and concise description of any alternative solutions or features you've considered
This could be done with Docker, but that isn't my infrastructure, and I'd rather be without the docker sock..

I made this ticket to track my work on making this and to start some form of communication with the team to know if there is a particular design implementation that is desired, so I can build it in that preferred way, rather than just through guessing how you might want it structured.

CloudArtAdmins:
Forked here: https://github.com/CloudArtAdmins/appointment

CloudArtAdmins:
setting up quadlet format files....
Build files need to be separated from dockerfile.
"depends on" prevents direct podlet conversion of docker-compose file.

CloudArtAdmins:
In attempting to build the container image, the .env file gets pulled.
This is unusual, as we overwrite the .env file with the dockercompose file already.
Going to offer a correction PR...

MelissaAutumn:
Hello 👋 Thanks for your interest in setting up Appointment.

If you're going to be deploying Appointment in a non-dev environment you might want to take a look at our https://github.com/thunderbird/appointment/blob/main/backend/deploy.dockerfile. 

We also have https://github.com/thunderbird/appointment/blob/main/frontend/deploy.dockerfile for the frontend, but we don't use that particular dockerfile anymore as we've moved to hosting the frontend in s3 and use a cloudfront function to do some light url routing. (https://github.com/thunderbird/appointment/blob/main/tofu/modules/services/frontend-infra/main.tf#L194)

The "Depends on" is mainly for dev purposes and isn't really needed if you're deploying mysql in another docker container. A sleep could also work.

Additionally for getting the first user setup you'll need to set `APP_ALLOW_FIRST_TIME_REGISTER=True` in the backend environment / .env (depending on how you want to load your env vars.) 

Appointment is still in beta, and things can change. Feel free to submit bugs or feature requests as they come up.

CloudArtAdmins:
>The "Depends on" is mainly for dev purposes and isn't really needed if you're deploying mysql in another docker container. A sleep could also work.

[...]

PR #952, opened 2025-03-18
Unique .env files shouldn't be added to container images

CloudArtAdmins:
## Description of the Change
Removing developer used line from main code (thus assumable production code, once out of beta)

## Benefits
First time container builder reason: Yes, it echos that .env needs to be made, but this allows a successful build immediately upon cloning the repo, and more importantly..
Image/infrastructure security reason: The .env file should not be built in to the image, as one can take the image and pull the password out if anyone ever uploads an image for public pulling. This is especially true, since the docker-compose file utilizes the .env file on the host system, not on the container, anyway.
Removing it from the Dockerfile prevents many possible future accidents.

Leaving them commented, rather than removed still allows for an easy change by devs if they want these lines.

## Applicable Issues
Working on designing a method for using this with podman. #949
It requires image files to already exist, which means building the images first.
This isn't a required change, though you might find it beneficial to do long term for reasons other than "making it easier to run with podman".

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -20,10 +20,6 @@ COPY scripts/cron /etc/cron.d/appointment-cron
 RUN chmod 0644 /etc/cron.d/appointment-cron
 RUN crontab /etc/cron.d/appointment-cron
 
-# Dev only
-#RUN echo '!! If the next step fails, copy .env.example to .env in the backend folder !!'
-#COPY .env .
-
 RUN pip install --upgrade pip
 RUN pip install .'[deploy]'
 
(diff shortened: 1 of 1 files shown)


MelissaAutumn:
Hello 👋  Thanks for your contribution!

I agree we can probably link up the env file via a docker volume mount instead of directly copying it to the container. Mainly for developer convenience. 

I realize you probably don't use docker, but if you could just follow their doc and add a volume mount for the .env file in the docker-compose file, and then remove the copy .env from the dockerfile then I should be able to approve/merge this in.

I'll add some more deploy-related information in your podman ticket.

Thanks!

MelissaAutumn (approved):
Thanks!

PR #958, opened 2025-03-24
Set the E2E test browser context timezone and fix download test

### LINKED ISSUE #957: Ensure E2E tests browser context timezone is set to `America/Toronto`
There's an intermittent failure in the production sanity test, [here's an example](https://automate.browserstack.com/dashboard/v2/builds/f6a274c2030a14f4b6bca1fab368bb77fa9df1e1/sessions/1259ee4d5851d71e8e71ba77398ec31541267640) of the nightly tests failing with this issue.

The booking request page defaults to displaying time slots in the local browser context timezone. The problem is that the timezone in the production (and stage) test accounts is set to `America/Toronto` however when the tests run in BrowserStack they are running in the `America/Los_Angeles` timezone; which means the selected time slot is in the pacific time zone; and when the test navigates to the dashboard pending appointments page (in order to verify the appointment was created) the appointments are listed in EST timezone so the appointment cannot be found and the test fails.

The solution is to ensure that playwright always sets the browser context timezone to `America/Toronto` so that the share link / booking request page will also be in that same timezone, and the selected time slot will match the created appointment time in the dashboard.

rwood-moz:
Ensure the playwright browser context timezone is set for the E2E tests (fixes #957).

Also one of the settings tests (download data) fails when running in BrowserStack because BrowserStack doesn't support the `download.path` playwright method; so fixing that too.

```
    Error: Path is not available when connecting remotely. Use saveAs() to save a local copy.

      294 |     // now verify the browser download did happen successfully
      295 |     const download = await downloadPromise;
    > 296 |     const downloadedPath = await download.path(); // waits for download to finish
```

Here's a [link of the updated tests ](https://automate.browserstack.com/dashboard/v2/builds/ba55998a9a17f973261692be5995fa087b1a6004)running successfully on stage via BrowserStack in both Firefox and Chromium.

### DIFF
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -25,11 +25,11 @@ The E2E tests require an existing Appointment (and corresponding FxA account) an
     - The share link is found in Appointment => Settings => Account => My Link.
 - The tests also require an email address to be used as the appointment bookee's email address when actually requesting bookings. This is the email address entered on the `Book selection` dialog (after an appointment slot was selected on the booking share link page). Note that real Appointment emails will be sent to this email address.
 
-The tests expect that the default Appointment application settings haven't been changed for the provided test user:
+The tests expect the following Appointment application settings:
 - The user scheduling availability hasn't been changed from the default settings;
 - In the dashboard the default calendar view is the current month view; this is important so that the tests can find an available booking slot, etc.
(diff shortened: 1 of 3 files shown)


MelissaAutumn (approved):
looks good, thanks!

PR #961, opened 2025-03-25
Slug change confirmation

### LINKED ISSUE #903: Should warn users of the impact of changing their slug if they use the 'Save' button
If you change your slug and press the button next to the field:

Image

You'll get a warning telling you that your previous url will not work, but if you don't hit that button and instead hit the 'Save' button at the end of the form, you won't get this message

the message should show up regardless of which button you use to update the field.

devmount:
## Description of the Change

This PR changes the slug changed confirmation modal from showing when clicking the refresh button to clicking the save button. The confirmation appears only, if the slug was changed manually or per refresh button.

appointment_schedule_slug_change

## Benefits

Uses get warned even if they manually edited their link.

## Applicable Issues
Closes #903

### DIFF
--- a/frontend/src/components/ScheduleCreation.vue
+++ b/frontend/src/components/ScheduleCreation.vue
@@ -187,6 +187,9 @@ const getScheduleAppointment = (): ScheduleAppointment => ({
 const isFormDirty = computed(
   () => JSON.stringify(scheduleInput.value) !== JSON.stringify(referenceSchedule.value),
 );
+const isSlugDirty = computed(
+  () => scheduleInput.value.slug !== referenceSchedule.value.slug,
+);
 
 // handle notes char limit
 const charLimit = 250;
@@ -328,6 +331,7 @@ const saveSchedule = async (withConfirmation = true) => {
   // Update our reference schedule!
   referenceSchedule.value = { ...scheduleInput.value };
   revertForm(false);
+  closeModals();
 };
 
 // Update slug with a random 8 character string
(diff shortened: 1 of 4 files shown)


frontend/src/components/ScheduleCreation.vue
```diff
@@ -847,9 +851,9 @@ watch(
     :open="refreshSlugModalOpen"
     :title="t('label.refreshLink')"
     :message="t('text.refreshLinkNotice')"
-    :confirm-label="t('label.refresh')"
+    :confirm-label="t('label.save')"
```
devmount, line 854: At this place, I found "save" to be more convenient, since we continue to save data here instead of just refreshing the link.

frontend/src/locales/de.json
devmount, line 1: Just some minor improvements to the German translation.

frontend/src/components/ScheduleCreation.vue
```diff
@@ -262,6 +265,17 @@ const scheduleValidationError = (schedule: Schedule): string|null => {
...
+const refreshSlugModalOpen = ref(false);
+const showRefreshSlugConfirmation = async () => {
+  refreshSlugModalOpen.value = true;
+};
+
+const closeModals = () => {
+  savedConfirmation.show = false;
+  refreshSlugModalOpen.value = false;
+};
+
```
devmount, line 278: This needed to be here since we cannot use a function before we declare it 😇

MelissaAutumn (approved):
I'd maybe adjust it to "Saving will change your schedule's link. Any old links will no longer work." Although I'm not sure if schedule it the right word here 🤔 

Looks good regardless, thanks!

devmount:
Thank you Mel, I added your suggestion.

PR #967, opened 2025-03-28
Update python dependencies on minor/patch level

### LINKED ISSUE #939: Update Python Deps
At minimum we should update caldav which is at 1.4.0 (we're at 1.3.9)
https://github.com/python-caldav/caldav/releases

And ensure everything still works

devmount:
## Description of the Change

This brings all Python dependencies to their latest minor or patch version. I manually tested this a lot with (Nextcloud) CalDAV and Google calendars and didn't find issues. Just a type error were we forgot one positional argument.

Also fixes https://github.com/thunderbird/appointment/security/dependabot/43

For reference: We also have a few major updates available, but I didn't touch them for now.

## Applicable Issues

Closes #939


MelissaAutumn (approved):
Looks good, thanks!

PR #968, opened 2025-03-28
Fix typo in E2E stage workflow

rwood-moz:
My latest update (#966) broke the workflow as I had a typo; fixing.


MelissaAutumn (approved):
oops

PR #970, opened 2025-03-31
Update node dependencies

devmount:
## Description of the Change

This PR upgrades major versions of:

- pinia
- vue-i18n
- jsdom
- ua-parser-js
- vueuse
- eslint

Eslint was migrated to use flat config (as done in services-ui) and new linter warnings got fixed. Technically we need to update Tailwindcss and the two Sentry dependencies as well. But the former gets replaced by scoped styles (so I don't know if updating to v4 is worth the effort) and for the latter I cannot really debug if there are issues with it.

Oh and I sneaked in a fix in the German language file 😇


MelissaAutumn (approved):
Thanks, seems to work on my end. I'll punt it to stage to see if there's any live issues.

Issue #969, opened 2025-03-28
Intermittent login failure (via FxA)


rwood-moz:
**Describe the bug**
Once and awhile after signing into Appointment (via FxA) instead of the dashboard appearing, there is a big red `⛔ Sorry, this page requires you to be logged in` error.  

**To Reproduce**
Sign into Appointment on production or stage via FxA; enter correct FxA credentials.

**Expected behavior**
The Appointment dashboard appears.

**Actual behavior**
A big red `⛔ Sorry, this page requires you to be logged in` error is displayed.

**Screenshots**
This error is causing some intermittent failures of the E2E tests.

Here's a [BrowserStack E2E test link](https://automate.browserstack.com/dashboard/v2/builds/4d2e21e7553ab46ad8b08c8514e9f67196c7bb26/sessions/d264685a38bba486561da9903aaab538b8bbd8d6) to an example of this error happening on Chromium on production.

And here's a [Browserstack E2E test link](https://automate.browserstack.com/dashboard/v2/builds/8896d29839b98655a05c17fe11cddc7fe0f77b00/sessions/fad9998ca8a2c6419199153f193344319820b7a3) to an example of this error happening on Firefox on stage.

<img width="892" alt="Image" src="https://github.com/user-attachments/assets/60a3b8c2-cbba-4f43-903f-b90f3556959d" />

**Additional context**
Perhaps this issue will magically be resolved when we switch Appointment over to using TB Accounts for auth, but wanted to submit this anyway in case it still persists.

aaspinwall:
I have a similar issue in send. I narrowed it down to the site calling `window.open` for fxa login. The problem is that the `window.close` is called immediately. So even if you log in correctly, the site won't run the rest of the logic because it assumes the window was closed without a successful login.

PR #972, opened 2025-03-31
Hopeful fix for being logged out after logging in

MelissaAutumn:
Related #969 

This is just for staging right now. Based off the sentry data the user's minimum time is one second after issued at time. The only thinig that sets the minimum time is the webhooks. So I guess the webhooks are getting fired off after we ask them to destroy a token. If someone logs in just at the right time, they can be immediately logged out lol.

Neither systems are long for this world, so hopefully this reduces the issue until we switch auth over.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -146,10 +146,6 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install -e .'[test]'
 
-      - name: Backend lint check
-        run: |
-          cd ./backend && ruff check
-
       - name: Test with pytest
         working-directory: ./backend
         run: |
@@ -161,6 +157,10 @@ jobs:
         continue-on-error: true
         run: |
           coverage report
+
+      - name: Backend lint check
+        run: |
+          cd ./backend && ruff check
 
   validate-frontend:
     needs: detect-changes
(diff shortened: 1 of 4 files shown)


.github/workflows/validate.yml
```diff
@@ -162,6 +158,10 @@ jobs:
         run: |
           coverage report
 
+      - name: Backend lint check
+        run: |
+          cd ./backend && ruff check
```
devmount, line 163: Running the linter after running tests seems reasonable for me.

backend/src/appointment/controller/auth.py
```diff
@@ -24,7 +24,8 @@ def logout(
 ):
     """Sets a minimum valid issued at time (time). This prevents access tokens issued earlier from working."""
     if deny_previous_tokens:
-        subscriber.minimum_valid_iat_time = datetime.datetime.now(datetime.UTC)
+        # Reduce the parry-window for webhooks. This will hopefully prevent the login immediately logged out issue.
+        subscriber.minimum_valid_iat_time = datetime.datetime.now(datetime.UTC) - datetime.timedelta(minutes=1)
```
devmount, line 28: Good workaround! Should we mark this as FIXME or TODO and recheck, when we switched Auth?

devmount (approved):
Thanks for investigating this! Just one nit below. (and I saw there is a backend test failing, but that's weird and seems unrelated?)

PR #981, opened 2025-04-04
Start of week user setting

### LINKED ISSUE #589: Make the start of the week configurable

devmount:
## Description of the Change

This change add a setting for the user to set the start of week to either Sunday or Monday (see #818).

image

## Benefits

Users can set their own start of week and it is taken into account on the dashboard as well as in the week day selection in the schedule configuration sidebar.

## Applicable Issues

Fixes #589 
Code is mostly stolen from the changes already implemented by @MelissaAutumn in #818


frontend/src/components/SettingsGeneral.vue
```diff
@@ -95,6 +99,21 @@ const timezones = Intl.supportedValuesOf('timeZone');
...
   <div class="pl-6">
     <div class="text-xl">{{ t('heading.dateAndTimeFormatting') }}</div>
+    <div class="mt-6 pl-6">
+      <div class="text-lg">{{ t('label.startOfTheWeek') }}</div>
+      <label class="mt-4 flex items-center pl-4">
+        <div class="w-full max-w-2xs">{{ t('label.startOfTheWeek') }}</div>
+        <select
+          v-model="user.data.settings.startOfWeek"
+          class="w-full max-w-sm rounded-md"
+          data-testid="settings-general-star-of-week-select"
```
rwood-moz, line 109: typo, missing the t in settings-general-`star`-of-week-select

rwood-moz (approved):
Just the one typo (see inline). Looks great!

I gave it a try locally and it worked great (including changing the start day on the share link/request booking page too). I also ran the E2E tests against your branch and they passed fine.

I also created #983 to update the E2E tests for this new feature.

frontend/src/components/SettingsGeneral.vue
devmount: Good catch, fixed. Thanks!

PR #994, opened 2025-04-17
Username only URLs

### LINKED ISSUE #879: Allow users to set https://apmt.day/username/
By default, our booking urls are https://apmt.day/username/randomString, which was done to counter malicious behaviour around guessing people's booking urls and making requests, and to futureproof our approach to multi-schedule support.

It would be ideal to let users select if they want to have a single https://apmt.day/username/ url, and also to let them select which schedule it points to once we have multi-schedule support. Calendly allows these single urls, so it seems more like a feature we're lacking imo

devmount:
## Description of the Change

Sorry, this change turned out to be bigger than expected 😅  Here is what I changed:

- Appointment now allows to configure an empty slug for the custom user URL
- The username can be set as part of the URL in FTUE
- The slug can be changed (already possible) and removed (new) in the account settings
- @malini @solangevalverde I reused our existing input component that allows for a prefix. This isn't exactly like shown in the wireframes, so let me know if you want to change that still
- I also converted all buttons, text inputs and selects on the settings page to the new tbpro components

appointment_settings_slug
appointment_ftue_username

@laurelterlesky

I also added a danger variant for our base button component:

image

## Applicable Issues

Closes #879

### DIFF
--- a/backend/src/appointment/controller/auth.py
+++ b/backend/src/appointment/controller/auth.py
@@ -84,5 +84,10 @@ def schedule_links_by_subscriber(db, subscriber: models.Subscriber):
 
     # Empty space at join is for trailing slash!
     return list(
-        map(lambda sch: '/'.join([short_url, url_safe_username, urllib.parse.quote_plus(sch.slug), '']), schedules)
+        map(
+            lambda sch:
+                '/'.join([short_url, url_safe_username, urllib.parse.quote_plus(sch.slug), '']) if sch.slug
+                else '/'.join([short_url, url_safe_username, '']),
+            schedules
+        )
     )
--- a/backend/src/appointment/controller/auth.py
+++ b/backend/src/appointment/controller/auth.py
@@ -82,12 +82,12 @@ def schedule_links_by_subscriber(db, subscriber: models.Subscriber):
 
     url_safe_username = urllib.parse.quote_plus(subscriber.username)
 
-    # Empty space at join is for trailing slash!
-    return list(
-        map(
-            lambda sch:
-                '/'.join([short_url, url_safe_username, urllib.parse.quote_plus(sch.slug), '']) if sch.slug
-                else '/'.join([short_url, url_safe_username, '']),
-            schedules
-        )
-    )
+    return list(map(lambda sch: schedule_link(sch, short_url, url_safe_username), schedules))
+
+
+def schedule_link(schedule: models.Schedule, base: str, username: str):
+    """helper to generate a schedule link depending on the existence of a schedule slug"""
+    if schedule.slug:
+        return f"{base}/{username}/{urllib.parse.quote_plus(schedule.slug)}/"
+    else:
+        return f"{base}/{username}/"
--- a/backend/src/appointment/controller/auth.py
+++ b/backend/src/appointment/controller/auth.py
@@ -49,8 +49,9 @@ def sign_url(url: str):
     return signature
 
 
-def signed_url_by_subscriber(subscriber: schemas.Subscriber):
-    """helper to generated signed url for given subscriber"""
+def user_link_by_subscriber(subscriber: models.Subscriber):
+    """Returns a user link based off the subscriber's username
+    Note that this contains a trailing slash"""
     short_url = os.getenv('SHORT_BASE_URL')
     base_url = f"{os.getenv('FRONTEND_URL')}/user"
 
@@ -60,34 +61,29 @@ def signed_url_by_subscriber(subscriber: schemas.Subscriber):
 
     url_safe_username = urllib.parse.quote_plus(subscriber.username)
 
+    return f'{short_url}/{url_safe_username}/'
+
+
+def signed_url_by_subscriber(subscriber: schemas.Subscriber):
+    """helper to generated signed url for given subscriber"""
+    base_url = user_link_by_subscriber(subscriber)
+
     # We sign with a different hash that the end-user doesn't have access to
     # We also need to use the default url, as short urls are currently setup as a redirect
-    url = f'{base_url}/{url_safe_username}/{subscriber.short_link_hash}'
+    url = ''.join([base_url, subscriber.short_link_hash])
 
     signature = sign_url(url)
 
     # We return with the signed url signature
-    return f'{short_url}/{url_safe_username}/{signature}'
+    return ''.join([base_url, signature])
 
 
-def schedule_links_by_subscriber(db, subscriber: models.Subscriber):
+def schedule_slugs_by_subscriber(db, subscriber: models.Subscriber) -> dict:
     # Generate some schedule links
     schedules = repo.schedule.get_by_subscriber(db, subscriber_id=subscriber.id)
-    short_url = os.getenv('SHORT_BASE_URL')
-    base_url = f"{os.getenv('FRONTEND_URL')}/user"
 
-    # If we don't have a short url, then use the default url with /user added to it
-    if not short_url:
-        short_url = base_url
+    slugs = {}
+    for schedule in schedules:
+        slugs[schedule.id] = urllib.parse.quote_plus(schedule.slug) if schedule.slug else None
 
-    url_safe_username = urllib.parse.quote_plus(subscriber.username)
-
-    return list(map(lambda sch: schedule_link(sch, short_url, url_safe_username), schedules))
-
-
-def schedule_link(schedule: models.Schedule, base: str, username: str):
-    """helper to generate a schedule link depending on the existence of a schedule slug"""
-    if schedule.slug:
-        return f"{base}/{username}/{urllib.parse.quote_plus(schedule.slug)}/"
-    else:
-        return f"{base}/{username}/"
+    return slugs
--- a/backend/src/appointment/routes/schedule.py
+++ b/backend/src/appointment/routes/schedule.py
@@ -113,7 +113,7 @@ def create_calendar_schedule(
 
     # The first schedule currently is initialized without a slug to provide a username only availability link.
     # If we already have at least one schedule and slug isn't provided, give them the last 8 characters from a uuid4.
-    if len(repo.schedule.get_by_subscriber(db, subscriber.id) > 0) and not schedule.slug:
+    if len(repo.schedule.get_by_subscriber(db, subscriber.id)) > 0 and not schedule.slug:
         slug = repo.schedule.generate_slug(db, db_schedule.id)
         if not slug:
             # A little extra, but things are a little out of place right now..
--- a/backend/src/appointment/routes/schedule.py
+++ b/backend/src/appointment/routes/schedule.py
@@ -111,13 +111,14 @@ def create_calendar_schedule(
 
     db_schedule = repo.schedule.create(db=db, schedule=schedule)
 
-    # If slug isn't provided, give them the last 8 characters from a uuid4
-    # TODO: We might want to set initial slug to None if the user should start with an username only url
-    slug = repo.schedule.generate_slug(db, db_schedule.id)
-    if not slug:
-        # A little extra, but things are a little out of place right now..
-        repo.schedule.hard_delete(db, db_schedule.id)
-        raise validation.ScheduleCreationException()
+    # The first schedule currently is initialized without a slug to provide a username only availability link.
+    # If we already have at least one schedule and slug isn't provided, give them the last 8 characters from a uuid4.
+    if len(repo.schedule.get_by_subscriber(db, subscriber.id) > 0) and not schedule.slug:
+        slug = repo.schedule.generate_slug(db, db_schedule.id)
+        if not slug:
+            # A little extra, but things are a little out of place right now..
+            repo.schedule.hard_delete(db, db_schedule.id)
+            raise validation.ScheduleCreationException()
 
     return db_schedule
 
--- a/backend/src/appointment/routes/schedule.py
+++ b/backend/src/appointment/routes/schedule.py
@@ -112,6 +112,7 @@ def create_calendar_schedule(
     db_schedule = repo.schedule.create(db=db, schedule=schedule)
 
     # If slug isn't provided, give them the last 8 characters from a uuid4
+    # TODO: We might want to set initial slug to None if the user should start with an username only url
     slug = repo.schedule.generate_slug(db, db_schedule.id)
     if not slug:
         # A little extra, but things are a little out of place right now..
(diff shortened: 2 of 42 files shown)


backend/src/appointment/routes/schedule.py
```diff
@@ -112,6 +112,7 @@ def create_calendar_schedule(
     db_schedule = repo.schedule.create(db=db, schedule=schedule)
 
     # If slug isn't provided, give them the last 8 characters from a uuid4
+    # TODO: We might want to set initial slug to None if the user should start with an username only url
     slug = repo.schedule.generate_slug(db, db_schedule.id)
```
devmount, line 116: I didn't know the flow here, currently there is a slug generated when first creating a schedule.

backend/src/appointment/utils.py
```diff
@@ -60,28 +60,33 @@ def decrypt(value):
...
+    if split_path is None or len(split_path) == 0 or len(split_path) > 3:
         return False
-    # If we have more than two entries, grab the last two
-    elif len(split_path) > 2:
-        split_path = split_path[-2:]
+
+    # Normalize short and long urls to only username and slug/signature (remove the /user/ segment)
+    # FIXME: Handle edge case: A user with username='user' might make trouble here
+    if len(split_path) > 1 and split_path[0] == 'user':
+        split_path.pop(0)
```
devmount, line 74: I didn't find a better way to handle that. The problem is, that we might have a `/user/` segment or might not.

backend/test/unit/test_auth_dependency.py
devmount, line 1: I added tests for all 4 URL segment variations

frontend/src/tbpro/elements/BaseButton.vue
devmount, line 1: I added a danger variant. We might want to port that change to the services-ui repo some time.

frontend/src/tbpro/elements/TextInput.vue
```diff
@@ -75,7 +75,7 @@ const onChange = () => {
   <label class="wrapper" :for="name">
     <span class="label">
       <slot/>
-      <span v-if="required && model?.length === 0" class="required">*</span>
+      <span v-if="required && $slots.default && model?.length === 0" class="required">*</span>
```
devmount, line 78: Only show the required asterisk when a label is set.

frontend/src/tbpro/elements/TextInput.vue
```diff
@@ -105,9 +105,6 @@ const onChange = () => {
     <span v-else-if="help" class="help-label">
       {{ help }}
     </span>
-    <span v-else class="help-label">
-      <!-- Empty space -->
-    </span>
```
devmount, line 110: This was causing issues in the layout. I hope it's fine that I deleted it 😇

devmount (commented):
Some pointers.

rwood-moz (commented):
Just my opinion, but from a user point of view I find it a bit confusing that on the FTUE dialog, the username field contains the host name inside the same field. At first it looks like I should be able to edit the actual host name as well since it's all listed inside one big input box (but in reality of course I can only edit the last part of the URL - the username.

Screenshot 2025-04-21 at 10 18 37 AM

I see in the issue #879 the wireframe shows the host being like a label to the username input box:

Screenshot 2025-04-21 at 10 23 56 AM

Unless maybe that was updated after?

rwood-moz (commented):
Maybe I am misunderstanding perhaps, but on the FTUE when I choose my username, the share link is still being created with a slug. I thought that it would only be my username with no slug?

Screenshot 2025-04-21 at 10 53 58 AM

Screenshot 2025-04-21 at 10 54 24 AM

Screenshot 2025-04-21 at 10 54 55 AM

After the FTUE I was able to go into account settings and remove the slug fine, and the updated share link without the slug worked; but since the FTUE said that the username would become my share link I wasn't expecting the slug to be on the end. Maybe it should be added in the FTUE username input field too then if it is going to be there by default?

rwood-moz (changes requested):
FYI I'm seeing an error in the backend logs when running the stack locally:

```
ckend-1   |     |   File "/usr/local/lib/python3.12/site-packages/anyio/to_thread.py", line 56, in run_sync
backend-1   |     |     return await get_async_backend().run_sync_in_worker_thread(
backend-1   |     |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
backend-1   |     |   File "/usr/local/lib/python3.12/site-packages/anyio/_backends/_asyncio.py", line 2470, in run_sync_in_worker_thread
backend-1   |     |     return await future
backend-1   |     |            ^^^^^^^^^^^^
backend-1   |     |   File "/usr/local/lib/python3.12/site-packages/anyio/_backends/_asyncio.py", line 967, in run
backend-1   |     |     result = context.run(func, *args)
backend-1   |     |              ^^^^^^^^^^^^^^^^^^^^^^^^
backend-1   |     |   File "/app/appointment/routes/metrics.py", line 123, in ftue_step
backend-1   |     |     posthog.set(distinct_id=subscriber.unique_hash, properties=payload)
backend-1   |     |     ^^^^^^^^^^^
backend-1   |     | AttributeError: 'NoneType' object has no attribute 'set'
```

rwood-moz (changes requested):
A corner case, if I go into settings and change my username so that it has a space at the start, it accepts the space (the URL puts `+` in for the space) so the share link works ie:

Screenshot 2025-04-21 at 11 09 24 AM

However if you go back into settings then the `my link` field is showing the host name twice. Should we allow spaces at the start of user names?
Screenshot 2025-04-21 at 11 09 19 AM

devmount:
@rwood-moz Great feedback, thanks! I'll take a look at that.

backend/src/appointment/controller/auth.py
```diff
@@ -84,5 +84,10 @@ def schedule_links_by_subscriber(db, subscriber: models.Subscriber):
 
     # Empty space at join is for trailing slash!
     return list(
-        map(lambda sch: '/'.join([short_url, url_safe_username, urllib.parse.quote_plus(sch.slug), '']), schedules)
+        map(
+            lambda sch:
+                '/'.join([short_url, url_safe_username, urllib.parse.quote_plus(sch.slug), '']) if sch.slug
+                else '/'.join([short_url, url_safe_username, '']),
+            schedules
```
MelissaAutumn, line 91: I think we can split this to a function call. It's getting a little chunky.

backend/src/appointment/utils.py
MelissaAutumn: We should create a list of "unavailable" usernames. Like user, username, user_name, admin, thunderbird, etc...

backend/src/appointment/utils.py
```diff
@@ -60,28 +60,33 @@ def decrypt(value):
...
+    # Check for general validity of the path
+    if split_path is None or len(split_path) == 0 or len(split_path) > 3:
         return False
-    # If we have more than two entries, grab the last two
-    elif len(split_path) > 2:
-        split_path = split_path[-2:]
+
+    # Normalize short and long urls to only username and slug/signature (remove the /user/ segment)
+    # FIXME: Handle edge case: A user with username='user' might make trouble here
+    if len(split_path) > 1 and split_path[0] == 'user':
```
MelissaAutumn, line 73: Why did this need to change to account for /user/ part? Was this not working as intended? The /user/ portion was always there iirc.

frontend/src/components/SettingsAccount.vue
```diff
@@ -76,6 +92,16 @@ const errorDisplayName = ref<string>(null);
 
 // Save user data
 const updateUser = async () => {
+  // First update the slug if it was changed
+  if (activeSlug.value !== user.mySlug) {
+    const response = await schedule.updateFirstSlug(activeSlug.value);
+    // eslint-disable-next-line no-prototype-builtins
+    if (response.hasOwnProperty('error')) {
+      // TODO: error message is in data
```
MelissaAutumn, line 100: We'll need to show this

frontend/src/components/SettingsGeneral.vue
```diff
@@ -4,14 +4,38 @@ import { inject, watch, computed } from 'vue';
...
+import SelectInput from '@/tbpro/elements/SelectInput.vue';
 
 // component constants
 const { t, locale, availableLocales } = useI18n({ useScope: 'global' });
 const call = inject(callKey);
 const user = createUserStore(call);
 const isoWeekdays = inject(isoWeekdaysKey);
+
+const localeOptions = availableLocales.map((l) => ({
+  label: l.toUpperCase() + ' — ' + t('locales.' + l),
```
MelissaAutumn, line 16: Use template strings please!

frontend/src/components/SettingsGeneral.vue
```diff
@@ -4,14 +4,38 @@ import { inject, watch, computed } from 'vue';
...
+  value: l,
+}));
+
+const colourSchemeOptions = Object.values(ColourSchemes).map((c) => ({
+  label: t('label.' + c),
```
MelissaAutumn, line 21: Also template string these!

frontend/src/models.ts
```diff
@@ -420,7 +420,7 @@ export type TimeFormatted = Time<string>;
 
 export type SelectOption = {
   label: string;
-  value: number;
+  value: number|string;
```
MelissaAutumn, line 423: tbh this should really just be a string, and whatever needs to consume value should parseInt it if needed.

frontend/src/tbpro/elements/BaseButton.vue
MelissaAutumn: Definitely, we should check in with @solangevalverde / @laurelterlesky for design nits. Can you add a screenshot for 'em?

MelissaAutumn (changes requested):
Thanks! It's still early morning for me, but I have some edit requests!

Also I must have missed the github issue decision, but I feel like we should have removed the link from settings and kept it in the calendar settings. Because once we add another calendar schedule to the mix we'll have to move it back there lol!

But that's all good, thanks for the hard work!

devmount:
@rwood-moz I had a look at all your comments and here is what I did now:

1. Our TextInput component now can handle "outer prefixes". The URL now isn't part of the input field anymore as shown in the wireframe mockups.
2. Yes, the flow wasn't clear to me there. I now changed it so that the first schedule created won't have a slug
3. That's weird, that shouldn't be related to any of my changes, but yes, I can see it too. Anyway, I added a check to prevent that error.
4. Ups, that's a weird one too. I added a trim for whitespaces at both ends here. @MelissaAutumn @malini We might want to discuss to forbid whitespaces in slugs in general (currently they are converted to `+`).

Feel free to have a look again and thank you for your thorough testing!

backend/src/appointment/routes/schedule.py
devmount: This is now changed to only generate slugs if we have more than 1 schedule and no slug was initially given.

backend/src/appointment/utils.py
devmount: It wasn't working for schedules without a slug on my end, so I thought this was caused by my local instance not having configured a short URL and always using long URLs. I made those changes to support both application states (having a short URL and having none), but please let me know if I got sth fundamentally wrong here 😅

backend/src/appointment/utils.py
devmount: Agreed! Added https://github.com/thunderbird/appointment/issues/1004 for this

frontend/src/components/SettingsGeneral.vue
devmount: Of course!

frontend/src/models.ts
devmount: No problem, I can look into this. Just thought to keep it compatibel, but I totally agree.

frontend/src/tbpro/elements/BaseButton.vue
devmount: Done

devmount:
@MelissaAutumn I implemented all of your feedback too. Feel free to look again if you have the time. Thanks!

rwood-moz:
Thanks for the updates @devmount! I really like how the input field is now separated in the share link.

I'm seeing something strange when running locally though, by default the text input has part of the URL (not just the user name):

Label: `http://localhost:8080/user/robtbtest/` input field: `80/user/robtbtest/1b84864b`

I wonder if anyone else is seeing this too?

Note that the E2E tests fail on this PR branch (but they pass on main). I notice your branch doesn't have the latest main branch rebased so that might be why the E2E tests are failing on your branch, unless there is some other change breaking the tests.

rwood-moz:
Thanks for the main rebase, I pulled your latest branch and ran the E2E tests. The following tests fail on this branch:

Screenshot 2025-04-24 at 11 37 49 AM

The failures are because:

- The dom for the theme select has changed; the test-data-id is gone and also the select name is incorrect (it's now called 'timezone':

Screenshot 2025-04-24 at 11 38 50 AM

<br />
<br />

- The dom for the timezone select has changed (the test-data-id is gone):

Screenshot 2025-04-24 at 11 39 34 AM

<br />
<br />

- And the account settings test fails as it is trying to grab the share link dom element which has changed, the test fails here:

https://github.com/thunderbird/appointment/blob/main/test/e2e/pages/settings-page.ts#L264
Which uses this test-data-id locator which is also gone
`this.profileMyLinkInput = this.page.getByTestId('settings-account-mylink-input');`

So that will need to be updated to use the new share link locator. :)

MelissaAutumn:
Hmm I went ahead and rebased this PR but it seems like the merge commits have broken the chain. I will need to mess with this a little more.

MelissaAutumn:
Okay I think I fixed it. Had to remove both merge commits. It no longer thinks changes done on main are new. 

@devmount please backup your local branch (aka branch that to a new one locally) and then hard reset to this remote branch and see if I've accidentally removed any changes. Apologies!

MelissaAutumn:
I've pushed some changes to split up retrieving slugs and urls. This should work now I think? This fixes the issue where if you had a slug/hash in the quick link/my link thingy it would spit out the entire url. 

@rwood-moz want to give it a spin?

rwood-moz:
> I've pushed some changes to split up retrieving slugs and urls. This should work now I think? This fixes the issue where if you had a slug/hash in the quick link/my link thingy it would spit out the entire url.
> 
> @rwood-moz want to give it a spin?

Yep that fixed it (only the existing slug appears in the text input now and not the entire URL), nice! 🥳

MelissaAutumn:
Okay I think I've got everything in place, even fixed a random test failure.

rwood-moz:
FYI unfortunately this merge broke the Appointment E2E tests (I [noted above](https://github.com/thunderbird/appointment/pull/994#issuecomment-2828113374) that it would break them but I think my comment must have been missed and it was merged hahaha

I created #1009 to address the broken tests.

PR #997, opened 2025-04-22
Don't pass our scopes to the zoom auth url (Fixes #996)

### LINKED ISSUE #996: Can't login to zoom app
Our oauth library passes in scopes automatically when we ask for a authorization url. Zoom suddenly hates this. I'm not sure why, but they do.

MelissaAutumn:
Fixes #996 

I'm not sure why they hate having scopes in the url, since our third-party oauth library does it for us. So it's within spec, but at least it doesn't seem to affect our meeting link creation abilities. 

Tested and working fine locally.

### DIFF
--- a/backend/src/appointment/controller/apis/zoom_client.py
+++ b/backend/src/appointment/controller/apis/zoom_client.py
@@ -82,7 +82,10 @@ class ZoomClient:
         pass
 
     def get_redirect_url(self, state):
+        # Ensure we pass "None" to scope, as zoom doesn't accept scopes here...for some reason?
+        self.client.scope = None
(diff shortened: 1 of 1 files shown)


backend/src/appointment/controller/apis/zoom_client.py
```diff
@@ -82,7 +82,10 @@ def setup(self, subscriber_id=None, token=None):
         pass
 
     def get_redirect_url(self, state):
+        # Ensure we pass "None" to scope, as zoom doesn't accept scopes here...for some reason?
+        self.client.scope = None
         url, state = self.client.authorization_url(self.OAUTH_AUTHORIZATION_URL, state=state)
```
MelissaAutumn, line 87: Internally self.client.authorization_url passes in self.scope to the url.

PR #998, opened 2025-04-22
Only use one set of scopes, and catch any scope errors that may occur.

### LINKED ISSUE #996: Can't login to zoom app

MelissaAutumn:
Fixes #996 for realies.

Basically we're experiencing a scope change error: https://thunderbird.sentry.io/issues/6561374703/?project=4505428124827648&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&stream_index=0

Where we seem to have initialized our app with the old scopes instead of the new scopes...which doesn't make sense because we set an env var to use new scopes. Since everything should be using new scopes, let's just use those and delete the old ones.

This also catches the scope change error and sends the user back an error message.

### DIFF
--- a/backend/src/appointment/controller/apis/zoom_client.py
+++ b/backend/src/appointment/controller/apis/zoom_client.py
@@ -1,6 +1,5 @@
 import json
 import logging
-import os
 import time
 
 import sentry_sdk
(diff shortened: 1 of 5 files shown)


MelissaAutumn:
Ah this is actually configuration error. I'll keep this PR up, but it's not as urgent.

devmount (approved):
lgtm, thanks!

PR #1002, opened 2025-04-23
Sync up tofu's mysql version with rds

MelissaAutumn:
Ah the real cause of the failed deployments. It's weird how the error in tofu/terraform is reverse. But RDS is already running on *.40, so we should specify that.


radishmouse (approved):
Looks +`0.0.05` better :joy:

Issue #1015, opened 2025-05-06
Dashboard availability/settings panel slow loading time


rwood-moz:
**Describe the bug**
Sometimes the availability/settings panel on the main dashboard page is taking several seconds to load. The E2E tests detected this issue as even after 10 seconds sometimes the dashboard panel doesn't appear.

**To Reproduce**

- Load the main dashboard
- On the top right click on the `Booking` tab
- Click back on the `Dashboard` tab

**Expected behavior**
Expect the dashboard page to fully load including the availability/settings panel on the right.

**Actual behavior**
Sometimes the dashboard loads but the availability/settings panel is missing. It does eventually load.

**Screenshots**
<img width="995" alt="Image" src="https://github.com/user-attachments/assets/1e700ce7-0f53-44d8-b6ca-12b0e330e6c8" />

rwood-moz:
Created separate issue #1016 to update the E2E tests to have a longer loading wait time so that the tests will stop failing.

PR #1017, opened 2025-05-06
Increase E2E test wait time when loading dashboard

rwood-moz:
Update the E2E tests to wait longer for the dashboard to fully load because sometimes the availability / settings panel on the right side is taking a very long time to load (see issue #1015).

Also using this PR to increase the wait time allowed for the FxA sign-in page to load as I notice once and awhile the tests timeout waiting for FxA.


devmount (approved):
Looking good, thanks for reporting this dashboard loading delay bug.

PR #1020, opened 2025-05-12
Fix type errors

### LINKED ISSUE #1019: Messed up SelectOption value type
**Describe the bug**
After #994 landed, we have some messed up types for the SelectOption type.

**Expected behavior**
TS being fully satisfied.

**Actual behavior**
TS complaining about "int" not being a known data type.

devmount:
## Description of the Change

This PR fixes a wrong type with the SelectOption value. I also reverted some of the stringified number select values where appropriate. I think it's best if we keep the `string | number` type for select values, since there are valid use cases for both imo.

## Applicable Issues

Closes #1019

### DIFF
--- a/frontend/src/components/FTUE/SetupSchedule.vue
+++ b/frontend/src/components/FTUE/SetupSchedule.vue
@@ -38,15 +38,15 @@ const { timeToBackendTime, timeToFrontendTime } = scheduleStore;
 
 const calendarOptions = computed<SelectOption[]>(() => connectedCalendars.value.map((calendar) => ({
   label: calendar.title,
-  value: calendar.id.toString(),
+  value: calendar.id,
 })));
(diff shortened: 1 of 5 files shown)


rwood-moz (approved):
LGTM and I ran the E2E tests against this branch, all passed fine.

devmount:
@rwood-moz Awesome! Thanks for always also checking the E2E tests 👏🏻

MelissaAutumn (commented):
I think last time we touched this type the weekday selector broke because we store as one type instead of another.

I'm trying to find the actual commit I undid a change here, but can you confirm the weekday selection bubbles work?

devmount:
I tested this, yes. Locally the weekday selection works just fine. This also fixes a bug where the slots preview was not properly shown due to comparison of string and number.

MelissaAutumn (approved):

PR #1025, opened 2025-05-14
E2E Test Improvements

rwood-moz:
Some E2E test improvements:

- At the start of each E2E settings test, ensure the Appointment settings are set to default values. This ensures they are in a known state in case settings were changed by previous test failures/issues.
- Split the settings tests up into multiple tests to have smaller individual test run times
- Add some scrolling to elements so they're visible in browserstack logs (good for debugging)
- Add some small pauses when changing some of the settings values


rwood-moz:
Ok looks like this will clean up some intermittent test failures that have been happening lately. I ran the updated tests 10x in BrowserStack (some on stage and some on prod) and all of the tests passed each time. I also ran them locally as well multiple times, and all green.

Screenshot 2025-05-14 at 4 03 33 PM

ryanjjung (approved):

Issue #422, opened 2024-05-22
Allow users to fine-tune their availability


malini:
**Is your feature request related to a problem? Please describe.**
Right now, you can only select a single time-frame and what days to apply it on. There isn't a way to block out certain days, or shift availability on certain days vs others, or designate timeslots in a day that you're willing to be booked

**Describe the solution you'd like**
We should work with design to incorporate an easy-to-use configuration form for variable availability on a more granular level. Details on what to fine-tune are TBD.

ryanleesipes:
Agreed. I'm just not using Fridays right now because I can't mark the morning only for that day. 🤷

malini:
Assigning this to myself to do a little competitive analysis soon

laurelterlesky:
Design is working on this. First pass at this ticket for Beta 3 will be:
- Required: Check boxes for each day, so user can easily select which days they want
- Explore unlimited Interval support
- Explore copying availability and pasting into another day of the week

Out-of-Scope - date specific overrides (we'll look at this during MVP)

solangevalverde:
Wireframe delivery 💌📬 

![Image](https://github.com/user-attachments/assets/0d7b6f89-bde6-4a00-bdfe-bcad27e3a956)

devmount:
FYI: The first version of this is on stage, where you can set one custom availability per day. I'm now working on the next iteration to provide multiple slots per day.

devmount:
@malini @laurelterlesky The following topics / edge cases that were found during review or just implemented by me should be discussed / decided. I'm just throwing them in here, so they don't get lost. There might come more.

1. Update of the dashboard calendar "availability preview entries" isn't implemented yet (since we wanted to remove the calendar iirc). So those entries won't update when setting fine-tune availability entries.
![Image](https://github.com/user-attachments/assets/c503acec-98ec-4708-b4e8-e3f757ac89c0)
2. Removal of the last availability entry for a weekday disables that day.
![Image](https://github.com/user-attachments/assets/33cee884-e591-499c-aa33-821c868f45f8)
3. Availability can be added until 11pm (for validation reasons, but that can be changed). Do we want to add slots until midnight?
![Image](https://github.com/user-attachments/assets/d229fe2b-3ffb-460b-9191-dc9b86f5d193)
4. The copy function shows every weekday as disabled that was already disabled in the configuration as well as the day that serves as the copy source.
[...]

PR #1027, opened 2025-05-16
Fine-tune availabilities

devmount:
## Description of the Change

This PR enables users to set custom times for each day of week for their schedule:

image
image

## Benefits

Days can now have custom time limits.

## Applicable Issues

This is the first iteration on #422


frontend/src/components/ScheduleCreation.vue
```diff
@@ -137,7 +144,9 @@ onMounted(() => {
   generateZoomLink.value = scheduleInput.value.meeting_link_provider === MeetingLinkProviderType.Zoom;
 
   // Set a new reference
-  referenceSchedule.value = { ...scheduleInput.value };
+  // Sidenote: I'm too dumb to properly clone this Object without referencing the same availability array. So for now
+  // this is the good ol' JSON workaround (structuredClone, Object assign or spread won't work here)
+  referenceSchedule.value = JSON.parse(JSON.stringify(scheduleInput.value));
```
devmount, line 149: I had a lot of issues due to the availabilities array being referenced from all over the components. The only thing that helped was the JSON workaround. Let me know, if you have a more proper way to deep clone that object 😇

frontend/src/elements/AlertBox.vue
```diff
@@ -123,7 +123,6 @@ const toggleDetails = () => {
   .title {
     font-weight: 700;
     line-height: 1.5;
-    white-space: pre;
```
devmount, line 126: This fixes an issue with long error titles bleeding out of the container.

frontend/src/locales/de.json
devmount, line 1: Sorry, that was me sorting the JSON keys again 😂

frontend/src/models.ts
```diff
@@ -177,12 +177,19 @@ export type ExternalConnectionCollection = {
...
+  schedule_id?: number;
+  day_of_week: number;
+  start_time: string;
+  end_time: string;
+  slot_duration?: number;
 };
 
+export type AvailabilitySet = {
+  [key:number]: Availability[];
+}
```
devmount, line 191: I already prepared this for the next step, when we have multiple availability entries per day.

frontend/src/tbpro/elements/TextInput.vue
```diff
@@ -95,6 +96,7 @@ const onChange = () => {
           :maxLength="maxLength"
           @invalid.prevent="onInvalid"
           @change="onChange"
+          @blur="emit('blur')"
```
devmount, line 99: lmk if there's a better way to bubble those events up

devmount (commented):
Some pointers on my changes. Thanks for taking the time to look over them!

backend/src/appointment/database/repo/availability.py
```diff
@@ -0,0 +1,77 @@
...
+from .. import models, schemas, repo
+
+
+""" AVAILABILITY repository functions
+"""
+
+
+def get(db: Session, availability_id: int) -> models.Availability | None:
+    """retrieve availability by id"""
+    if availability_id:
```
MelissaAutumn, line 16: I'm sure this is just a copy and paste, but since we're typing as an int this will only get called if you pass in 0.

MelissaAutumn: Although I guess this makes sense 🤔

frontend/src/components/ScheduleCreation.vue
MelissaAutumn: Naw that's fine. Feel free to remove the comments, and just mentioned deep clone.

frontend/src/tbpro/elements/TextInput.vue
```diff
@@ -48,7 +48,7 @@ const props = withDefaults(defineProps<Props>(), {
   maxLength: null,
 });
 
-defineEmits(['submit']);
+const emit = defineEmits(['submit', 'blur']);
 defineExpose({ focus });
```
MelissaAutumn, line 52: I think you can expose blur
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/blur

then you can just grab the ref and call .blur. But the event also works.

MelissaAutumn (commented):
Thanks! Can you add some tests to make sure custom availability time doesn't break on timezones, and also does the switching weekdays from a list of strings to a list of ints break existing schedules? (like when we do the migration)

backend/src/appointment/database/repo/availability.py
devmount: This is how we do it in all repo get functions. I thought this is exactly to handle the 0 case, since ids always start at 1... Or maybe I don't understand the problem here 😂

MelissaAutumn: just thinking out loud, all good here!

frontend/src/tbpro/elements/TextInput.vue
MelissaAutumn: (I think the emit is fine)

rwood-moz (changes requested):
Very cool! I pulled the branch and ran the existing E2E tests on desktop Firefox and they pass.

One thing, a corner case. Looks like there's no verification on the end time. Example, when NOT using custom times per day if you attempt to set the availability end time to be earlier than the start time you get this error, as expected:

Screenshot 2025-05-20 at 11 18 54 AM

However if you do the same using `customize per day` it actually saves the setting with no error, and funny thing is, in this case, it creates availability slots on the share link from 9:00am on the Monday all day and all night until 8:00am the next day haha

Screenshot 2025-05-20 at 11 19 25 AM

Screenshot 2025-05-20 at 11 23 58 AM

devmount:
@rwood-moz I added validation now (backend and frontend too). Let me know if you can give it another try.

rwood-moz:
> @rwood-moz I added validation now (backend and frontend too). Let me know if you can give it another try.

Cool thanks for the update! It does show the error now for an end time earlier than the start time on custom days, however it is highlighting the wrong day/field i.e. the invalid time is on the Thursday but it highlights Wednesday:

Screenshot 2025-05-26 at 11 23 36 AM

rwood-moz (approved):
LGTM (and thanks for the validation fix, latest update worked great)!

MelissaAutumn (approved):

PR #1026, opened 2025-05-15
Build a nightly mobile web E2E tests job on BrowserStack

rwood-moz:
Create a new nightly production E2E test job that will run on mobile web on BrowserStack. Creating a separate job for mobile so that it can be manually triggered via GHA on it's own without having to run all of the E2E tests on the desktop browsers as well.

The new nightly E2E test web mobile job runs the settings tests (for now); an additional PR will be made to add the request a booking test.

Here's [a BrowserStack link with video](https://automate.browserstack.com/dashboard/v2/builds/bdf7c3062ad9424b633af548d24583ce46034df5/sessions/a27cd54f4c2a5b6bea3e819e255c59398ec93a66) running this PR / the settings tests on google pixel 5 chrome.

Also here's [a BrowserStack link](https://automate.browserstack.com/dashboard/v2/builds/f1f6d1f2e26545abc83fdb1c673355c2e3bb46e2) running this PR and all of the E2E tests on desktop firefox and chrome (just verifying the changes didn't break the desktop browser tests job).

PR #1028, opened 2025-05-16
Fix deployment and stage BrowserStack GHA jobs

rwood-moz:
My previous PR adding the nightly E2E mobile tests job broke the deployment jobs; here's the fix (goes with #1026).


devmount (approved):
Whoops, didn't see that either 😅

PR #1034, opened 2025-05-21
Fix CalDAV connector not supported relative urls, and CalDAV connector not supporting an empty response from component check (Fixes #1032) (Fixes #1033)

### LINKED ISSUE #1032: CalDAV doesn't support relative urls from .well-known.

### LINKED ISSUE #1033: Treat "supported-calendar-component-set" being empty as VEVENT is supported
Right now we check if VEVENT is returned within a `supported-calendar-component-set` call. However this is another optional feature of caldav. If it's empty we should expect every feature to be supported.

MelissaAutumn:
#1032 
#1033 

I immediately ran into these issues with Stalwart's new CalDAV support. Which is great, because it probably affects other CalDAV servers too.

### DIFF
--- a/backend/src/appointment/controller/calendar.py
+++ b/backend/src/appointment/controller/calendar.py
@@ -8,6 +8,7 @@ import logging
 import time
 import zoneinfo
 import os
+from functools import cache
 from urllib.parse import urlparse, urljoin
 
 import caldav.lib.error
@@ -361,7 +362,6 @@ class CalDavConnector(BaseConnector):
         perf_end = time.perf_counter_ns()
         print(f"CALDAV FreeBusy response: {(perf_end - perf_start) / 1000000000} seconds")
 
-
         items = []
 
         # This is sort of dumb, freebusy object isn't exposed in the icalendar instance except through a list of tuple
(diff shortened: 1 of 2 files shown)


backend/src/appointment/routes/caldav.py
```diff
@@ -86,21 +87,11 @@ def caldav_autodiscover_auth(
         lookup_url = Tools.well_known_caldav_lookup(connection.url)
 
     # If we have a lookup_url then apply it
-    if lookup_url:
+    if lookup_url and 'http' not in lookup_url:
+        connection.url = urllib.parse.urljoin(connection.url, lookup_url)
```
MelissaAutumn, line 91: Join the absolute and relative urls. 

Ex/ lookup_url = '/cal/dav/' and connection.url = 'https://example.org/' = 'https://example.org/cal/dav'

backend/src/appointment/controller/calendar.py
```diff
@@ -379,18 +379,21 @@ def get_busy_time(self, calendar_ids: list, start: str, end: str):
 
         return items
 
+    @staticmethod
+    @cache
+    def _is_supported(calendar):
+        """Is this calendar supported? Checks for VEVENT support, if an empty list is provided by the server that means"""
+        supported_comps = calendar.get_supported_components()
+        return len(supported_comps) == 0 or 'VEVENT' in supported_comps or None in supported_comps
```
MelissaAutumn, line 387: Since we do this twice, it's in a static method that's cached on property.

radishmouse (approved):

PR #1044, opened 2025-05-29
Have E2E tests use a single Appointment sign-in (and other improvements)

rwood-moz:
Currently each of the E2E tests sign into Appointment first and then perform the test. This PR updates the tests to use shared authentication. Before the tests start a setup task is ran which signs into Appointment, and saves the auth context. When the E2E tests run they are already signed into Appointment.

This makes the tests run much faster and be more stable. 🥳

The setup task also sets our default Appointment settings that the tests expect, so that doesn't have to be done inside each of the settings tests now either.

Note that I haven't been able to get the auth storage to work on mobile (talking with BrowserStack) and also since setting the timezone on android doesn't work (also talking to BrowserStack about this) I am scaling back the nightly E2E mobile test job to just test accessing the booking page.

This PR also splits up some of the larger tests into smaller files, so the individual run time and BrowserStack sessions are smaller.


rwood-moz:
Here's some BrowserStack links testing out this PR:

- Running the tests on [stage on desktop](https://automate.browserstack.com/dashboard/v2/builds/2bdf67e7aa645d8a625efd018f277d3a25142337) Firefox and Chrome
- Running the tests on [production on desktop](https://automate.browserstack.com/dashboard/v2/builds/4628f56cd417b8744ac42fc4a9f1a6e9b00dff16) Firefox and Chrome
- Running the scaled-back tests on [google pixel 5](https://automate.browserstack.com/dashboard/v2/builds/c6ceb6ba747b87cc6cbe80b3f0860b092725635e/sessions/015c47f3e56c39df5d08fa3d79e30381d81deef4) android

test/e2e/tests/access-booking-page.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+  bookingPage = new BookingPage(page);
+  dashboardPage = new DashboardPage(page);
+});
+
+// the share link (request a booking page) will display in the local browser context timezone but the main
+// appointment account settings could be a different timezone (if so the test will fail to find the booked
+// appointment since the time slot value will not match); set the browser context to always be in
+// `America/Toronto` so the share link will be in the same timezone as the main account settings
+test.use({
+  timezoneId: APPT_TIMEZONE_SETTING_TORONTO,
```
MelissaAutumn, line 66: We should maybe use a APPT_TIMEZONE_SETTING_PRIMARY

or something to indicate that this will be used for all tests.

test/e2e/tests/account-settings-other.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+    // note: we are already signed into Appointment with our default settings (via our auth-setup)
+    settingsPage = new SettingsPage(page);
+    dashboardPage = new DashboardPage(page);
+
+    // navigate to the account settings page
+    await settingsPage.gotoAccountSettingsPage();
+  });
+
+  test('invite codes available', async ({ page }) => {
+    await expect(settingsPage.inviteCodesHeader).toBeVisible({ timeout: TIMEOUT_30_SECONDS }); // generous time in case a large list
```
MelissaAutumn, line 31: User should only get 10.

test/e2e/tests/auth.setup.ts
```diff
@@ -0,0 +1,53 @@
...
+    APPT_MAIN_SETTINGS_PAGE,
+    TIMEOUT_2_SECONDS,
+} from "../const/constants";
+
+const fs = require('fs');
+const directoryPath = path.join(__dirname, '../test-results/.auth');
+
+fs.mkdir(directoryPath, (err: any) => {
+  if (err) {
+    console.error('error creating auth storage directory:', err);
```
MelissaAutumn, line 17: A nit: You can return after this line and get rid of the else to reduce indentations.

MelissaAutumn (approved):
Looks good! A few comments but no blockers.

PR #1053, opened 2025-06-05
caldav 1.4.0 -> 1.6.0

MelissaAutumn:
Changelog:
https://github.com/python-caldav/caldav/blob/master/CHANGELOG.md#160---2025-05-30

Seems okay on my end, I upgraded to test something but didn't run into any issues. So should be okay to push up.


MelissaAutumn:
...lint fix is on https://github.com/thunderbird/appointment/pull/1052. Ignore that..

PR #1058, opened 2025-06-09
Add a data migration to fill in short_link_hash if empty. (Fixes #1057)

### LINKED ISSUE #1057: short_link_hash may not exist for some users
Sentry ticket:
https://thunderbird.sentry.io/issues/6668638865/

MelissaAutumn:
Fixes #1057 

Short_link_hash was added in a later migration so some early users (like me actually!) have this set to None. This should have been caught earlier, but it's environment-specific.


devmount (commented):
Looking good! Thanks for fixing that!

PR #1065, opened 2025-06-20
Multiple fine-tune availability entries per day

devmount:
## Description of the Change

This change introduces multiple availability entries per day of week. It adds action buttons to add, copy and remove custom availabilities.

image

Validation happens on frontend (with instant feedback for users) and backend (for safety and data consistence). Validation includes checking if start and end time of an entry are reasonable and have at least a distance of the slot duration. It is also checked that times of adjacent entries are valid.

image

A copy button can be used to copy all entries from the selected day to all other days that are checked. Existing availability on those days gets replaced.

image

A delete button can be used to remove an entry. If the last entry gets removed, the day is disabled.

## Benefits

More fine-grained control to set up schedule availability.

## Applicable Issues

Related to #422


backend/src/appointment/controller/calendar.py
```diff
@@ -725,55 +725,61 @@ def available_slots_from_schedule(schedule: models.Schedule) -> list[schemas.Slo
...
-                hour=start_local.hour,
-                minute=start_local.minute,
-                tzinfo=timezone,
-            )
+            customAvailabilities = [x for x in availabilities if date.isoweekday() == x.day_of_week.value]
+            if custom_times and len(customAvailabilities) > 0:
+                parts = [(x.start_time_local, x.end_time_local) for x in customAvailabilities]
+
+            # Now we cut every part of availability time to available time slots
+            for (start_local, end_local) in parts:
```
devmount, line 738: The big change here is, that we now iterate over multiple slots instead of just assuming we have one time slot per day.

backend/src/appointment/database/repo/schedule.py
```diff
@@ -87,6 +87,27 @@ def has_slot(db: Session, schedule_id: int, slot_id: int):
...
+            if not utils.is_valid_time_range(a.start_time, a.end_time, schedule.slot_duration):
+                return False
+            # Chack valid predecessor
+            if (i > 0 and a.day_of_week == schedule.availabilities[i-1].day_of_week
+                and not utils.is_valid_time_range(schedule.availabilities[i-1].end_time, a.start_time)):
+                return False
+            # Check valid successor
+            if (i < size-1 and a.day_of_week == schedule.availabilities[i+1].day_of_week
+                and not utils.is_valid_time_range(a.end_time, schedule.availabilities[i+1].start_time)):
+                return False
```
devmount: Let me know if you think this is redundant. It felt right to check both sides when implementing this 😇

frontend/package-lock.json
devmount, line 1: I updated the @vueuse dependencies.

frontend/src/components/CalendarQalendar.vue
```diff
@@ -394,6 +394,7 @@ watch(route, () => {
       @updated-period="dateChange"
       @updated-mode="modeChange"
       ref="qalendarRef"
+      class="h-auto"
```
devmount, line 397: This makes sure the calendar view on the left doesn't scale-y when the schedule configuration column on the right gets higher (due to addition of custom availabilities e.g.).

frontend/src/models.ts
```diff
@@ -428,9 +428,9 @@ export type Time<T> = {
 export type TimeNumeric = Time<number>;
 export type TimeFormatted = Time<string>;
 
-export type SelectOption = {
+export type SelectOption<T = number> = {
   label: string;
-  value: string | number;
+  value: T;
```
devmount, line 433: FYI: I refactored the SelectOption to process a given type instead of switching between string and number. That solved a lot of conversion issues.

frontend/src/tbpro/elements/CheckboxInput.vue
```diff
@@ -150,5 +150,14 @@ const onChange = () => {
...
   }
+
+  &:disabled {
+    cursor: not-allowed;
+    opacity: .5;
+
+    & ~ span {
+      color: var(--colour-ti-muted);
+    }
+  }
```
devmount, line 164: This is now part of one of @davinotdavid's PRs too.

frontend/src/views/ScheduleView.vue
```diff
@@ -127,7 +127,7 @@ const dismiss = () => {
   <!-- page content -->
   <div class="mt-8 flex flex-col-reverse items-stretch gap-2 md:flex-row-reverse lg:gap-4">
     <!-- schedule creation dialog -->
-    <div class="z-10 mx-auto mb-10 w-3/4 min-w-80 sm:w-1/4 md:mb-0 xl:w-1/6">
+    <div class="z-10 mx-auto mb-10 w-3/4 min-w-96 sm:w-1/4 md:mb-0 xl:w-1/6">
```
devmount, line 130: I took the liberty to increase the width of the schedule configuration sidebar 😇 This gives more space for all the buttons for custom availability.

devmount (commented):
Here are some pointers. Thanks for looking over the changes!

backend/src/appointment/database/repo/schedule.py
rwood-moz: So if any one of the `schedule.availabilities` is not valid then this will return False from the function even if there might be another availability that is valid (just so I'm understanding, is that correct)?

Also `if size is 0` (there are no schedule.availabilties) this function will return `True` even though there are no availabilities.

rwood-moz (commented):
This looks really cool!

Suggestion: Add some integration tests for setting multiple availabilities, maybe similar to:

https://github.com/thunderbird/appointment/blob/main/backend/test/integration/test_schedule.py#L244

frontend/src/elements/AvailabilityCopyDropdown.vue
```diff
@@ -0,0 +1,81 @@
...
+        <checkbox-input
+          v-for="d in isoWeekdays"
+          v-model="selectedDays[d.iso]"
+          :key="d.iso"
+          :value="d.iso"
+          :name="d.short.toLowerCase()"
+          :label="d.long"
+          :disabled="disabledWeekdays.includes(d.iso)"
+        />
+        <hr>
```
rwood-moz, line 70: If you could please add some `data-testid` attributes that would be useful for future E2E tests :)

backend/src/appointment/database/repo/schedule.py
devmount: > So if any one of the `schedule.availabilities` is not valid then this will return False from the function even if there might be another availability that is valid (just so I'm understanding, is that correct)?

That is correct. Thanks for pointing out, that this was not obvious! Should I change the function name to `all_availability_is_valid` to make this more clear?

> Also `if size is 0` (there are no schedule.availabilties) this function will return `True` even though there are no availabilities.

Technically this case won't ever happen, since we always have at least one entry on one day of week when custom availability is activated. If you think it would be better to switch the conditions and move this to a `invalid_availability_exists`, I can happily do that.

frontend/src/elements/AvailabilityCopyDropdown.vue
devmount: Of course, sorry for always missing that 😅

devmount: This would be something dynamic like `dashboard-availability-copy-<isoweekday>-input`?

backend/src/appointment/database/repo/schedule.py
rwood-moz: Ok thanks for clarifying, no I think renaming to `all_availability_is_valid` would be good

frontend/src/elements/AvailabilityCopyDropdown.vue
rwood-moz: No worries, yes something like that would be awesome, thank you :) also maybe if you could add some in the `frontend/src/elements/AvailabilitySelect.vue` too if you have a chance.

rwood-moz:
One thing I noticed, when I change the availability and save, my calendar on the dashboard is not updated and it still shows the old availability:

Screenshot 2025-06-24 at 11 15 03 AM

devmount:
@rwood-moz 

> One thing I noticed, when I change the availability and save, my calendar on the dashboard is not updated and it still shows the old availability:

Yes, I didn't touch the calendar view, since (I thought) we're about to drop that calendar view... still a good point we should discuss.

frontend/src/elements/AvailabilityCopyDropdown.vue
devmount: I added them. Let me know if you need more anywhere else!

backend/src/appointment/database/repo/schedule.py
devmount: Alright, renamed now!

devmount:
> This looks really cool!
> 
> Suggestion: Add some integration tests for setting multiple availabilities, maybe similar to:
> 
> https://github.com/thunderbird/appointment/blob/main/backend/test/integration/test_schedule.py#L244

@rwood-moz  I already did that in the previous iteration, but you're right - I'll add more with multiple availability entries per day 👍🏻  Thanks for catching that!

devmount:
@rwood-moz Two more integration tests added now, if you want to take a look again 😇  For the dashboard calendar issue I opened a discussion here https://github.com/thunderbird/appointment/issues/422#issuecomment-3001347614

rwood-moz:
> @rwood-moz Two more integration tests added now, if you want to take a look again 😇 For the dashboard calendar issue I opened a discussion here [#422 (comment)](https://github.com/thunderbird/appointment/issues/422#issuecomment-3001347614)

Awesome, looks great, thank you @devmount

rwood-moz (approved):
LGTM

backend/src/appointment/database/repo/schedule.py
```diff
@@ -87,6 +87,27 @@ def has_slot(db: Session, schedule_id: int, slot_id: int):
     return db_slot and db_slot.schedule_id == schedule_id
 
 
+def all_availability_is_valid(schedule: schemas.ScheduleValidationIn):
+    """check if availability is valid"""
+    size = len(schedule.availabilities)
+    if size > 0:
```
MelissaAutumn: You can flip to be 
```suggestion
    if size <= 0:
        return True
```
and then remove a bunch of indentation.

backend/src/appointment/database/repo/schedule.py
```diff
@@ -87,6 +87,27 @@ def has_slot(db: Session, schedule_id: int, slot_id: int):
...
+        # Make sure, that the availabilities are sorted by weekday AND by start time. This is important for checking
+        # validity of times of adjacent availabilities at the same day
+        for i, a in enumerate(sorted(schedule.availabilities, key = lambda x: (x.day_of_week.value, x.start_time))):
+            # Check valid times and duration of a single availability entry
+            if not utils.is_valid_time_range(a.start_time, a.end_time, schedule.slot_duration):
+                return False
+            # Chack valid predecessor
```
MelissaAutumn:
```suggestion
            # Check valid predecessor
```

backend/src/appointment/database/repo/schedule.py
```diff
@@ -87,6 +87,27 @@ def has_slot(db: Session, schedule_id: int, slot_id: int):
...
+            if (i > 0 and a.day_of_week == schedule.availabilities[i-1].day_of_week
```
MelissaAutumn: I'm still a little sick today, but is this checking the end time to make sure it doesn't overlap with the prev/next day's first start time?

backend/test/integration/test_schedule.py
```diff
@@ -292,6 +293,137 @@ def test_update_schedule_availabilities(self, with_client, make_schedule):
         assert len(data['availabilities']) == 3
         assert data['availabilities'][2]['day_of_week'] == 6
 
+    def test_update_multiple_valid_schedule_availabilities_per_weekday(self, with_client, make_schedule):
```
MelissaAutumn, line 296: Can you add a test for the invalid adjacent day/slot check?

frontend/src/elements/AvailabilitySelect.vue
```diff
@@ -143,35 +225,56 @@ const toggleBubble = (option: SelectOption) => {
...
+              :name="`end_time_${option.value}_${i}`"
+              v-model="availability.end_time"
+              :error="validationErrors[option.value][i]"
+              :disabled="disabled"
+              :small-input="true"
+              :data-testid="`dashboard-availability-end-time-${option.value}-${i}-input`"
+              @change="update()"
+            />
+            <span>
+              <link-button v-if="i === 0" size="small" class="action-btn" @click="addAvailability(option)">
```
MelissaAutumn: Can you add a title attribute, screen reader is just saying "Button" 😄

frontend/src/elements/AvailabilitySelect.vue
```diff
@@ -143,35 +225,56 @@ const toggleBubble = (option: SelectOption) => {
...
+            </span>
+            <span>
+              <availability-copy-dropdown
+                v-if="i === 0"
+                :disabled-weekdays="[option.value, ...disabledWeekdays]"
+                @copy="event => copyAvailability(option, event)"
+              />
+            </span>
+            <span>
+              <link-button size="small" class="action-btn action-remove" @click="removeAvailability(option, i)">
```
MelissaAutumn: ditto here as it just says "Button" 😅

frontend/src/elements/AvailabilityCopyDropdown.vue
```diff
@@ -0,0 +1,82 @@
...
+<template>
+  <drop-down class="self-center" ref="copyDropdown">
+    <template #trigger>
+      <link-button size="small" class="action-btn action-copy">
+        <icon-asterisk class="w-4" aria-hidden="true"/>
+      </link-button>
+    </template>
+    <template #default>
+      <container-box class="flex flex-col gap-2">
+        Copy to:
```
MelissaAutumn: l10n?

MelissaAutumn (changes requested):
Thanks for grabbing this one! Availability math always messes with me even when I'm not sick, so just some questions / feedback.

backend/src/appointment/database/repo/schedule.py
devmount: Good idea!

backend/src/appointment/database/repo/schedule.py
devmount: Whoopsi

backend/src/appointment/database/repo/schedule.py
devmount: That should be covered, yes. With the current implementation slot configuration cannot go over midnight. Start and end need to be on the same day of week, otherwise the first check will fail.

devmount: I improved the comments a bit, I hope it's more clear now.

backend/test/integration/test_schedule.py
devmount: Already there, starting at l348 😇  It tests invalid adjacent time slots, invalid start / end times and I added another test for a slot duration that is too small.

frontend/src/elements/AvailabilitySelect.vue
devmount: Of course, added!

frontend/src/elements/AvailabilitySelect.vue
devmount: Good catch!

frontend/src/elements/AvailabilityCopyDropdown.vue
devmount: Haha this happens if I forget to test things in German at the end 😂  Thanks for pointing that one out, I added titles, labels and everything to that component now!

MelissaAutumn (approved):

Issue #1061, opened 2025-06-11
Optionally send a response when declining a booking request


ryanjjung:
**Is your feature request related to a problem? Please describe.**

When I decline a booking request, the system automatically sends a message to the booker saying I have done so. But it's my responsibility now to email the booker independently to explain why.

**Describe the solution you'd like**

When I click Deny, I should be prompted to optionally enter a message which will be delivered to the booker if I so choose.

**Describe alternatives you've considered**

I grabbed the booker's email from the booking request before declining it and emailed them independently.

devmount:
Thank you Ryan. This is related to #171, where we discussed providing a reason when cancelling/declining a booking request.

devmount:
@laurelterlesky @solangevalverde I think this needs some UI exploration, although we could implement a first version with just a textinput and a submit button to trigger the booking decline, if you want.

davinotdavid:
@ryanjjung I've incorporated the idea of this ticket into the cancel confirmed appointment feature (https://github.com/thunderbird/appointment/pull/1068) but the reason can only be added _after_ the appointment has been accepted / confirmed first (in other words, during cancellation and not deny).

Would this meet the expectation of this ticket or would you also want the option to add a reason on deny (in other words, before it was even accepted in the first place)?

laurelterlesky:
Adjustments to the confirmed booking panel to include an option to cancel with a response:
https://zpl.io/p0x5Woj & https://zpl.io/30xmGvx

ryanjjung:
Yeah, the idea is that I may decline for various reasons. Maybe it's a bad time/date. Maybe I have decided I do not have any business with this person at all. "They declined your booking request" doesn't convey anything more than that and leaves me having to send them an email to clarify. I think that defeats the purpose of using this tool (I should *not* have to send an email to book an appointment), and I think it's useful to be able to send a clarifying message at the time that I decline a booking request. I probably just need to tell the person to try another time.

Issue #171, opened 2023-11-14
Cancel a pending/booked appointment


MelissaAutumn:
You should be able to cancel a pending or booked appointment. This includes any appointments scheduled via general availability. 

This should cascade into removing the event from the users external calendar, and cancelling any zoom meetings attached to the appointment. 

This should also send an email to the attendees with a optional reason why the event was canceled.

devmount:
For reference: Rejecting a pending booking request, event removal and email notification for attendees is now possible with #768.

@MelissaAutumn I'm wondering about the difference of cancelling and rejecting a pending event. Despite giving an optional reason (which we could implement for rejecting a booking request too), those are somehow the same, or aren't they?

In any case, cancelling a confirmed booking still needs to be done.

malini:
The zoom app publishing admin expects us to be able to cancel meetings prior to getting our app published (Waiting on confirmation, but it seems to be the case). So we should add 'Cancel booking' button like so:

![Image](https://github.com/user-attachments/assets/f8e0b218-e1c2-4eb7-865c-884a4306319b)

@Jesse-Moz @solangevalverde ^ tagging to let you know of a new button/feature coming to the bookings page!

MelissaAutumn:
We'll probably also want to cancel any zoom meeting links on the appointment. I think this api endpoint should work: https://developers.zoom.us/docs/api/meetings/#tag/meetings/DELETE/meetings/{meetingId}

Jesse-Moz:
New designs have buttons for all the following:

HOST: Cancel Appointment (on the Edit view for an individual Booking)
HOST: Decline Appointment (on the info slide-out for a Booking in the Pending state)
BOOKER: Cancel Appointment (click-through from the confirmation email to a Booking Details page)

Hopefully this covers all the functionality -- let me know if we're missing any major flows!

davinotdavid:
@MelissaAutumn Would the "Cancel confirmed appointment feature" PR (https://github.com/thunderbird/appointment/pull/1068) solve this issue or do you see other parts that still need working?

MelissaAutumn:
I believe so

PR #1068, opened 2025-06-20
[WIP] Cancel confirmed appointment feature

davinotdavid:
## Description of the Change

[**Main work**]
- In the frontend, added a textarea for an optional reason and a `Cancel booking` button in the `AppointmentModal` component
- In the backend, added a new endpoint to cancel an appointment (`POST` to `/apmt/{id}/cancel`) that looks up all the slots associated with that appointment and:
   1. Sends a cancellation email with and optional reason (new email template added)
   2. If the `meeting_link_provider` is Zoom, calls the Zoom client to delete the meeting

[**Small refactor**]
- We were using `reject` and `cancel` a little bit interchangeably which creates some confusion (at least for me) in terms of _when_ that action took place. For this PR, I have separated the concepts so that `reject` means that an appointment that is pending approval has been rejected (as in, never accepted) and `cancelled` is cancelling an appointment that has already been confirmed.

[**Fix**]
- When a schedule is created and the settings for creating a zoom link is enabled, it gets set in the `meeting_link_provider` of the schedule itself. However, the appointment doesn't seem to ever get its `meeting_link_provider` updated so upon appointment creation, I've "synced" their fields so that I can later read it from appointment cancellation.

## Benefits

- New cancellation of confirmed booking feature that also deletes Zoom meetings if any
- Separation of cancel vs. reject concepts

## Applicable Issues

https://github.com/thunderbird/appointment/issues/171
https://github.com/thunderbird/appointment/issues/1061

### DIFF
--- a/backend/src/appointment/routes/api.py
+++ b/backend/src/appointment/routes/api.py
@@ -18,9 +18,10 @@ from ..controller.calendar import CalDavConnector, Tools, GoogleConnector
 from fastapi import APIRouter, Depends, HTTPException, BackgroundTasks, Request
 from ..controller.apis.google_client import GoogleClient
 from ..controller.auth import signed_url_by_subscriber, schedule_slugs_by_subscriber, user_links_by_subscriber
-from ..database.models import Subscriber, CalendarProvider, InviteStatus
+from ..database.models import Subscriber, CalendarProvider, InviteStatus, MeetingLinkProviderType
 from ..defines import DEFAULT_CALENDAR_COLOUR
 from ..dependencies.google import get_google_client
+from ..dependencies.zoom import get_zoom_client
 from ..dependencies.auth import get_subscriber
 from ..dependencies.database import get_db, get_redis
 from ..exceptions import validation
@@ -454,6 +455,39 @@ def delete_my_appointment(id: int, db: Session = Depends(get_db), subscriber: Su
 
     repo.appointment.delete(db=db, appointment_id=id)
     return True
+
+
+@router.post('/apmt/{id}/cancel')
+def cancel_my_appointment(id: int, background_tasks: BackgroundTasks, payload: schemas.AppointmentCancelRequest, db: Session = Depends(get_db), subscriber: Subscriber = Depends(get_subscriber)):
+    """endpoint to cancel an appointment from db"""
+    if not repo.appointment.exists(db, appointment_id=id):
+        raise validation.AppointmentNotFoundException()
+    if not repo.appointment.is_owned(db, appointment_id=id, subscriber_id=subscriber.id):
+        raise validation.AppointmentNotAuthorizedException()
+
+    appointment = repo.appointment.get(db, id)
+    zoom_client = None
+
+    # If needed, get a hold of the Zoom client
+    if appointment.meeting_link_provider == MeetingLinkProviderType.zoom:
+        try:
+            zoom_client = get_zoom_client(subscriber)
+        except Exception as ex:
+            sentry_sdk.capture_exception(ex)
+
+    # Send cancel information to all slots' bookees
+    for slot in appointment.slots:
+        Tools().send_cancel_vevent(background_tasks, appointment, slot, subscriber, slot.attendee, payload.reason)
+
+        # If needed, delete appointment's Zoom meeting through Zoom client
+        if zoom_client:
+            try:
+                zoom_client.delete_meeting(slot.meeting_link_id)
+            except Exception as ex:
+                sentry_sdk.capture_exception(ex)
+
+        # Delete the appointment, this will also delete the slot
+        repo.appointment.delete(db, slot.appointment_id)
 
 
 @router.post('/support')
--- a/backend/src/appointment/routes/api.py
+++ b/backend/src/appointment/routes/api.py
@@ -445,6 +445,7 @@ def public_appointment_serve_ics(slug: str, slot_id: int, db: Session = Depends(
         data=Tools().create_vevent(appointment=db_appointment, slot=slot, organizer=organizer).decode('utf-8'),
     )
 
+
 @router.delete('/apmt/{id}')
 def delete_my_appointment(id: int, db: Session = Depends(get_db), subscriber: Subscriber = Depends(get_subscriber)):
     """endpoint to remove a appointment from db"""
@@ -458,7 +459,13 @@ def delete_my_appointment(id: int, db: Session = Depends(get_db), subscriber: Su
 
 
 @router.post('/apmt/{id}/cancel')
-def cancel_my_appointment(id: int, background_tasks: BackgroundTasks, payload: schemas.AppointmentCancelRequest, db: Session = Depends(get_db), subscriber: Subscriber = Depends(get_subscriber)):
+def cancel_my_appointment(
+    id: int,
+    background_tasks: BackgroundTasks,
+    payload: schemas.AppointmentCancelRequest,
+    db: Session = Depends(get_db),
+    subscriber: Subscriber = Depends(get_subscriber),
+):
     """endpoint to cancel an appointment from db"""
     if not repo.appointment.exists(db, appointment_id=id):
         raise validation.AppointmentNotFoundException()
--- a/backend/src/appointment/templates/email/cancelled.jinja2
+++ b/backend/src/appointment/templates/email/cancelled.jinja2
@@ -1,0 +1,9 @@
+{% extends "includes/base.jinja2" %}
+{# Code begins! #}
+{% block introduction %}
+  <div style="color: {{ colour_text_base }}; text-align: center; line-height: 16px;">
+    <p>
+      {{ l10n('cancel-mail-html-heading', {'owner_name': owner_name, 'date': date, 'reason': reason}) }}
+    </p>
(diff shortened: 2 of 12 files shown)


backend/test/integration/test_appointment.py
devmount, line 1: Great stuff, thanks for providing those tests!

frontend/src/components/AppointmentModal.vue
```diff
@@ -38,6 +38,8 @@ interface Props {
 }
 const props = defineProps<Props>();
 
+const cancelReason = ref<string>("");
```
devmount: Just a nit: We're using single quotes for strings in JS (and double quotes in HTML/Markup)

```suggestion
const cancelReason = ref<string>('');
```

frontend/src/components/AppointmentModal.vue
```diff
@@ -62,6 +64,13 @@ const deleteAppointment = () => {
   apmtStore.deleteAppointment(props.appointment?.id);
   emit('close');
 };
+
+// Handle cancel
+const cancelAppointment = () => {
+  apmtStore.cancelAppointment(props.appointment?.id, cancelReason.value);
+  cancelReason.value = "";
```
devmount: See comment above

```suggestion
  cancelReason.value = '';
```

frontend/src/stores/appointment-store.ts
devmount, line 1: Can you provide a test for the new cancelAppointment function too? We already have a appointment-stor.test.js that can be easily extended.  That'd be great!

backend/src/appointment/routes/api.py
```diff
@@ -456,6 +458,45 @@ def delete_my_appointment(id: int, db: Session = Depends(get_db), subscriber: Su
     return True
 
 
+@router.post('/apmt/{id}/cancel')
+def cancel_my_appointment(
```
MelissaAutumn, line 464: Like @devmount mentioned we'll need to call calendar.delete_event, we don't have a good one-line way to initialize that right now, so feel free to copy and paste from another function.

backend/test/integration/test_appointment.py
```diff
@@ -124,3 +129,92 @@ def test_dont_show_other_subscribers_appointments(
...
+        appointment = make_appointment()
+        payload = {'reason': 'Test cancellation'}
+
+        with patch('appointment.routes.api.Tools') as mock_tools:
+            response = with_client.post(f'/apmt/{appointment.id}/cancel', headers=auth_headers, json=payload)
+
+            assert response.status_code == 200
+            assert mock_tools().send_cancel_vevent.call_count == len(appointment.slots)
+
+        db = with_db()
```
MelissaAutumn: You can do `with with_db() as db:` to avoid having to call .close(). I do recognize that's maybe one with too many though 😅

frontend/src/components/AppointmentModal.vue
```diff
@@ -176,8 +185,23 @@ const deleteAppointment = () => {
         </div>
         <div class="rounded-lg border border-gray-400 p-4 dark:border-gray-600">{{ appointment.details }}</div>
       </div>
-      <div class="p-6" v-if="status === BookingStatus.Booked">
-        <p>This booking is confirmed.</p>
+      <div class="px-4" v-if="status === BookingStatus.Booked">
+        <p class="mb-8">This booking is confirmed.</p>
+        <form class="gap-4" @submit.prevent>
+          <label for="cancelReason">
+            {{ t("label.cancelReason") }}
```
MelissaAutumn:
```suggestion
            {{ t('label.cancelReason') }}
```

MelissaAutumn (changes requested):

frontend/src/components/AppointmentModal.vue
davinotdavid: Thank you! Updated!

backend/test/integration/test_appointment.py
davinotdavid: Oh right! Good catch, thanks! Updated!

backend/src/appointment/routes/api.py
davinotdavid: Ah that makes sense! I am not sure architecture-wise where it would be a good location to extract that logic so I followed the advice to copy from another function. Thanks, updated in the code!

frontend/src/stores/appointment-store.ts
davinotdavid: For sure, yes! I've added one just to assert that the call is being done to the correct endpoint with the expected body. Is it similar with what you were thinking for this test?

frontend/src/components/AppointmentModal.vue
davinotdavid: Thanks! Updated!

frontend/src/components/AppointmentModal.vue
davinotdavid: Oh thank you! I will keep that in mind for future PRs, good catch! 🙏

backend/test/integration/test_appointment.py
davinotdavid: Thank you so much! I really appreciate your reviews 🙏

davinotdavid:
@MelissaAutumn @devmount Thank you so much for the review / insights! I really appreciate it 🙇

I've updated the PR description with the new commits but the diff / changes are:
- It is now getting the remote calendar connection and calling `delete_event`
- Added a check on the frontend for `isPast` to not render the cancellation form if the appointment is in the past

backend/src/appointment/templates/email/cancelled.jinja2
```diff
@@ -0,0 +1,9 @@
+{% extends "includes/base.jinja2" %}
+{# Code begins! #}
+{% block introduction %}
+  <div style="color: {{ colour_text_base }}; text-align: center; line-height: 16px;">
+    <p>
+      {{ l10n('cancel-mail-html-heading', {'owner_name': owner_name, 'date': date, 'reason_line': reason_line}) }}
```
rwood-moz, line 6: The `cancel-mail-html-heading` already includes the `reason_line` - do we want the reason line appearing twice?

frontend/src/components/AppointmentModal.vue
```diff
@@ -176,8 +186,23 @@ const deleteAppointment = () => {
...
+        <p class="mb-8">This booking is confirmed.</p>
+        <form v-if="!isPast" class="gap-4" @submit.prevent>
+          <label for="cancelReason">
+            {{ t('label.cancelReason') }}
+            <textarea
+                name="cancelReason"
+                v-model="cancelReason"
+                :placeholder="t('placeholder.writeHere')"
+                class="w-full h-24 rounded-md resize-none mt-2 mb-8"
+                data-testid="appointment-modal-reason-input"
```
rwood-moz: Thanks for adding! Suggest data-testid=`appointment-modal-cancel-reason-input` (add cancel)

backend/src/appointment/templates/email/cancelled.jinja2
rwood-moz: Oh wait, nvm, that builds the heading with those params hah

frontend/src/components/AppointmentModal.vue
davinotdavid: Awesome! Thanks for the suggestion, just changed here!

backend/src/appointment/templates/email/cancelled.jinja2
davinotdavid: Good point! It might be a bit unclear but the intention was to build the `reason_line` as `Reason: { $reason }` but I figured that `Reason` should also be in i18n so I resorted for calling it `reason_line` and pass it to the templates instead 😅

frontend/src/stores/appointment-store.ts
devmount: Awesome, thanks!

rwood-moz:
> Thanks for the quick implementation of the review feedback! Looks good on my end now! @MelissaAutumn @rwood-moz ?

LGTM! Just one question (that doesn't need to be addressed in this PR) - after an appointment is cancelled there is no history of it. Do we want cancelled bookings to appear on the `Booking` tab => `Past` appointment list, and show a status of cancelled? What if a user cancels the wrong appointment but they aren't able to confirm as it just disappears from the bookings history? Not sure if that should be considered.

davinotdavid:
@devmount @rwood-moz thank you so much for the re-reviews! I really appreciate it!

Oh that's a very good point! IMHO I agree it would be nice to be able to see the previously cancelled appointments, even if just for confirming that it was actually cancelled. Perhaps even a dedicated "Trash" / "Deleted" section to hold deleted appointments (regardless of their status) in the future?

MelissaAutumn:
feel free to punt that to a new ticket, are appointments soft deleted or is that just users?

MelissaAutumn (approved):

davinotdavid:
@MelissaAutumn cool, created a [new ticket](https://github.com/thunderbird/appointment/issues/1073) to track that! Thanks for the review!

Appointments seem to be hard-deleted so there would be a need for updating the logic around that to soft-delete them instead and maybe even have a job to hard-delete after X amount of days would be nice.

Issue #1008, opened 2025-04-25
Generate zoom meeting link should turn on after connecting a zoom account


MelissaAutumn:
There seems to be an issue where if someone disconnects and reconnects a zoom account it doesn't activate the setting on the schedule. 

This is confusing behaviour and we should make sure it enables on connection.

davinotdavid:
Closing as work has been merged but feel free to re-open if needed!

PR #1069, opened 2025-06-21
Update schedules' meeting_link_provider upon Zoom connection

davinotdavid:
## Description of the Change

Upon Zoom disconnection, we are resetting the schedule's `meeting_link_provider` but we were not setting it back upon re connection, so now whenever Zoom is successfully connected, upon its callback, set the subscriber's schedules' `meeting_link_provider` to be `zoom`.

## Benefits

Prevents confusion / extra steps on connecting Zoom and finding out that the meeting link generation setting is off.

## Applicable Issues

https://github.com/thunderbird/appointment/issues/1008

### DIFF
--- a/backend/src/appointment/controller/zoom.py
+++ b/backend/src/appointment/controller/zoom.py
@@ -4,6 +4,16 @@ from appointment.database import repo, models
 from appointment.database.models import ExternalConnectionType
 
 
+def update_schedules_meeting_link_provider(db: Session, subscriber_id: int) -> bool:
+    """Updates existing schedules' meeling link provider to be zoom"""
+    schedules = repo.schedule.get_by_subscriber(db, subscriber_id)
+    for schedule in schedules:
+        schedule.meeting_link_provider = models.MeetingLinkProviderType.zoom
+        db.add(schedule)
+    db.commit()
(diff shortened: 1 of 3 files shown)


backend/src/appointment/controller/zoom.py
```diff
@@ -4,6 +4,16 @@
 from appointment.database.models import ExternalConnectionType
 
 
+def update_schedules_meeting_link_provider(db: Session, subscriber_id: int) -> bool:
+    """Updates existing schedules' meeling link provider to be zoom"""
```
devmount: Typo

```suggestion
    """Updates existing schedules' meeting link provider to be zoom"""
```

devmount (approved):
Just a typo found, looking good! Thanks for providing a test 💪🏻

backend/src/appointment/controller/zoom.py
davinotdavid: OMG 🤦  hahaha great catch! Thanks for the review and suggested changes!

MelissaAutumn (approved):
This will need some additional thought in the future when we have multiple schedules active at once. But for now it looks good!

Thanks!

Issue #973, opened 2025-04-01
Give the user some feedback if they try to generate Zoom links but haven't connected their Zoom account


malini:
**Is your feature request related to a problem? Please describe.**
If you do not have a Zoom account connected, then in the GA settings config, if you click the checkbox to generate Zoom meetings, nothing happens:

![Image](https://github.com/user-attachments/assets/5f0d7804-89b8-480f-bd1d-f753c6b96d6a)

**Describe the solution you'd like**
There should be some indication to the user that this is not set up, so they can't generate a zoom link. Something along the lines of:

disabling the "Generate Zoom Meeting"  checkbox if no Zoom account is hooked up yet, and show the following text: "Zoom is not connected yet. Please [connect a Zoom account](https://appointment.day/settings/connectedAccounts) to use this feature.", with the link taking them to https://appointment.day/settings/connectedAccounts so they can set up an account

davinotdavid:
Closing this issue as work has been checked on Staging but feel free to re-open if needed!

Issue #296, opened 2024-03-11
Determine our l10n workflow


malini:
**Is your feature request related to a problem? Please describe.**
We should support a multitude of languages, but we currently only have EN and DE

**Describe the solution you'd like**
- Get l10n strings for all supported languages
- Determine how we will maintain l10n strings moving forward

malini:
@Sancus is taking this on and as an update, we briefly discussed https://www.npmjs.com/package/vue-i18n as our i18n framework on the frontend side of things.

PR #1071, opened 2025-06-23
Add explainer to "generate Zoom meeting checkbox" when Zoom is not connected

davinotdavid:
## Description of the Change

In case Zoom is not enabled, there was no message to the subscriber explaining why they can't click on the checkbox to generate zoom meeting. This adds a message with a link to the related setting (`/settings/connectedAccounts`) and changes the cursor to `not-allowed` when checkboxes are disabled.

Jun-23-2025 14-37-33

## Benefits

- Better feedback to the subscriber on the available actions around Zoom's settings to generate meeting link.
- Cursor update when checkbox is disabled for all usages of the `CheckboxInput` component

## Applicable Issues

https://github.com/thunderbird/appointment/issues/973

### DIFF
--- a/frontend/src/components/ScheduleCreation.vue
+++ b/frontend/src/components/ScheduleCreation.vue
@@ -389,6 +389,9 @@ const copyLink = async () => {
   }, 4000);
 };
 
+// controls the generate zoom link checkbox
+const isZoomEnabled = computed(() => scheduleInput.value.active && hasZoomAccount.value);
+
 // track schedule activation toggle changes
 watch(
   () => scheduleInput.value.active,
@@ -688,9 +691,16 @@ watch(
             name="generateZoomLink"
             :label="t('label.generateZoomLink')"
             v-model="generateZoomLink"
-            :disabled="!scheduleInput.active || !hasZoomAccount"
+            :disabled="!isZoomEnabled"
             @change="toggleZoomLinkCreation"
           />
+          <i18n-t v-if="!isZoomEnabled" keypath="text.generateZoomMeetingHelpDisabled.text" tag="span" scope="global" class="text-xs">
+            <template v-slot:link>
+              <router-link class="underline" to="settings/connectedAccounts">
+                {{ t('text.generateZoomMeetingHelpDisabled.link') }}
+              </router-link>
+            </template>
+          </i18n-t>
           <label class="relative flex flex-col gap-1">
             <div class="input-label ">
               {{ t("label.notes") }}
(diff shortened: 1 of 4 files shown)


frontend/src/components/ScheduleCreation.vue
```diff
@@ -389,6 +389,9 @@ const copyLink = async () => {
   }, 4000);
 };
 
+// controls the generate zoom link checkbox
+const isZoomEnabled = computed(() => scheduleInput.value.active && hasZoomAccount.value);
```
devmount, line 393: Thanks for throwing that into a computed prop 👍🏻

frontend/src/locales/de.json
```diff
@@ -461,6 +461,10 @@
     "defineDaysAndTimeSlots": "Definiere Tage und Zeitfenster, damit Teilnehmer die Zeit des Events auswählen können …",
     "denialSentToAddress": "Die Informationen über die Ablehnung der Anfrage wurde per E-Mail an {address} geschickt.",
     "disclaimerGABooking": "Bitte beachten: Der Eigentümer muss die Buchung bestätigen.",
+    "generateZoomMeetingHelpDisabled": {
+      "link": "ein Zoom-Konto verbinden",
+      "text": "Zoom ist nicht verbunden. Bitte {link}, um diese Funktion zu verwenden."
```
devmount, line 466: Awesome, no need for any modifications on this from my site! Sounds natural.

frontend/src/tbpro/elements/CheckboxInput.vue
```diff
@@ -150,5 +150,9 @@ const onChange = () => {
     border-color: var(--colour-highlight);
     color: var(--colour-neutral-raised);
   }
+
+  &:disabled {
+    cursor: not-allowed;
+  }
```
devmount: Would you mind extending this a bit for making the disabled state even more obvious?

```suggestion
  &:disabled {
    cursor: not-allowed;
    opacity: .5;

    & ~ span {
      color: var(--colour-ti-muted);
    }
  }
```

I'm currently working on this too and this is what I used 😇

davinotdavid: Oooh absolutely! That looks better indeed. Thank you so much for the suggested change, I really appreciate it 🙇

frontend/src/locales/de.json
davinotdavid: All kudos to auto translation to be honest as I don't speak German 😅 Speaking of which, how are we currently doing the translation workflow for other languages?

frontend/src/tbpro/elements/CheckboxInput.vue
devmount: Awesome, thanks

frontend/src/locales/de.json
devmount: Well... we don't have a proper flow yet, since we still only support English (obviously) and German (mostly done by myself, since I'm a native speaker 😇 ).

devmount: But I think there are plans to move to a translation platform also used by other Thunderbird/Mozilla projects

devmount: This is tracked here: #296

davinotdavid: Gotcha, thanks! I will keep an eye out for that so I can help with the Brazilian Portuguese translation too 💪 🎉

devmount: Nice, that'd be awesome!

PR #1075, opened 2025-06-26
Switch appointment UUID type

davinotdavid:
## Description of the Change

This PR prepares the move from MySQL -> Postgresql. The pending point was that the `appointments` table had a `uuid` column as `BINARY(16)` but that can be converted to either `CHAR(36)` (if using uuid with hyphens) or `CHAR(32)` (otherwise) [[ref](https://dev.mysql.com/blog-archive/storing-uuid-values-in-mysql-tables/)].

Changes:
- Using [SQLAlchemy's UUID](https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.Uuid) type instead of [SQLAlchemy Utils' UUID](https://sqlalchemy-utils.readthedocs.io/en/latest/data_types.html#module-sqlalchemy_utils.types.uuid) type.
- Add a migration that:
  - Creates a temporary column `uuid_str` to host the new CHAR(32) values
  - Converts the existing BINARY(16) into CHAR(32) and inserts into `uuid_str` table to maintain potential existing data
  - Deletes `uuid` column
  - Rename `uuid_str` to `uuid`

## Benefits

It makes our DB code more agnostic in the uuid usage as SQLAlchemy's UUID states: "_For backends that have no 'native' UUID datatype, the value will make use of CHAR(32) and store the UUID as a 32-character alphanumeric hex string._" [[ref](https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.Uuid)]

## Applicable Issues

https://github.com/thunderbird/appointment/issues/487

### DIFF
--- a/backend/src/appointment/database/models.py
+++ b/backend/src/appointment/database/models.py
@@ -11,8 +11,8 @@ import uuid
 import zoneinfo
 from functools import cached_property
 
-from sqlalchemy import Column, ForeignKey, Integer, String, DateTime, Enum, Boolean, JSON, Date, Time
-from sqlalchemy_utils import StringEncryptedType, ChoiceType, UUIDType
+from sqlalchemy import Column, ForeignKey, Integer, String, DateTime, Enum, Boolean, JSON, Date, Time, UUID
+from sqlalchemy_utils import StringEncryptedType, ChoiceType
 from sqlalchemy_utils.types.encrypted.encrypted_type import AesEngine
 from sqlalchemy.orm import relationship, as_declarative, declared_attr, Mapped
 from sqlalchemy.sql import func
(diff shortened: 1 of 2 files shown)


davinotdavid:
Steps to test, if needed:

0. Have some appointments in your DB to verify that data is not lost
1. While running the app in Docker (per docs), run `docker ps` to take note of the Container ID (or copy it from Docker Desktop)
2. Run `docker exec -it <CONTAINER_ID> bash`
3. Run `alembic upgrade head`
4. Verify that your `appointments` table's `uuid` column has changed types
5. Use appointments as usual to check there's no changes in functionality and no data is lost

MelissaAutumn (approved):
Works as expected on my local machine for both upgrading and downgrading.

Note: The entry script for the backend container auto runs migrations, and you can just use `docker-compose exec backend bash` to jump into the backend container. 

Thank you!!

PR #1076, opened 2025-06-26
WIP: E2E test updates: check availability panel, and stability improvements

rwood-moz:
Some E2E test updates:

- Add a test to check the dashboard availability panel view (with and without the customize per day option on)
- Update the download data test to just verify the download data button is available, as this test was flaky and causing intermittent failures (and downloading user data is not a common user flow)
- Update the appointment booking test so that when checking for the newly created pending appointment, if the new appointment isn't visible yet wait some time and check again (and repeat for up to 2 minutes). Sometimes after the the test requested an appointment slot, the new pending appointment doesn't show up in the pending appointments list immediately and the test would fail intermittently
- Do a similar thing (poll the UI) for some other tests that were failing sometimes when the UI didn't update as fast as expected (i.e. changing the timezone)


rwood-moz:
Note:
- Currently one post-production deployment test (sign-in) is failing but this PR will fix it.
- With this PR the nightly E2E tests job will fail until the next Appointment production deployment (with the custom daily time slot locators)

test/e2e/tests/sign-in.spec.ts
```diff
@@ -28,20 +26,8 @@ test.describe('sign-in', {
...
-      await expect(signInPage.signInHeaderText).toBeVisible({ timeout: TIMEOUT_30_SECONDS }); // generous time for fxa to appear
-      await expect(signInPage.userAvatar).toBeVisible({ timeout: TIMEOUT_30_SECONDS});
-      await expect(signInPage.signInButton).toBeVisible();
-      await signInPage.signIn();
-    } else {
-      await splashscreenPage.localApptSignIn();
-    }
-
-    await expect(page).toHaveTitle(APPT_PAGE_TITLE, { timeout: TIMEOUT_30_SECONDS }); // give generous time for sign-in
+    await navigateToAppointmentAndSignIn(page);
```
davinotdavid, line 29: nice cleanup! 🙌

test/e2e/tests/book-appointment.spec.ts
```diff
@@ -81,9 +81,9 @@ test.describe('book an appointment', () => {
     // now verify a corresponding pending booking was created on the host account's list of pending bookings
     // (drop the day of the week from our time slot string as this function just needs the month, day, and year)
     const expMonthDayYear = expDateStr.substring(expDateStr.indexOf(',') + 2);
-    // wait a few seconds for the appointment dashboard to update, sometimes the test is so fast when it
+    // wait N seconds for the appointment dashboard to update, sometimes the test is so fast when it
     // switches back to the dashboard the new pending appointment hasn't been added/displayed yet
-    await page.waitForTimeout(TIMEOUT_10_SECONDS);
+    await page.waitForTimeout(TIMEOUT_15_SECONDS);
```
davinotdavid, line 86: Just for curiosity, is the event creation delay on showing up >= 10 seconds already to net this change?

test/e2e/tests/availability-panel.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+    // just verify that the default availability options are available/displayed
+    // timzone display is covered in other tests so not checked here
+    expect(await dashboardPage.getAvailabilityPanelHeader()).toContain(APPT_DISPLAY_NAME);
+    await expect(dashboardPage.setAvailabilityText).toBeVisible();
+    await expect(dashboardPage.customizePerDayCheckBox).toBeVisible();
+    await expect(dashboardPage.customizePerDayCheckBox).toBeEnabled();
+
+    // if customize per day checkbox is NOT turned on then we will see overall start and end time inputs
+    const customize_day = await dashboardPage.customizePerDayCheckBox.isChecked();
+    if (!customize_day) {
```
davinotdavid, line 35: Isn't this always going to be the initial state? As in, we are checking if it is `checked` but we are not interacting with it, right? Same for L52-L54 in the same file.

test/e2e/playwright.config.ts
```diff
@@ -26,7 +26,7 @@ export default defineConfig({
   // tests will time out at the locator/test level first anyway; but there is no default so best to specify
   globalTimeout: 12 * 60 * 1000,
   // Individual test timeout - a single test will time out if it is still running after this time (ms)
-  timeout: 90 * 1000, // 1.5 minutes
+  timeout:  150 * 1000, // 2.5 minutes
```
davinotdavid, line 29: Curious to know more about this change, was there any test cases that were culprits on taking too long yet passing?

test/e2e/tests/sign-in.spec.ts
rwood-moz: Thank you :)

test/e2e/tests/book-appointment.spec.ts
rwood-moz: Yes I've seen the test fail sometimes on production because of this; that's why I also added the polling in case the first 15 seconds timeout isn't enough.

test/e2e/tests/availability-panel.spec.ts
rwood-moz: It *should* be but I want to make the test more resilient because if for some reason the test fails and doesn't change the checkbox back to the original state I don't want the test to forever fail in CI after that; or if we mistakenly change that setting in the test account manually. This setting isn't something the E2E tests can set themselves each time before the test runs unlike the other settings that are stored in the local browser store.

test/e2e/playwright.config.ts
rwood-moz: Again this is the worst case scenario that for example the booking request test has to poll for 2 minutes before the new appointment appears in the pending list.

davinotdavid (approved):

Issue #123, opened 2023-07-26
Allow multiple google accounts per Subscriber


radishmouse:
**Is your feature request related to a problem? Please describe.**

Currently, each `Subscriber` is limited to a single Google account, since we store the `google_tkn` in the `Subscriber` table.

However, it's likely that users may have multiple Google accounts that they want to connect to.

**Describe the solution you'd like**

- Move the `google_*` fields to a new `Accounts` table
- Link the `Accounts` table to the `Subscriber` table
- Link the `Calendar` table to the `Accounts` table

(If it makes sense, we could also remove the `owner` field from the `Calendar` table, and derive that knowledge from the linked `Accounts` table)

**Describe alternatives you've considered**

If multiple Google account support isn't on the roadmap, feel free to disregard this ticket!

umgpy:
Any update on this issue? Really an important feature given most have work and personal google accounts.

malini:
@umgpy yes, it's targeted for release by the end of May, however, we can prioritize this issue to see if we can get it in earlier.

umgpy:
Thanks @malini!
Would be happy to test and debug whenever something is ready.

davinotdavid:
@radishmouse as the PR has been merged in, can we close this issue? Happy to continue working on it if there's anything else that you see as a requirement of course 🙇

radishmouse:
Closed by #1078

PR #1078, opened 2025-06-27
[WIP] Allow multiple google accounts

davinotdavid:
## Description of the Change

[**Main changes**]
- Added a [DB migration](https://github.com/thunderbird/appointment/commit/d16c9952143f18d6a689f203226456841a8a5870) to connect `external_connections` with `calendars` to enable add / remove / sync of calendars from different Google Accounts.
  - The migration also attempts to backfill the `external_connection_id` by going through Calendar -> Owner -> External Connection and grabbing its id. Note that we are using `ENUM`s for the `provider`, tried to add SQL comments for clarity.

- Added a `repo.external_connection.get_by_id` method to allow for retrieving the external connection from a `calendar.external_connection_id`.
  - This allowed us to remove the `utils.list_first(repo.external_connection.get_by_type)` for google related flows where were were only considering the first connection.

[**Cleanup / Small refactor**]
- There were some parts of the code that I removed since it seemed unreachable through the Google Calendar flows:
  - `POST` to `/cal`, since it seemed to be only CalDav related + related tests
  - `POST` to `/rmt/calendars` + tests (for now but still need to confirm whether or not this should be restored / CalDav only)

## Benefits

It is a common use-case for people to have multiple Google Accounts (e.g. Work and Personal) so it would be great to add multiple google accounts capabilities.

## Applicable Issues

https://github.com/thunderbird/appointment/issues/123


backend/src/appointment/database/repo/external_connection.py
```diff
@@ -23,6 +23,9 @@ def create(db: Session, external_connection: ExternalConnection):
 def update_token(
     db: Session, token: str, subscriber_id: int, type: models.ExternalConnectionType, type_id: str | None = None
 ):
+    if type == models.ExternalConnectionType.google and type_id is None:
+        raise ValueError("Google external connection requires type_id to be provided")
```
devmount: Is this propagated to the frontend? Then we might want to provide i10n for this. If this is only for us/Sentry, this is fine.

backend/test/integration/test_appointment.py
devmount, line 1: Thanks for the tests!

devmount (commented):
Looking good! I added the corresponding German translation. Thanks for your work!

backend/test/integration/test_general.py
```diff
@@ -44,7 +44,6 @@ def test_health_for_locale(self, with_client):
         ('delete', '/cal/1'),
         ('post', '/caldav/auth'),
         ('post', '/caldav/disconnect'),
-        ('post', '/rmt/calendars'),
```
radishmouse, line 47: Out of curiosity, is this not necessary anymore?

backend/src/appointment/routes/caldav.py
```diff
@@ -134,7 +134,7 @@ def disconnect_account(
     db: Session = Depends(get_db),
     subscriber: models.Subscriber = Depends(get_subscriber),
 ):
-    """Disconnects a google account. Removes associated data from our services and deletes the connection details."""
+    """Disconnects a caldav account. Removes associated data from our services and deletes the connection details."""
```
radishmouse, line 168: :+1:

backend/test/integration/test_general.py
MelissaAutumn: Haha Davi asked me on element about this. It seems like we missed a deprecated api route. I couldn't find it being used anywhere.

backend/src/appointment/controller/calendar.py
```diff
@@ -882,7 +881,29 @@ def existing_events_for_schedule(
         # handle calendar events
         for calendar in calendars:
             if calendar.provider == CalendarProvider.google:
-                google_calendars.append(calendar)
+                # Get the specific external connection for this calendar
```
MelissaAutumn: This will probably bring a big performance hit if someone has a lot of calendars attached. Google's api allows us to batch up calendars for free busy requests, and each request to google's api has a silly amount of overhead time. 😅 

Could you revert this change, and swap the original implementation to use a dict of lists (keyed by external connection id I guess.) And do batching per key?

I'd also be happy if you cleaned up the original implementation because it was a last minute hack to fix performance issues.

backend/src/appointment/database/models.py
```diff
@@ -189,8 +189,15 @@ class Subscriber(HasSoftDelete, Base):
         foreign_keys='[Invite.owner_id]'
     )
 
-    def get_external_connection(self, type: ExternalConnectionType) -> 'ExternalConnections':
-        """Retrieves the first found external connection by type or returns None if not found"""
+    def get_external_connection(
+        self, type: ExternalConnectionType, type_id: str | None = None
+    ) -> 'ExternalConnections':
+        """Retrieves the first found external connection by type and type_id (if provided)
```
MelissaAutumn: Retrieves an owned external connection by type and type_id or the first external connection by type.

backend/src/appointment/database/models.py
```diff
@@ -189,8 +189,15 @@ class Subscriber(HasSoftDelete, Base):
...
+           or returns None if not found"""
+
+        if type_id:
+            return next(filter(lambda ec: ec.type == type and ec.type_id == type_id, self.external_connections), None)
```
MelissaAutumn, line 198: type_id is unique so we can probably do:
`self.external_connections.filter(type=type, type_id=type_id).first()`
(Code not tested, I've been djangoing a bit too much lately to remember if that's how sqlalchemy works.)

backend/src/appointment/database/repo/external_connection.py
MelissaAutumn: This should just propagate to sentry, any frontend errors have to extend APIException iirc.

backend/src/appointment/routes/api.py
```diff
@@ -155,39 +155,19 @@ def create_my_calendar(
...
-        con = CalDavConnector(
-            db=db,
-            redis_instance=None,
-            url=calendar.url,
-            user=calendar.user,
-            password=calendar.password,
-            subscriber_id=subscriber.id,
-            calendar_id=None,
-        )
+    con = CalDavConnector(
```
MelissaAutumn: Can we throw an error if calendar.provider == CalendarProvider.google?

MelissaAutumn: or move this (And the frontend call) to the caldav.py routes file?

backend/src/appointment/database/repo/external_connection.py
davinotdavid: No, it is supposed to be for us / should break as we can't determine which account to update the token for. The frontend shows a generic `Internal server error` page at the moment!

MelissaAutumn (changes requested):
Sorry I guess I got distracted and forgot to push this up

backend/src/appointment/routes/api.py
davinotdavid: Sure! Moved to `caldav.py` in this commit: https://github.com/thunderbird/appointment/pull/1078/commits/ad6c9742f9a480d86e398906193bd3b016dc72e3

backend/src/appointment/database/repo/external_connection.py
davinotdavid: Moved to Sentry, added call to `logging` and returning `None` in this commit: https://github.com/thunderbird/appointment/pull/1078/commits/2279e0bb8e068b3a5ae5cd328a9afa44e45c6c62

backend/src/appointment/database/models.py
davinotdavid: Updated!

backend/src/appointment/controller/calendar.py
davinotdavid: Oh that makes sense, I didn't know that was such a hit! Thanks for letting me know and updated the batching in this commit: https://github.com/thunderbird/appointment/pull/1078/commits/3d50dc1d49da2a75c50f1425662c5c8c9791b9dd

backend/src/appointment/database/models.py
davinotdavid: after some initial reading, seems like `external_connections.filter()` don't work on a SQLAlchemy _relationship collections_  but they work on SQLAlchemy _queries_ so this could potentially be changed to a more "raw"-er SQL queries if better (not super well versed in sqlalchemy bits so please let me know if it doesn't make sense)

PR #1079, opened 2025-07-01
Build parallel infrastructure using Pulumi against new domain names

ryanjjung:
The overall goal of this PR is to create new and completely isolated environments in which to run Appointment. The major differences are as follows:

- Domain names are changing. `appointment.day` is being replaced with `appointment.tb.pro` and its lower-environment counterpart, `appointment-stage.tb.pro` to align with our other offerings. `apmt.day` for short URLs is supplanted with `apt.mt` and `stage.apt.mt`.
- The "old" and currently live infrastructure is managed by Terraform wrapped with Terragrunt and Tofu. The "new" infrastructure is managed by Pulumi using the [tb_pulumi](https://github.com/thunderbird/pulumi/) library.
- Changes have been made to bring together some parts of the build process. We relied on separate dev and deployment Dockerfiles and entrypoint scripts. These have been merged into one process for all of these purposes.

The environments are being built in parallel because the Terraform stuff is live in prod and is what everyone is using. We want to be able to build and tweak these other envs without interfering with the live app. Additionally, we do not have an FXA client for the new domains. Getting that would take a long time, and we are moving away from FXA auth right now. When that task is done, we can meet in the middle with these new environments and really begin testing them. (Realistically, that means a future update to point E2E tests at the new envs; they will definitely fail right now.)

This PR also inserts new steps into the existing Github Actions workflows so that new code changes get deployed to both the live infra and the proposed new infra. This should keep things aligned until we are ready to prove out the services. At some future point when the Terraform infra is no longer in use, we will need a new PR to remove all the Terraform code and the relevant steps in the GHA workflows.

This pertains to Issue #1010. Following that, we will need to migrate traffic to the new infrastructure (#1104). When that is done, we can tear out the old code (#1103).

### DIFF
--- a/.github/workflows/deploy-production.yml
+++ b/.github/workflows/deploy-production.yml
@@ -307,13 +307,15 @@ jobs:
 
       - name: Deploy to prod
         shell: bash
+        env:
+          PULUMI_ACCESS_TOKEN: "${{ secrets.PULUMI_ACCESS_TOKEN }}"
         run: |
           # Update the PATH to include the right version of Pulumi; this is non-trivial or impossible
           # to do with the GHA workflow "env" settings.
           export PATH="/home/runner/.pulumi/bin:$PATH"
 
           # Create a YAML config stump containing only the nested tree leading to the image tag update
-          cd packages/send/pulumi
+          cd pulumi
           cat << EOF > newimage.yaml
           resources:
             tb:fargate:FargateClusterWithLogging:
@@ -330,7 +332,7 @@ jobs:
 
           # Set up the Pulumi environment and update the service
           source ./venv/bin/activate
-          pulumi login s3://tb-send-suite-pulumi
+          pulumi login
           pulumi stack select prod
           TBPULUMI_DISABLE_PROTECTION=True \
             pulumi up -y --diff --target \
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -368,7 +368,7 @@ jobs:
         shell: bash
         run: |
           python -m pip install virtualenv
-          cd packages/send/pulumi
+          cd pulumi
           virtualenv ./venv
 
       - name: Unzip ECR tag (Pulumi)
@@ -383,6 +383,7 @@ jobs:
         shell: bash
         env:
           ECR_TAG: "${{ steps.get_ecr_tag_pulumi.outputs.new_ecr_tag }}"
+          PULUMI_ACCESS_TOKEN: "${{ secrets.PULUMI_ACCESS_TOKEN }}"
         run: |
           # Update the PATH to include the right version of Pulumi; this is non-trivial or impossible
           # to do with the GHA workflow "env" settings above.
@@ -407,7 +408,7 @@ jobs:
           # Set up the Pulumi environment and update the service
           export PULUMI_CONFIG_PASSPHRASE="${{ secrets.PULUMI_PASSPHRASE }}"
           source ./venv/bin/activate
-          pulumi login s3://tb-send-suite-pulumi
+          pulumi login
           pulumi stack select stage
           pulumi up -y --diff --target \
             'urn:pulumi:stage::appointment::tb:fargate:FargateClusterWithLogging$aws:ecs/taskDefinition:TaskDefinition::appointment-stage-fargate-backend-taskdef'


ryanjjung:
At this point, stage is built out, but there's an odd port/URL issue in the frontend that I'm working out.

backend/requirements.txt
```diff
@@ -4,35 +4,36 @@ argon2-cffi-bindings==21.2.0
...
+hiredis==2.3.2
 jinja2==3.1.6
 icalendar==5.0.13
 itsdangerous==2.2.0
 markdown==3.7
 MarkupSafe==2.1.2
 nh3==0.2.21
+oauthlib==3.2.2
+posthog==3.23.0
+psycopg2==2.9.10
```
ryanjjung, line 23: The changes in this file are mostly just me pedantically alphabetizing these dependencies. I do want to highlight the addition of the psycopg2 library, though, which is used later for some changes related to using a Postgres backend (which we are not doing yet).

backend/src/appointment/commands/update_db.py
```diff
@@ -18,6 +19,7 @@ def run():
     alembic_cfg = Config('./alembic.ini')
 
     # If we have our database url env variable set, use that instead!
+    normalize_secrets()
```
ryanjjung, line 22: This call brings in our database settings from an environment variable. This variable is set in AWS Secrets Manager.

backend/src/appointment/routes/schedule.py
```diff
@@ -9,7 +9,6 @@
 from sqlalchemy.exc import SQLAlchemyError
 from sqlalchemy.orm import Session
 
-from .. import utils
```
ryanjjung, line 12: Ruff auto-removed this unused import.

backend/src/appointment/utils.py
```diff
@@ -117,8 +120,18 @@ def normalize_secrets():
...
+            driver = 'mysqldb'
+        elif dialect == 'postgresql':
+            driver = 'psycopg2'
+        else:
+            driver = None
+        proto = f'{dialect}+{driver}' if driver else dialect
+
         os.environ['DATABASE_URL'] = (
-            f"mysql+mysqldb://{secrets['username']}:{secrets['password']}@{hostname}/appointment"
+            f"{proto}://{secrets['username']}:{secrets['password']}@{hostname}/{secrets['dbname']}"
```
ryanjjung, line 134: We are not currently using Postgres. In the process of working this ticket, I did test it out. I ran into some bugs, which were fixed elsewhere, but I have not gone back to test this again. For the time being, I am leaving this code in place, which allows us to support both data stores while we work through migration issues.

.github/workflows/deploy-production.yml
```diff
@@ -151,6 +151,30 @@ jobs:
...
+        run: rm -rf dist
+
+      - name: Get Artifact from Release (Pulumi)
+        uses: dsaltares/fetch-gh-release-asset@master
+        with:
+          version: ${{ github.event.release.id }}
+          file: pulumifrontend.zip
+
+      - name: Unzip Artifact (Pulumi)
+        run: unzip pulumifrontend.zip
```
ryanjjung, line 171: These have to be separate files for now because the builds use different domain names to refer to the different environments. Later, as part of #1103, we will return these to more sensible names.

MelissaAutumn:
Fun fact: When we have #1105 merged we can enable it for just appointment.tb.pro using env variables. That way we'll be able to test out the new infra.

.github/workflows/deploy-production.yml
```diff
@@ -151,6 +151,30 @@ jobs:
...
+
+      - name: Deploy frontend to S3 (Pulumi)
+        run: |
+          aws s3 sync ./dist "s3://${{ secrets.FRONTEND_BUCKET }}"
+          aws cloudfront create-invalidation --distribution-id ${{ secrets.FRONTEND_CF_DISTRO_ID }} --paths "/*"
```
MelissaAutumn, line 176: Are these new secrets? Or will this potentially overlap with the tofu buckets?

.github/workflows/deploy-staging.yml
```diff
@@ -188,6 +188,24 @@ jobs:
...
+      
+      - name: Build stage frontend (Pulumi)
+        run: |
+          rm -rf frontent/dist  # Clean up from previous build
+          cp frontend/.env.pulumistage.example frontend/.env.stage
+          cd frontend && npm run build -- --mode ${{ vars.APP_ENV }}
+
+      - name: Deploy frontend to S3 (Pulumi)
+        run: |
+          aws s3 sync frontend/dist s3://${{ secrets.FRONTEND_BUCKET }}
```
MelissaAutumn, line 207: Ditto

.github/workflows/deploy-production.yml
```diff
@@ -234,6 +258,85 @@ jobs:
...
+          # Create a YAML config stump containing only the nested tree leading to the image tag update
+          cd packages/send/pulumi
+          cat << EOF > newimage.yaml
+          resources:
+            tb:fargate:FargateClusterWithLogging:
+              backend:
+                task_definition:
+                  container_definitions:
+                    backend:
+                      image: "${{ steps.pulumi-tag-extract.outputs.new_ecr_tag }}"
```
MelissaAutumn, line 326: This step seems to differ from stage, can we normalize them?

pulumi/Pulumi.prod.yaml
```diff
@@ -0,0 +1,19 @@
...
+  cloudflare:apiToken:
+    secure: AAABAAKhnDM/GGaF24PhvSpcC0CNapmKminJzqP8QBpE32uKHDG0S9ae2jG3q5BDc7Qxx4JsIpG1t3G2TBwMZheohDy+VXQR
```
MelissaAutumn, line 19: I'll be super excited to eventually removed the bundled secrets and the normalize_secrets function. But I guess they continue to live for now 😅

pulumi/__main__.py
```diff
@@ -0,0 +1,112 @@
...
+backend_cache_sg, container_sgs, lb_sgs = security_groups(project=project, resources=resources, vpc=vpc)
+
+# Create the Redis memory cache
+backend_cache, cache_dns = redis_cache(
+    cloudflare_zone_id=cloudflare_zone_id,
+    project=project,
+    security_group=backend_cache_sg,
+    vpc=vpc,
+    resources=resources,
+)
```
MelissaAutumn, line 60: Is this a cluster or single instance?

pulumi/config.prod.yaml
```diff
@@ -0,0 +1,224 @@
...
+      recovery_window_in_days: 0
+      secret_names:
+        - database-connection
+        - database-encryption
+        - fxa-secrets
+        - google-oauth
+        - smtp-connection
+        - x-allow-secret
+        - zoom-secrets
+        # - neon-database-connection
```
MelissaAutumn, line 97: Left over?

MelissaAutumn (commented):
Few questions, I'll need to look over this a few more times as infra code always makes my head hurt. (Not your fault, my mind is still somewhere between opentofu-land and aws' labyrinth of a webui.)

.github/workflows/deploy-production.yml
ryanjjung: None of these overlap. I had to create all of these. The ones that vary per environment I set up as environment secrets for "staging" and "prod".

.github/workflows/deploy-staging.yml
ryanjjung: Ditto, these are all new variables.

pulumi/Pulumi.prod.yaml
ryanjjung: If your idea is to just pull each component (host, port, user, etc.) from a separate secret, that's very do-able. If there's a ticket, I can maybe even put that on my queue. If not, I'll make one.

MelissaAutumn: Yep would clean up the setup function. No worries for this ticket, as long as there is a ticket for it!

.github/workflows/deploy-production.yml
ryanjjung: I did find some issues here which I've fixed (and will push shortly after I've looked through the rest of this review), but we can't completely normalize these steps. Maybe we could variablize the environment names. But they get their image tag for deployment from two different places. The stage build gets the tag from a previous step that does the build. The stage workflow later packages that up as an artifact and staples it to a draft release. If we choose to complete that release, the deploy-production workflow will pick it up from a job that unpacks the artifacts.

pulumi/__main__.py
ryanjjung: It's a "serverless cache". I'm not sure if it runs in cluster mode or not. The AWS console only shows one endpoint. But it auto-scales in response to usage. This is the same kind of resource that the Terraform code is building, so there is no change (except to the endpoint) to make on the application side.

pulumi/config.prod.yaml
ryanjjung: Kind of. I did the Postgres testing in stage and left some of that code in place in this PR. I kept this secret around because the Neon DB exists. Since it didn't work, I never built one in prod, but I anticipate doing so later. So I left it in and commented out for now.

pulumi/Pulumi.prod.yaml
ryanjjung: https://github.com/thunderbird/appointment/issues/1110

pulumi/__main__.py
MelissaAutumn: Okie so it should be cluster. There's a slight difference in how we connect to it. I don't think we need to do anything but there may be a "use cluster pls" flag.

MelissaAutumn (approved):
lgtm!

Issue #1067, opened 2025-06-20
Switch to PostgreSQL against NeonDB


ryanjjung:
When we move to the new environment, we will switch to using NeonDB, which uses PostgreSQL. We will need to tweak the SQLAlchemy config to use Postgres.

ryanjjung:
While working on #728 I made a quick attempt at hooking this service up to a Neon database. That meant a little legwork to support both backends, and that code will get merged with the work for the other issue. However, the code doesn't work as-is by just swapping out the ORM config. There are some queries somewhere (the `CREATE TABLE`s ones) that make use of incompatible data types. The actual logs are very ugly, but here are the important bits...

The exception thrown is:

```
sqlalchemy.exc.ProgrammingError: (psycopg2.errors.UndefinedObject) type "binary" does not exist
```

This is in reference to a `CREATE TABLE appointments ...` query which has a `uuid` field with datatype `BINARY(16)`.

I don't want to get bogged down in this ticket while working through the other, so I'm going to stop trying to solve this problem for the moment. There may be other places where we have incompatible data types or other issues.

MelissaAutumn:
Okay so this ticket is blocked by https://github.com/thunderbird/appointment/issues/487

MelissaAutumn:
Should be unblocked now. Pull the latest main commit and let the entry script run the migrations if needed.

davinotdavid:
Linking similar task: https://github.com/thunderbird/appointment/issues/855

ryanjjung:
Now working through a series of migration tasks. Will try to remember to link PRs back to this issue.

Issue #855, opened 2025-02-05
Swap to postgres+Neon


malini:
**Is your feature request related to a problem? Please describe.**
If we're in postgres, we'll be able to use Neon's scalable db

MelissaAutumn:
+1 it'd be nice to be able to branch an existing database and use it to debug issues that may crop up.

davinotdavid:
Linking related issues for tracking:

https://github.com/thunderbird/appointment/issues/1126

davinotdavid:
This has already been done! Closing

PR #1084, opened 2025-07-03
Switch from MySQL to PostgreSQL

davinotdavid:
## Description of the Change

- Changed adapter from `mysqldb` to `[psycopg](https://pypi.org/project/psycopg/)` (psycopg3)
- Updated `docker-compose.yml`
- Updated project architecture diagram with the recent DB changes
  - Added `external_connection_id` foreign key in `Calendars`
  - Changed `uuid` comment to be `Uuid` instead of `Binary`

Note: I did not update the `tofu` / `infra` yet so that it can be done alongside the Neon (?) integration

> [!WARNING]  
> It is necessary to delete all the container / volumes for this to work so you will lose all your data when testing this

To remove all your volumes, navigate to your `appointment` root folder and:
```bash
$ docker-compose down -v
```

## Benefits

- Hopefully a more seamless transition to Neon's Postgresql and having a local dev environment that will use the same DB backend as the other environments

## Applicable Issues

https://github.com/thunderbird/appointment/issues/855
https://github.com/thunderbird/appointment/issues/1067

### DIFF
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -17,6 +17,7 @@ cli = [
     "ruff",
 ]
 db = [
+    "mysqlclient==2.2.5",
     "psycopg==3.2.9",
 ]
 test = [
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -17,7 +17,7 @@ cli = [
     "ruff",
 ]
 db = [
-    "mysqlclient==2.2.5",
+    "psycopg==3.2.9",
 ]
 test = [
     "Faker==26.0.0",
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ Copy the Client ID and Client Secret values to your backend's .env file as `TB_A
 * Backend can be accessed via: <http://localhost:5000>
 * OpenAPI docs can be accessed via: <http://localhost:5000/docs> or <http://localhost:5000/redoc>
 
-A MySQL database will be accessible via `localhost:3306` with username and password set to: `tba`
+A PostgreSQL database will be accessible via `localhost:5433` with username and password set to: `tba`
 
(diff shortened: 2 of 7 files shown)


MelissaAutumn:
I've added some labels to ensure we don't merge this accidentally. 

We'll need to push existing changes up to prod first, and plan how we're going to move the data over from mysql to postgres. If we end up nuking the db then we can also switch to uuids as pk 🤔

ryanjjung:
Additionally, I just [merged some changes](https://github.com/thunderbird/appointment/pull/1161) that change how database connection strings are handled. Should probably pull/merge `main` and adjust accordingly.

MelissaAutumn:
> Additionally, I just [merged some changes](https://github.com/thunderbird/appointment/pull/1161) that change how database connection strings are handled. Should probably pull/merge `main` and adjust accordingly.

Please do, but use `git rebase main` instead of merging. Cleaner even if it's slightly more painful to perform.

davinotdavid:
@MelissaAutumn @ryanjjung 

Alright, rebased! Thanks for the heads up!

I've kept the comments related to mysql _and_ to postgresql just for reference since we are on the process of deprecating the `DATABASE_URL` settings in Docker. I gave it a spin in my local end after nuking all the data it seems to work well on my end.

Is this still blocked or does it need more changes before we can fully transition to postgresland?

backend/pyproject.toml
```diff
@@ -17,7 +17,7 @@ cli = [
     "ruff",
 ]
 db = [
-    "mysqlclient==2.2.5",
+    "psycopg==3.2.9",
```
MelissaAutumn, line 21: we'll need both atm

MelissaAutumn:
Good once you re-add the mysqlclient driver.

backend/pyproject.toml
davinotdavid: Good point, thanks for the context! Re-added and tested locally.

Issue #1085, opened 2025-07-04
Remove `meeting:update:meeting` zoom scope


MelissaAutumn:
I don't believe we use this scope and it's currently blocking our zoom review. 

If we deploy rescheduling features we should just delete and re-create the meeting link.

davinotdavid:
@MelissaAutumn since the PR has been merged, do we wait for the deploy to prod before closing this issue or is it enough to be on staging?

MelissaAutumn:
Awaiting on @malini's submission

davinotdavid:
Per today's Services meeting, closing this as it has already been deployed!

PR #1086, opened 2025-07-04
Remove Zoom scope meeting:update:meeting

davinotdavid:
## Description of the Change

Removes unnecessary Zoom scope `meeting:update:meeting` as it is blocking our app approval process and we can circumvent the permission scope by deleting and creating a new zoom link.

## Benefits

Unblocks app approval process

## Applicable Issues

https://github.com/thunderbird/appointment/issues/1085


MelissaAutumn (approved):
Thanks, this needs to be deployed next week but it's okay to go live on stage now since no one should be using that.

Issue #1088, opened 2025-07-08
Add "copy booking link" button to NavBar and show email address next to avatar


malini:
**Is your feature request related to a problem? Please describe.**
From the updated UI wireframes:

<img width="531" height="96" alt="Image" src="https://github.com/user-attachments/assets/7ab9618d-43dd-4518-b8d5-2dcae8da9255" />

Let's add a quick link for getting the booking link, and add the user's email address next to the avatar

davinotdavid:
@malini PR has been merged and available on [staging](https://stage.appointment.day/) 🚀

MelissaAutumn:
<img width="256" height="82" alt="Image" src="https://github.com/user-attachments/assets/e67672f4-6428-4c87-8109-9f95f35b6661" />
https://mozillausercontent.com/cec8f11ca10d5cae787679d521ca05e5

please help my cat is now very long 🙀 

Looking good, just one small issue. Looks like the forced height property is messing it up. 😄 

I'd suggest dropping the use of tailwind for these components too, as we want to move away from using it.

davinotdavid:
@MelissaAutumn LOL that legit gave me a good laugh, thanks for spotting that! Opened a PR for it even though I will miss long cat 

https://github.com/thunderbird/appointment/pull/1100

davinotdavid:
Closing as PR has been merged and it is live in staging for a while!

PR #1091, opened 2025-07-08
Add copy link icon and email to navbar

davinotdavid:
## Description of the Change

As part of the UI overhaul, adding a copy link icon button and user email to the navbar per wireframes.

Jul-08-2025 14-42-18

## Benefits

Be able to copy the appointment link from anywhere is handy! Also, seeing the name of the account is more clear than just the initials on the user avatar.

## Applicable Issues

https://github.com/thunderbird/appointment/issues/1088

### DIFF
--- a/frontend/src/components/NavBar.vue
+++ b/frontend/src/components/NavBar.vue
@@ -7,7 +7,8 @@ import UserAvatar from '@/elements/UserAvatar.vue';
 import DropDown from '@/elements/DropDown.vue';
 import NavBarItem from '@/elements/NavBarItem.vue';
 import TextButton from '@/elements/TextButton.vue';
-import LinkButton from '@/tbpro/elements/LinkButton.vue';
+import ToolTip from '@/tbpro/elements/ToolTip.vue';
+import { TooltipPosition } from '@/definitions';
 import { IconExternalLink, IconLink } from '@tabler/icons-vue';
 
 // component constants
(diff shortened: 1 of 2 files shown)


frontend/src/components/NavBar.vue
```diff
@@ -71,10 +86,32 @@ const isNavEntryActive = (item: string) => {
           :label="t(`label.${item}`)"
           :link-name="item"
         />
+        <div v-if="user.myLink" class="flex items-center justify-center pl-8 pr-12 relative">
```
devmount: Can we move the link icon a bit closer to the dropdown like in the mock-up? It seems to be a bit lost there, e.g.:

```suggestion
        <div v-if="user.myLink" class="flex items-center justify-center px-4 relative">
```

frontend/src/components/NavBar.vue
```diff
@@ -71,10 +86,32 @@ const isNavEntryActive = (item: string) => {
...
+              {{ myLinkTooltip }}
+            </tool-tip>
+          </button>
+        </div>
       </div>
+
       <drop-down class="self-center" ref="profileDropdown">
         <template #trigger>
-          <user-avatar />
+          <div class="flex items-center gap-4 border rounded-md border-gray-300 px-3 py-1 bg-white">
```
devmount: To keep the margin between avatar and outer border consistent, I would reduce the right margin like this:

```suggestion
          <div class="flex items-center gap-4 border rounded-md border-gray-300 pl-3 pr-1 py-1 bg-white">
```

frontend/src/elements/UserAvatar.vue
```diff
@@ -7,7 +7,7 @@ const user = useUserStore();
 
 <template>
 <div
-  class="flex-center mr-4 size-12 self-center rounded-full bg-white text-lg font-normal text-white"
+  class="flex-center size-10 self-center rounded-full bg-white text-lg font-normal text-white"
```
devmount: I would even reduce the size a bit more and make the initials more fitting:


```suggestion
  class="flex-center size-8 self-center rounded-full bg-white text-sm font-semibold text-white"
```

devmount (approved):
Wow, you're really fast! Thanks for implementing this, works great on my end. I suggested some visual improvements (in my opinion! please ignore them if you think those are not suitable). They would make the navbar look like this:

Image

For reference, this is how it looks without my suggestions:
image

davinotdavid:
@devmount thank you so much for the kind words / review! Applied your changes as I also agree that they look better! Thanks for suggesting 🙇

Issue #1089, opened 2025-07-08
Prod deploy is broken


MelissaAutumn:
Probably due to a plugin update

last build:
https://github.com/thunderbird/appointment/actions/runs/16147906970/job/45571982457

PR #1098, opened 2025-07-09
Lock Terraform AWS provider version

ryanjjung:
AWS released a new major version of their Terraform provider with a number of significant changes in it. Since we are replacing this Terraform code with Pulumi code, it's not worth it to do a full provider upgrade here. Instead, I'm replacing the [version constraint](https://developer.hashicorp.com/terraform/language/providers/requirements#version-constraints) with this `~=` constraint, which allows the rightmost semver digit to increment. This will prevent us from using the new major version while still getting the latest from the old major version.


ryanjjung:
This should resolve #1089

PR #1100, opened 2025-07-09
Fix user avatar image styling + remove Tailwind classes

davinotdavid:
## Description of the Change

There as a forced height on the user avatar image that was causing distortions. This PR now removes the forced height in favor of its container's.

It also moves the `UserAvatar` component away from Tailwind-specific classes as we transition over to vanilla CSS. Note that we were using the `teal` color from Tailwind and our shade of teal in the CSS variables is slightly different.

Before:
image

image

After:
image

image

## Benefits

No more image distortions + one component at a time to help the Tailwind -> vanilla CSS transition

## Applicable Issues

https://github.com/thunderbird/appointment/issues/1088

### DIFF
--- a/frontend/src/elements/UserAvatar.vue
+++ b/frontend/src/elements/UserAvatar.vue
@@ -7,15 +7,39 @@ const user = useUserStore();
 
 <template>
 <div
-  class="flex-center size-8 self-center rounded-full bg-white text-sm font-semibold text-white"
-  :class="{'!bg-teal-500': user.data.avatarUrl === null}"
+  class="user-avatar"
+  :class="{'user-avatar--fallback': user.data.avatarUrl === null}"
   data-testid="user-menu-avatar"
 >
   <span v-if="user.data.avatarUrl === null">
     {{ initials(user.data.name) }}
   </span>
   <span v-else>
-    <img class="size-[48px] rounded-full" :alt="initials(user.data.name)" :src="user.data.avatarUrl"/>
+    <img class="user-avatar__image" :alt="initials(user.data.name)" :src="user.data.avatarUrl"/>
   </span>
 </div>
 </template>
+
+<style scoped>
+.user-avatar {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  width: 2rem;
+  height: 2rem;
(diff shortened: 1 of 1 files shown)


frontend/src/elements/UserAvatar.vue
```diff
@@ -7,15 +7,39 @@ const user = useUserStore();
...
+  height: 2rem;
+  align-self: center;
+  border-radius: 50%;
+  background-color: var(--colour-neutral-base);
+  font-size: 0.875rem;
+  font-weight: 600;
+  color: var(--colour-neutral-base);
+}
+
+.user-avatar--fallback {
```
MelissaAutumn: I don't mind the `__` convention. Can we switch the -- to __ here too?

MelissaAutumn (commented):
Looks good, just one thing!

frontend/src/elements/UserAvatar.vue
MelissaAutumn: The style is also scoped so we probably drop the prefixes too. It'll be built as like user-avatar-<random hash>.

davinotdavid: Hmm, do you mean that we should prioritize targeting the element itself when possible since it is prefixed anyways? Something like:

```
div {
  display: flex;
  align-items: center;
  justify-content: center;
  width: 2rem;
  height: 2rem;
  align-self: center;
  border-radius: 50%;
  background-color: var(--colour-neutral-base);
  font-size: 0.875rem;
  font-weight: 600;
  color: var(--colour-neutral-base);
}

.fallback {
  background-color: var(--colour-accent-teal);
}

img {
  border-radius: 50%;
}
```

MelissaAutumn: Oh we could still use class, but we could simplify it to user-avatar, fallback, and image. We haven't really nailed down an exact style guide on how to write components.

davinotdavid: Ah gotcha, thanks! Updated and will keep in mind for next tasks 👍

MelissaAutumn (approved):
cool beans, thanks!

Issue #1101, opened 2025-07-09
Create SlidingPanel component


MelissaAutumn:
## Screenshot

<img width="566" height="580" alt="Image" src="https://github.com/user-attachments/assets/5cdf9f40-90ac-4d59-87b4-55167768b2ca" />

<img width="571" height="577" alt="Image" src="https://github.com/user-attachments/assets/e804cee1-a176-4f29-b3f4-9ca6c0f94a24" />
<img width="537" height="547" alt="Image" src="https://github.com/user-attachments/assets/526b6c6a-909a-4b90-8e97-b4cb8b3c1dde" />
<img width="546" height="570" alt="Image" src="https://github.com/user-attachments/assets/22036387-1bbe-492b-8f39-8ee171244be6" />
<img width="531" height="554" alt="Image" src="https://github.com/user-attachments/assets/f5b7a222-1b5b-42c4-90db-42cf6a0d4691" />
<img width="533" height="554" alt="Image" src="https://github.com/user-attachments/assets/5693b2e3-3902-46fa-b80c-aff0c539f9af" />
<img width="536" height="555" alt="Image" src="https://github.com/user-attachments/assets/e2af26cf-d895-4a04-982d-0a0befcd64e3" />
<img width="529" height="551" alt="Image" src="https://github.com/user-attachments/assets/e388a96d-4171-469b-b663-f06c1402e450" />
<img width="518" height="533" alt="Image" src="https://github.com/user-attachments/assets/8a7c4507-b584-478e-a069-2ff79f018b50" />
<img width="543" height="525" alt="Image" src="https://github.com/user-attachments/assets/a665c298-c10e-438a-854d-0923e3d64e91" />

## Details

* Should contain a full-screen scrim (partial opacity blocking mouse actions) in the background.
* Should contain a the default slot allowing whatever html we want in there.
* Should contain a cta/button slot allowing for buttons, empty be default.
* There should be a standard close button at the top right.
* There should be a title slot or property for a standard title at the top left.
* This should not live in services-ui as it's not planned for other services.

We should then extend this component for the individual instances in the wireframes.

davinotdavid:
Closing as the PRs related to this issue have been merged!

Issue #1095, opened 2025-07-08
Bookings page overhaul


malini:
From the updated UI wireframes:

<img width="730" height="398" alt="Image" src="https://github.com/user-attachments/assets/9aca2b09-06af-4fc9-b3a9-b9a96f7c5099" />

and slide out panel:
<img width="462" height="394" alt="Image" src="https://github.com/user-attachments/assets/349c3bb6-185f-4cf7-b81f-6ddae1f262c8" />

* Remove calendar from bookings
* Make bookings table sortable with Date, Time, Meeting Title, Status, Calendar columns
* Move booking filters to a dropdown and a checkbox for unconfirmed first
* Show booking details on a SlidingPanel upon selecting a booking

MelissaAutumn:
I've split the sliding panel to #1101

davinotdavid:
Closing this as all PRs related to this issue has been merged!

As a reminder, the Modify Appointment feature is still under scoping so that has not been fully implemented.

PR #1102, opened 2025-07-09
Add sliding panel component

davinotdavid:
## Description of the Change

- Adds a new `SlidingPanel` component
  - Slots
    - Default slot is the main content
    - Title slot is for the title
    - CTA slot is for the buttons (if slot is unused it doesn't render the bottom part)
  - Props
    - `side` prop controls whether the slide comes from the left or from the right and adjusts the close button + title positioning for RTL languages
    - `:isOpen` to control the panel open / close
    - `:title` for the title
    - `@close` to emit the close event
- Removes the existing `AppointmentModal` in favor of a new `AppointmentSlidingPanel` that uses the `SlidingPanel` component
  - The code for the modal vs panel is largely unchanged

https://github.com/user-attachments/assets/bcfbf518-0e82-4e5e-9fe6-96f8d5812488

## Benefits

- A new SlidingPanel component
- A showcase of its usage in the booking details

> [!NOTE]  
> This does not cover all the scenarios and UI details of the previous AppointmentModal. The intention is to showcase the usage of the sliding panel more testing purposes as this PR is meant to be focused on the SlidingPanel component. The rest of the booking UI / UX details will be done as part of https://github.com/thunderbird/appointment/issues/1095

## Applicable Issues

https://github.com/thunderbird/appointment/issues/1101


frontend/src/views/AppointmentsView.vue
```diff
@@ -195,13 +195,14 @@ const filteredAppointments = computed(() => {
 
 // handle single appointment modal
 const showAppointment = ref(null);
-const showAppointmentModal = (appointment) => {
+const isPanelOpen = computed(() => showAppointment.value !== null);
+
+const showAppointmentSlidingPanel = (appointment) => {
   showAppointment.value = appointment;
   router.replace(`/appointments/${viewToParam(tabActive.value)}/${appointment.slug}`);
```
MelissaAutumn, line 205: I'd like to be able to middle click on these items so I can open them in new tabs. Think it'd be hard to switch this function for an inlined anchor tag that covers the entire tr?

frontend/src/elements/SlidingPanel.vue
```diff
@@ -0,0 +1,233 @@
...
+  position: fixed;
+  top: 0;
+  background-color: var(--colour-neutral-base);
+  display: flex;
+  flex-direction: column;
+  width: 600px;
+  height: 100vh;
+  overflow: hidden;
+  z-index: 1001;
+}
```
MelissaAutumn: sliding-panel doesn't work under about 600px. Could we swap the width for max-width: 600px, and do width: 100%; I think that should fix it, if not we can add a breakpoint for smol.

frontend/src/elements/SlidingPanel.vue
```diff
@@ -0,0 +1,233 @@
...
+<template>
+  <Teleport to="body">
+    <Transition name="scrim">
+      <div
+        v-if="isOpen"
+        class="scrim"
+        @click="handleScrimClick"
+      />
+    </Transition>
+    <Transition :name="`panel-${side}`">
```
MelissaAutumn: Can we add a prefers-reduced-motion to remove the slide in effect?
https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion

You can test it by setting `ui.prefersReducedMotion` to `1` in your `about:config`.

frontend/src/elements/SlidingPanel.vue
```diff
@@ -0,0 +1,233 @@
...
+      <div v-if="isOpen" :class="panelClasses" @click.stop>
```
MelissaAutumn: I think we'll need to add a `role="dialog"` here so that voiceover and other screenreaders know there's a modal open. 

Additionally I think if you build this element around the [dialog element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog) it should work by default.

frontend/src/components/AppointmentSlidingPanel.vue
```diff
@@ -0,0 +1,413 @@
...
+    <div v-if="appointment" class="appointment-content">
+      <!-- Appointment status -->
+      <p :class="['status-label', bookingStatusInfo.color]">
+        {{ bookingStatusInfo.label }}
+      </p>
+
+      <div class="time-slots">
+        <template v-for="s in appointment.slots" :key="s.start">
+          <div class="time-slot">
+            <icon-calendar-event class="time-icon"/>
```
MelissaAutumn: Can we add a label/alt/title or aria-label to describe the icon? I think "Time of the event:" should fit.

MelissaAutumn (changes requested):
Looking good, just a few accessibility related things I've noticed, and one wish list 😅

frontend/src/views/AppointmentsView.vue
davinotdavid: I like this wish list item! The way that I could think about making them links without breaking layout nor behavior was a bit wacky [[commit link here](https://github.com/thunderbird/appointment/pull/1102/commits/4ba53d323cc29618d3bf98ec6a78ca7d194c7ce5)] but it works and VoiceOver seems to be able to access it well too.

In case this is not a super desirable approach perhaps we can revisit it in the [Booking Page Overhaul PR](https://github.com/thunderbird/appointment/pull/1106)?

frontend/src/elements/SlidingPanel.vue
davinotdavid: Absolutely yeah, my bad! Changed here.

frontend/src/elements/SlidingPanel.vue
davinotdavid: Yes! Thanks for asking in Matrix, I will keep in mind for developing without animations first and only if there's no-preference set, include it!

frontend/src/elements/SlidingPanel.vue
davinotdavid: Thanks! I didn't know much about the `<dialog>` element but I've refactored to use of it. One caveat is that it seems that browser support for exit transitions is a little all over the place at the moment so while it animates in with the slide effect, it disappears when closing.

After trying a bunch and failing, I've found some demos showcasing the problem:
https://frontendmasters.com/blog/the-dialog-element-with-entry-and-exit-animations/
https://developer.chrome.com/blog/entry-exit-animations#animating_elements_to_and_from_the_top-layer

On Chrome it works, Safari kinda works (a little clunky offset) and Firefox doesn't seem to work at all.

Unsure if it is a hard-requirement for the Design team but in case it is, do you have a suggestion on how to workaround this? Happy to pair!

frontend/src/components/AppointmentSlidingPanel.vue
davinotdavid: Absolutely yes, my bad! Added!

frontend/src/views/AppointmentsView.vue
MelissaAutumn: It works almost okay, but middle clicking opens both. I think we can prevent middle click events 

https://vuejs.org/guide/essentials/event-handling.html#mouse-button-modifiers

I think using `click.left.exact` will work. Just `click.left` will trigger on my command+clicks.

frontend/src/components/AppointmentSlidingPanel.vue
```diff
@@ -70,24 +62,33 @@ const bookingStatusInfo = computed(() => {
...
   apmtStore.deleteAppointment(props.appointment?.id);
+  panelRef.value?.closePanel();
   emit('close');
 };
 
 const cancelAppointment = () => {
   apmtStore.cancelAppointment(props.appointment?.id, cancelReason.value);
   cancelReason.value = '';
+  panelRef.value?.closePanel();
   emit('close');
```
MelissaAutumn: Can you put the panel.closePanel and emit('close') into one function (maybe closePanel?) And call it where needed.

frontend/src/elements/SlidingPanel.vue
```diff
@@ -46,112 +27,91 @@ const headerClasses = computed(() => [
...
+    <!-- Content -->
+    <div class="content">
+      <slot />
+    </div>
+
+    <!-- CTA Container -->
+    <div v-if="$slots.cta" class="cta-container">
+      <slot name="cta" />
+    </div>
+  </dialog>
```
MelissaAutumn, line 86: We can punt this to a different ticket but could we get a hidden link that's a "back to top"?

frontend/src/elements/SlidingPanel.vue
```diff
@@ -46,112 +27,91 @@ const headerClasses = computed(() => [
...
+  closePanel
+})
+</script>
 
-.panel-right-enter-to,
-.panel-right-leave-from {
-  transform: translateX(0);
-}
+<template>
+  <dialog ref="panel" :class="panelClasses" @click="handleDialogClick">
```
MelissaAutumn: Non-blocking can be moved to a different ticket, but for some reason in voice over the panel opens on the close button instead of the title. I think the title might need an autofocus attribute.

MelissaAutumn (approved):
A few non-blocking issues that can be punted to other tickets.

Also could you add a background colour to `.content` using the colour variables?

frontend/src/elements/SlidingPanel.vue
MelissaAutumn: We'll run with this until someone complains 😅

frontend/src/elements/SlidingPanel.vue
davinotdavid: Definitely! Created a new ticket for it here: https://github.com/thunderbird/appointment/issues/1109

frontend/src/elements/SlidingPanel.vue
davinotdavid: I agree that it is a bit weird but it seems that the `autofocus` should be placed on the close button in dialogs per [MDN <dialog> usage docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog#usage_notes) or it can be placed in the dialog itself.

I've added the `autofocus` to the dialog itself so at least there's a "downwards" movement path on the tree to access the title

frontend/src/components/AppointmentSlidingPanel.vue
davinotdavid: For sure, changed!

frontend/src/views/AppointmentsView.vue
davinotdavid: oh TIL about `.exact`, interesting! I've added `click.left.exact` and it works well, thanks!

PR #1105, opened 2025-07-10
OIDC Authentication

### LINKED ISSUE #1066: Switch to a generic oidc library for authentication
We'll be switching to keycloak. We should probably using a generic OIDC library to auth with keycloak instead of building out the flow ourselves. This will allow better open source usage. 

We'll authenticate directly against the OIDC for login/registration by linking to their login page (like how we do for fxa right now.) But we'll use accounts.tb.pro for user management and billing. So we'll need an env var for something like a `ACCOUNTS_MGMT_URL` pointing to `https://accounts.tb.pro/self-serve/`.

MelissaAutumn:
Fixes #1066 

I've implemented oidc-client-ts in the frontend and used authlib in the backend to verify the frontend token via token-introspect.

The credentials for testing locally are in `[Local] Appointment OIDC/Keycloak Settings` in the services vault. You can create an account in the tbpro-dev realm: https://auth-stage.tb.pro/admin/master/console/#/tbpro-dev/users (login with the stage admin user also in 1pass.)

You should be able to login, logout, and do about everything the same as if you had any other auth.


docker-compose.yml
```diff
@@ -28,7 +27,7 @@ services:
   mysql:
     image: mysql:8
     volumes:
-      - db:/var/lib/mysql
+      - apmt_db:/var/lib/mysql
```
MelissaAutumn, line 32: This is just so if you switch back to a different branch you won't have a db in a weird state.

backend/src/appointment/routes/auth.py
```diff
@@ -630,3 +633,86 @@ def permission_check(subscriber: Subscriber = Depends(get_admin_subscriber)):
...
+        if not is_in_allow_list:
+            # Use the invite code after we've created the new subscriber
+            used = repo.invite.use_code(db, data.invite_code, subscriber.id)
+
+            # This shouldn't happen, but just in case!
+            if not used:
+                repo.subscriber.hard_delete(db, subscriber)
+                raise HTTPException(500, l10n('unknown-error'))
+
+    # FIXME: OIDC should handle this check
```
davinotdavid, line 691: Is this going to handled after this PR / is there something missing from the OIDC implementation that blocks this?

frontend/src/components/SettingsConnections.vue
```diff
@@ -158,7 +159,8 @@ const editProfile = async () => {
           <div v-if="connection.length > 0">
             <div v-for="conn in connection" :key="conn.type_id" class="mb-2 flex items-center justify-between">
               <p>{{ t('label.connectedAs', { name: conn.name }) }}</p>
-              <div v-if="providers[provider] !== ExternalConnectionProviders.Fxa && providers[provider] !== ExternalConnectionProviders.Accounts">
+              {{ provider }}  - {{providers[provider]}}
```
davinotdavid: Not sure if this was intended but we are showing the raw enum value here, was this for debugging purposes?

image

frontend/src/stores/user-store.ts
```diff
@@ -290,27 +295,24 @@ export const useUserStore = defineStore('user', () => {
...
-    return profile();
+    return await profile();
   };
 
   /**
    * Do subscriber logout and reset store
    */
   const logout = async () => {
-    const { error }: BooleanResponse = await call.value('logout').get().json();
+    if (isOidcAuth) {
```
davinotdavid: Don't we need to call the `/logout` endpoint for other types of auth / are we phasing out all other types already? If so we could drop the conditional altogether or maybe we need some other changes in the backend here https://github.com/thunderbird/appointment/blob/main/backend/src/appointment/routes/auth.py#L583-L591 ?

frontend/src/views/PostLoginView.vue
```diff
@@ -22,12 +23,25 @@ onMounted(async () => {
   // Remove any ftue steps on new login
   window.localStorage?.removeItem('tba/ftue');
 
-  if (!isFxaAuth.value && !isAccountsAuth) {
+  if (isOidcAuth) {
+    // Stored in an internal store
+    const userData = await userManager.signinCallback(window.location.href);
+    const { error, data }: BooleanResponse = await call('oidc/token').post({
+      'access_token': userData.access_token,
+      //'invite_code': inviteCode.value,
```
davinotdavid: Just confirming here, the `invite_code` shouldn't be passed in?

frontend/src/views/PostLoginView.vue
```diff
@@ -22,12 +23,25 @@ onMounted(async () => {
...
+      'timezone': dj.tz.guess(),
+    }).json();
+
+    if (error.value) {
+      console.error("WOAH", data.value);
```
davinotdavid: Love it but perhaps we should do a Sentry call instead?

backend/src/appointment/routes/auth.py
MelissaAutumn: It works as is for now. I'll create a ticket so we can decide how we want to handle deleted users.

frontend/src/components/SettingsConnections.vue
MelissaAutumn: Should be fixed now!

frontend/src/stores/user-store.ts
MelissaAutumn: The auth here is a little different. Previously the login flow was handled by the backend. We only had one oauth client, and we made a token for the frontend to auth with.

Now we have two clients. A frontend one that is set to public (only needs a client id and pkce) which does the login flow, and a backend client that does have a client id and client secret. The backend uses a token introspection route to verify the token is legit and not expired. (we cache this request for about 5 minutes iirc)

So theres nothing for the backend to logout of, since the frontend handles it.

frontend/src/views/PostLoginView.vue
MelissaAutumn: fixed!

frontend/src/views/PostLoginView.vue
MelissaAutumn: Whoops, should be fixed now. Sentry should get error logs automagically.

frontend/src/stores/user-store.ts
davinotdavid: Hmmm I see, thanks for the explanation! I also see that the conditional was flipped now, does that mean that we currently still supporting other methods of auth or is this just a replacement for the previous `accounts` method (now deprecated)?

(Just gauging if / when we can delete more code that is unused)

MelissaAutumn: We still support fxa, and regular password auth for now. 

I'll rip accounts out in another ticket.

docker-compose.yml
devmount: Good thinking!

frontend/src/composables/authSchemes.ts
```diff
@@ -0,0 +1,5 @@
+import {AuthSchemes} from "@/definitions";
```
devmount, line 1:
```suggestion
import { AuthSchemes } from "@/definitions";
```

devmount (approved):
A very thorough implementation, thanks a lot, Mel! Just a nit below. I can confirm, that the local account auth still works flawlessly.

Edit: And I added German translation. So should be ready to be merged in!

Issue #1097, opened 2025-07-08
Date picker/Booker page overhaul


malini:
From the updated UI wireframes:
<img width="665" height="553" alt="Image" src="https://github.com/user-attachments/assets/121d86ee-ce4e-492b-9545-388c693ce897" />

Add a new section on the left summarizing the selection and add a panel for the booker's name and email. I'm okay with punting on the Sign In section, as we have no features that would be enabled by being logged in as a booker, so it's not useful just yet.

malini:
oh and the "Wellness session" section can be removed I think as well, as we don't have a place to set that information yet, right @laurelterlesky ?

laurelterlesky:
We have this already in beta - although the field doesn't have a label. It's the input directly below the 'Set General availability' toggle. 

In this MVP wireframe, Wellness session = Booking Page name. 
The Descriptions is new for MVP. 

See the Availability wireframe for how this informs the booker's view: https://zpl.io/XKKdjvp

@Jesse-Moz please as any more detail if needed

laurelterlesky:
I should add that the requirement of a Booking Page name sets us up for having multiple schedules in the future

MelissaAutumn:
This should re-use the same week-view component we create for https://github.com/thunderbird/appointment/issues/1093 so I'm adding a blocked tag until #1093 is resolved.

devmount:
Just FYI: I'm currently working on extending the WeekCalendar component for the booker page.

davinotdavid:
@devmount as you are wrapping up this ticket, I wonder if you could also add a `data-ref` or `data-testid` for the E2E tests for the available slots as @rwood-moz pointed out that our version of the calendar does not have it anymore (I believe this was some sort of default behaviour from Qalendar as I couldn't find a reference to it even going back months ago).

In the code, I believe it would be [here](https://github.com/thunderbird/appointment/blob/main/frontend/src/views/DashboardView/components/WeekCalendar.vue#L413-L426) but good to double check just in case!

This is the format that was expected in the e2e: `event-2025-01-08 09:30`.

devmount:
@davinotdavid Of course, will do! Thanks for the reminder!

devmount:
@rwood-moz @davinotdavid Let me know, if the test ids in #1273 are sufficient! I happily add more or change the format if needed.

rwood-moz:
[...]

PR #1106, opened 2025-07-10
Bookings page UI overhaul - Part 1 (Table, Filters, Unconfirmed First)

davinotdavid:
## Description of the Change

- Renamed and moved `views/AppointmentsView.vue` to `views/BookingsView/` folder
  - Name change to be on par with the route `/bookings`
  - `views/BookingsView/index.vue` is the main page
  - `views/BookingsView/components/*` host the components that are only used in this route

- Removed the existing appointment filters in favor of a new `AppointmentMultiSelectFilter` dropdown component
- Added a checkbox for "Unconfirmed first"

Light Mode

https://github.com/user-attachments/assets/411e257e-6360-40b7-8679-9a8ddafef218

Dark Mode

https://github.com/user-attachments/assets/04aed060-4169-4ef6-8080-aaae78becc42

- Added a sortable table to show the appointments

> [!NOTE]  
> Since this PR was already getting too large, I'll create another one for the infinite scroll on the frontend (need to add pagination metadata in the backend + update several other files) and another PR to support the `booking_status == cancelled` and `booking_status == declined`. All table column sorts are happening on the frontend only.

PS: I understand the motivation of calling the model `appointment` based on the underlying data / backend but for frontend purposes, perhaps we should rename it all to `booking` instead?

## Benefits

- UI / UX better aligned with Design team wireframes

## Applicable Issues

https://github.com/thunderbird/appointment/pull/1102


frontend/src/assets/styles/main.css
```diff
@@ -205,6 +205,18 @@
...
+  position: absolute;
+  width: 1px;
+  height: 1px;
+  padding: 0;
+  margin: -1px;
+  overflow: hidden;
+  clip: rect(0, 0, 0, 0);
+  white-space: nowrap;
+  border-width: 0;
+}
```
davinotdavid, line 218: This is the same as Tailwind's `sr-only` class but I am copying it here in main.css since we will move away from it eventually.

frontend/src/router.ts
```diff
@@ -112,9 +112,9 @@ const routes: RouteRecordRaw[] = [
     redirect: { name: 'bookings' },
   },
   {
-    path: '/bookings/:view?/:slug?',
+    path: '/bookings/:slug?',
```
davinotdavid, line 115: Filters synced with query params will be coming in a follow up PR

frontend/src/views/BookingsView/index.vue
```diff
@@ -0,0 +1,443 @@
...
+const sortDirection = ref<'asc' | 'desc'>('asc');
+
+// Handle data view
+const view = ref(BookingsViewTypes.List);
+
+// handle appointment sliding panel
+const appointmentSlidingPanelRef = ref<InstanceType<typeof AppointmentSlidingPanel>>()
+
+// Handle filtered appointments list
+const filteredAppointments = computed(() => {
```
davinotdavid, line 59: Sorts & filters are happening in-memory / frontend only for now, this will need to call the backend since we want infinite scrolling (also will be done in a follow up PR)

frontend/src/locales/de.json
```diff
@@ -205,7 +206,9 @@
     "closed": "Geschlossen",
     "code": "Code",
     "color": "Farbe",
+    "columnHeadersSortable": "Spaltenüberschriften mit Schaltflächen sind sortierbar",
```
```suggestion
    "columnHeadersSortable": "Spaltenüberschriften mit Buttons sind sortierbar",
```

This would be more common.

frontend/src/router.ts
```diff
@@ -10,8 +10,8 @@ import { useUserStore } from '@/stores/user-store';
 import LogoutView from '@/views/LogoutView.vue';
 
 // lazy loaded components
+const BookingsView = defineAsyncComponent(() => import('@/views/BookingsView/index.vue'));
```
devmount, line 13: We now have a `BookingsView` and a `BookingView` component, which might cause confusion. Should we rename the  `BookingView` to something like `AppointmentRequestView` or similar?

frontend/src/views/BookingsView/components/AppointmentMultiSelectFilter.vue
```diff
@@ -0,0 +1,209 @@
...
+import { ref, computed, onMounted } from 'vue';
+import { vOnClickOutside } from '@vueuse/components';
+import { useI18n } from 'vue-i18n';
+import { IconChevronDown } from '@tabler/icons-vue';
+import CheckboxInput from '@/tbpro/elements/CheckboxInput.vue';
+
+interface FilterOption {
+  value: string | number;
+  label: string;
+}
```
devmount: We already have a `SelectOption` type iirc

frontend/src/views/BookingsView/components/AppointmentMultiSelectFilter.vue
```diff
@@ -0,0 +1,209 @@
...
+    <!-- Pill trigger -->
+    <button @click="toggleDropdown" class="filter-trigger" type="button">
+      <span>{{ displayText }}</span>
+      <IconChevronDown class="chevron-icon" :class="{ 'rotated': isOpen }" />
+    </button>
+
+    <!-- Dropdown -->
+    <transition enter-active-class="dropdown-enter-active" enter-from-class="dropdown-enter-from"
+      enter-to-class="dropdown-enter-to" leave-active-class="dropdown-leave-active"
+      leave-from-class="dropdown-leave-from" leave-to-class="dropdown-leave-to">
```
devmount: Can we use a named transition / transition component here for reusability?

frontend/src/views/BookingsView/index.vue
```diff
@@ -0,0 +1,443 @@
...
+    showAppointmentSlidingPanel(appointments.value.filter((appointment) => appointment.slug === route.params.slug)[0]);
+  }
+});
+</script>
+
+<script lang="ts">
+export default {
+  name: 'BookingsView'
+}
+</script>
```
devmount, line 167: Interesting, why can't this be part of the setup script block?

frontend/src/views/BookingsView/index.vue
```diff
@@ -0,0 +1,443 @@
...
+        <caption>
+          <span class="sr-only">
+            {{ t("label.columnHeadersSortable") }}
+          </span>
+        </caption>
+        <thead>
+          <tr class="table-header-row">
+            <th
+              class="table-header"
+              :aria-sort="sortColumn === 'date' ? (sortDirection === 'asc' ? 'ascending' : 'descending') : null"
```
devmount: There are a few of those aria-sort attributes. Can we make this a method like `ariaSortForColumn(column)`?

devmount (approved):
No blockers, just some suggestions we can discuss. Thank you!

frontend/src/views/BookingsView/index.vue
davinotdavid: Good question! The `<script setup>` bit is a syntax sugar for the `setup()` function in a Vue component so I can't export an object from within it [[ref](https://vuejs.org/api/sfc-script-setup.html#basic-syntax)].

So we could theoretically do the other way around which is putting everything from the `<script setup>` inside the `<script>` and [export a function `setup()`](https://vuejs.org/api/composition-api-setup.html#basic-usage) within it with all the logic (removing the syntax sugar) but I thought it would be cleaner not to. Happy to change it if you think it would be clearer of course!

frontend/src/locales/de.json
davinotdavid: Thank you! 🙏

frontend/src/views/BookingsView/index.vue
devmount: No no this is totally fine! And makes sense, thanks for the explanation!

frontend/src/router.ts
davinotdavid: 100% agree, that gets me confused to be honest and I like the name of `AppointmentRequestView`! There's quite a bit of changes coming in soon-ish-hopefully in this ticket https://github.com/thunderbird/appointment/issues/1097 so perhaps it would be nice to change it then. What do you think?

devmount: Yes, that would be awesome!

frontend/src/views/BookingsView/index.vue
davinotdavid: Absolutely! Changed here, thanks for the suggestion!

frontend/src/views/BookingsView/components/AppointmentMultiSelectFilter.vue
davinotdavid: For sure, created a new transition component in the shared component folder. That also makes me think that transitions could be a good thing to add in `services-ui` as well in the future! Thanks!

frontend/src/views/BookingsView/components/AppointmentMultiSelectFilter.vue
davinotdavid: Oh you are right! Reused the `SelectOption` from @models, thanks!

frontend/src/views/BookingsView/components/AppointmentMultiSelectFilter.vue
MelissaAutumn: I think the plan right now is to keep things reusable (so yes component!) and if design wants it in another services then we move it to services-ui. 

They're trying to keep the core components used within their design framework pretty slim.

frontend/src/assets/styles/main.css
MelissaAutumn: can we rename this to screenreader-only? (or screen-reader-only?) 

part of the reason I dislike tailwind is all the abbreviations 😅.

frontend/src/components/FadeInOutTransition.vue
```diff
@@ -0,0 +1,44 @@
+<template>
```
MelissaAutumn, line 1: noice!

frontend/src/router.ts
MelissaAutumn: please do!

frontend/src/views/BookingsView/index.vue
MelissaAutumn: I can't remember, is infinite scroll in the design, or just inferred from the design?

frontend/src/views/BookingsView/index.vue
```diff
@@ -0,0 +1,449 @@
...
+  }
+
+  return list;
+});
+
+// Handle column sorting
+const handleColumnSort = (column: TableColumn) => {
+  if (sortColumn.value === column) {
+    // Toggle direction if same column
+    sortDirection.value = sortDirection.value === 'asc' ? 'desc' : 'asc';
```
MelissaAutumn: Can we enum asc and desc?

MelissaAutumn (commented):
Initial review, looks good though! I have to test it after a meeting.

frontend/src/views/BookingsView/index.vue
davinotdavid: It is in the design on a behavior note saying "_Scrollable area, with lazy loading for large numbers of elements_" which I assumed to be infinite scrolling

frontend/src/views/BookingsView/index.vue
davinotdavid: Sure! Enum'ed, thanks!

frontend/src/assets/styles/main.css
davinotdavid: Fair point, changed to `screen-reader-only`!

MelissaAutumn:
FYI I've rebased this PR. You'll need to pull before pushing again.

MelissaAutumn (changes requested):
I found the dropdown filter doesn't actually work with the "unconfirmed first" checbox. Can you double check this? 

I also found a checkboxinput problem with screen readers but that's a services-ui problem 😄

davinotdavid:
@MelissaAutumn great catch! omg the logic was quite flawed, the unconfirmed shouldn't be done first, it should be done last... sorry about that :( Pushed a change!

https://github.com/user-attachments/assets/ac7f11af-8a74-40d4-811b-e990b698c019

PR #1113, opened 2025-07-15
Fix backend's base URL paths for Mailpit & Fluent

davinotdavid:
## Description of the Change

Although the backend service works well, when mail templates are referenced, they fail to resolve correctly as the copy from the Dockerfile says `COPY src .` and the expected path for Mailpit Jinja templates is `src/appointment/templates/email`.

We could either do the change back from the Dockerfile _or_ change the [`get_jinja()` function from the mailer](https://github.com/thunderbird/appointment/blob/main/backend/src/appointment/controller/mailer.py#L23) and [`get_fluent()` function from the I10n](https://github.com/thunderbird/appointment/blob/main/backend/src/appointment/middleware/l10n.py#L15) and potentially other places that I might be missing.

~~For brevity, changing the Dockerfile back makes it work it again locally but unsure if this breaks other parts of the infra.~~

As @MelissaAutumn suggested, let's go with changing the base paths to be `sys.modules["appointment"].__path__[0]`

## Benefits

Emails & I18n work again!

## Applicable Issues

https://github.com/thunderbird/appointment/pull/1079

### DIFF
--- a/backend/src/appointment/controller/mailer.py
+++ b/backend/src/appointment/controller/mailer.py
@@ -7,6 +7,7 @@ import logging
 import os
 import smtplib
 import ssl
+import sys
 from email.message import EmailMessage
 
 import jinja2
@@ -18,9 +19,11 @@ from fastapi.templating import Jinja2Templates
 
 from ..l10n import l10n
 
+# Resolves to absolute appointment package path
+BASE_PATH = f'{sys.modules["appointment"].__path__[0]}'
 
 def get_jinja():
-    path = 'src/appointment/templates/email'
+    path = f'{BASE_PATH}/templates/email'
 
     templates = Jinja2Templates(path)
     # Add our l10n function
@@ -80,7 +83,7 @@ class Mailer:
 
     def _attachments(self):
         """provide all attachments as list, add tbpro logo to every mail"""
-        with open('src/appointment/templates/assets/img/tbpro_logo.png', 'rb') as fh:
+        with open(f'{BASE_PATH}/templates/assets/img/tbpro_logo.png', 'rb') as fh:
             tbpro_logo = fh.read()
 
         return [
@@ -207,7 +210,7 @@ class BaseBookingMail(Mailer):
 
     def _attachments(self):
         """We need these little icons for the message body"""
-        path = 'src/appointment/templates/assets/img/icons'
+        path = f'{BASE_PATH}/templates/assets/img/icons'
 
         with open(f'{path}/calendar.png', 'rb') as fh:
             calendar_icon = fh.read()
--- a/backend/src/appointment/controller/mailer.py
+++ b/backend/src/appointment/controller/mailer.py
@@ -7,7 +7,6 @@ import logging
 import os
 import smtplib
 import ssl
-import sys
 from email.message import EmailMessage
 
 import jinja2
@@ -17,13 +16,12 @@ import validators
 from html import escape
 from fastapi.templating import Jinja2Templates
 
+from ..defines import BASE_PATH
 from ..l10n import l10n
 
-# Resolves to absolute appointment package path
-BASE_PATH = f'{sys.modules["appointment"].__path__[0]}'
 
 def get_jinja():
-    path = f'{BASE_PATH}/templates/email'
+    path = os.path.join(BASE_PATH, 'templates/email')
 
     templates = Jinja2Templates(path)
     # Add our l10n function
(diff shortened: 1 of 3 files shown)


backend/Dockerfile
```diff
@@ -20,11 +20,14 @@ COPY scripts/cron /etc/cron.d/appointment-cron
 RUN chmod 0644 /etc/cron.d/appointment-cron
 RUN crontab /etc/cron.d/appointment-cron
 
-COPY src .
-
 RUN pip install --upgrade pip
 RUN pip install .'[deploy]'
 
+# Add this hack to line it up with our dev environment.
+# I'll buy whoever fixes this a coffee.
```
davinotdavid: Also added the comment back here so that the coffee bounty remains

ryanjjung:
I haven't poked at this myself yet, but it looks like the problem is the code expects `src/` to be prepended on the path but because we do a `COPY src .` the `src/` part is gone?

davinotdavid:
@ryanjjung sort of! The `WORKDIR` for the Dockerfile is `/app` so all files are copied to `/app` but the code expects the paths to be `src/` so the symbolic link there was doing the heavy lifting to make it work.

Unsure if is necessary for other reasons but maybe we could switch the `WORKDIR` to be `/src` instead?

backend/Dockerfile
MelissaAutumn: ☕

ryanjjung:
Ideally, we want all of these things in alignment. The Docker container should be built as though there is no difference between local and remote deployment. The docker-compose should overlay the local files onto the container so the file watching/service reloading features work. If there's something that applies to local dev but not cloud environments, then it should be set in the docker-compose. Is this broken in stage/prod or only local dev?

MelissaAutumn:
Okay I think we can fix the reference issue if we use the magical and totally annoying `sys.modules['appointment'].__path__[0]`.

This retrieves a reference to the appointment module (which we have installed as an editable package in both docker and non-docker (for tests) and will resolve to that package's root dir. 

In docker it's `/app/appointment` and outside it's `'/Users/$user/Dev/appointment/backend/src/appointment'` (or where ever your path is)

davinotdavid:
@ryanjjung @MelissaAutumn wait, I am bit confused (as usual) 😅 

Would there be a problem to change the backend's `Dockerfile` to be:

```
RUN mkdir src
WORKDIR /src

# ... other parts remain the same

COPY src .
```

and then the `docker-compose.yml` to be:

```
backend:
    build:
      context: ./backend
      dockerfile: ./Dockerfile
    ports:
      - 5000:5000
    volumes:
      - ./backend/.env:/src/.env
      - ./backend/src/appointment:/src/appointment
    # ... other stuff
```

Then we could still use `src/` based paths all around as usual?

MelissaAutumn:
Ah okay I see. Yeah, I think it should just work. We might have to adjust the volume mounts to not include the src directory. But basically we install the appointment code as an editable package. This allows it to work via an open directory (like src/appointment), and the `sys.modules['appointment'].__path__[0]` code will tell us where that package's absolute root directory is. 

I think if we adjust the template and fluent directory code to use `sys.modules['appointment'].__path__[0]` as the base path we should be okay. Happy to help debug any issues with you that come up!

davinotdavid:
@MelissaAutumn @ryanjjung

Great, thanks for input folks! Let's go with the base url `sys.modules['appointment'].__path__[0]` changes and leave the Docker specific configs as is then.

I've updated the code and can verify that I am able to book an appointment and emails get sent / translations are done! Ready for review 🙏

backend/src/appointment/controller/mailer.py
```diff
@@ -18,9 +19,11 @@
 
 from ..l10n import l10n
 
+# Resolves to absolute appointment package path
+BASE_PATH = f'{sys.modules["appointment"].__path__[0]}'
```
MelissaAutumn: Can we move this to defines.py?

backend/src/appointment/controller/mailer.py
```diff
@@ -18,9 +19,11 @@
...
 
 def get_jinja():
-    path = 'src/appointment/templates/email'
+    path = f'{BASE_PATH}/templates/email'
```
MelissaAutumn: Can we switch all of these paths to `os.path.join`? Sorry 😅

MelissaAutumn (changes requested):
Some small things but looking good.

backend/src/appointment/controller/mailer.py
davinotdavid: Ah yes, definitely! Moved, thanks!

backend/src/appointment/controller/mailer.py
davinotdavid: For sure! That's a good point, no need for sorry haha changed!

MelissaAutumn (approved):

Issue #1111, opened 2025-07-15
Update E2E tests for latest UI changes


rwood-moz:
More specifically PR #1102 has made some UI changes that have broken the existing E2E tests; update the tests to work with these new UI changes.

Browserstack link of the failing E2E tests after the PR merge can be [found here](https://automate.browserstack.com/dashboard/v2/builds/2f53a0937b6bfe6703cecb591f011cced51397e8).

rwood-moz:
More specifically, the new test failures:

1. `calendar-settings => able to edit calendar`

After the addition of the user's email address being displayed beside the user avatar, the test is breaking as it finds the user email in two places which breaks a locator:

```
Error: expect.toBeVisible: Error: strict mode violation: getByText('<email>', { exact: true }) resolved to 2 elements:
    1) <span class="text-sm text-gray-500"><email></span> aka getByRole('navigation').getByText('<email>')
    2) <span class="calendar-title"><email></span> aka getByRole('main').getByText('<email>')
```

2. `connected-accounts-settings => disconnect calendar button`

The data-testid for the google account disconnect button is no longer found by the test, as it has been changed so it is dynamic: `:data-testid="'connected-accounts-settings-' + t(provider) + '-disconnect-btn-' + conn.type_id"`

PR #1114, opened 2025-07-15
Update E2E tests for latest UI changes

rwood-moz:
See #1111. This PR will fix the E2E tests that are currently broken on the stage env due to new UI changes. Note: The E2E tests will still fail on production until the next production deployment.

### DIFF
--- a/test/e2e/pages/settings-page.ts
+++ b/test/e2e/pages/settings-page.ts
@@ -95,7 +95,8 @@ export class SettingsPage {
     // calendar settings
     this.connectedCalendarsHeader = this.page.getByText('Connected Calendars', { exact: true });
     this.editCalendarBtn = this.page.getByTestId('settings-calendar-edit-calendar-btn');
-    this.connectedCalendarTitle = this.page.getByText(APPT_LOGIN_EMAIL, { exact: true });
+    this.connectedCalendarTitle = this.page.getByRole('main').getByText(APPT_LOGIN_EMAIL, { exact: true });
+
     this.editCalendarTitleInput = this.page.getByTestId('settings-calendar-title-input');
     this.editCalendarColorInput = this.page.getByTestId('settings-calendar-color-input');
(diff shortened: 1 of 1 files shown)


rwood-moz:
BrowserStack link of the updated E2E tests passing on stage can be [found here](https://automate.browserstack.com/dashboard/v2/builds/87282bfb8fe34ab61448806c9bc33a78c6692a7e).

test/e2e/pages/settings-page.ts
```diff
@@ -111,7 +112,11 @@ export class SettingsPage {
     this.editProfileBtn = this.page.getByRole('button', { name: 'Edit profile' });
     this.mozProfilePageLogo = this.page.getByTestId('logo');
     this.mozProfileSettingsSection = this.page.getByTestId('settings-profile');
-    this.disconnectGoogleCalendarBtn = this.page.getByTestId('connected-accounts-settings-google-disconnect-btn');
+
+    // getByTestId requires an exact match; our disconnect btn data-testid is now dynamic; our test account
+    // only has one google account connected so we can just use a wildcard to locate the disconnect btn
+    this.disconnectGoogleCalendarBtn = this.page.locator('[data-testid*="connected-accounts-settings-google-disconnect-btn"]');
```
davinotdavid, line 118: In the near future I think we can probably have more than 1 google account connected for e2e testing purposes so I wonder if this `Locator` would match both disconnect buttons? Or would that be something to be handled on the assertion that uses this Locator?

rwood-moz: Thanks for the review!

I'm not sure it will be necessary to have two google accounts connected for the E2E tests. The E2E tests are meant really to just test the main user flows and not be exhaustive, as that makes them more difficult to maintain. Having more than one google account connected should really be tested at the integration test levels instead. So I think this is sufficient for now to get the tests fixed. We could also revisit this in the future if we want. :)

PR #1119, opened 2025-07-16
Setup new infra with oidc settings

MelissaAutumn:
For some reason I'm hitting
```
TypeError: pulumi_aws.cloudfront.distribution.Distribution() got multiple values for keyword argument 'default_cache_behavior'
```

when trying to run pulumi up, so I can't actually push the secrets up to tie them in. So you might have to do that. 

But here's all the oidc settings, it should just work!


frontend/.env.pulumiprod.example
```diff
@@ -24,3 +24,6 @@ VITE_REPORT_BUG_URL=https://github.com/thunderbird/appointment/issues/new?assign
 VITE_DEFAULT_HOUR_FORMAT=12
 
 VITE_SUPPORT_URL=https://tbpro.zendesk.com/hc/requests/new
+
+VITE_OIDC_CLIENT_ID=thunderbird-appointment-frontend
+VITE_OIDC_ROOT_URL=https://auth.tb.pro/realms/tbpro/
```
MelissaAutumn, line 29: This client_id doesn't actually exist yet. But it's here for later.

ryanjjung:
I'll try the Pulumi here in a bit. Might be a tb_pulumi version mismatch.

frontend/.env.pulumiprod.example
```diff
@@ -8,10 +8,10 @@ VITE_SHORT_BASE_URL=apt.mt
 VITE_SENTRY_DSN=https://66f57328baac4929be7d62b77f41b510@o4505428107853824.ingest.sentry.io/4505428280279040
 
 # -- Auth scheme --
-VITE_AUTH_SCHEME=fxa
+VITE_AUTH_SCHEME=oidc
 
 # For fxa
-VITE_FXA_EDIT_PROFILE=https://accounts.firefox.com/settings
+VITE_FXA_EDIT_PROFILE=https://accounts.tb.pro/self-serve
```
ryanjjung, line 14: At some point, we should peer these VPCs and route this traffic over the internal network. I [filed an issue](https://github.com/thunderbird/appointment/issues/1120) about it.

ryanjjung (approved):

frontend/.env.pulumiprod.example
MelissaAutumn: in this case this var is used to redirect the user to accounts.

PR #1124, opened 2025-07-17
[WIP] Bookings page UI overhaul - Part 2 (Filters on URL + declined / cancelled BookingStatus)

davinotdavid:
## Description of the Change

- Added a `useQueryParamState` composable so that we can sync query params to a piece of reactive state. Now the dropdown filters are tracked on the URL alongside the `unconfirmed` checkbox as well.
- Added `cancelled` and `declined` to the `BookingStatus` enum both in the backend and in the frontend. In the frontend specifically, I had to create some conversions between the number <-> query param (which is not i18n intentionally as it is only in the URL).
- Appointments and slots are not deleted anymore, upon schedule request and showing availability we are now ignoring the slots that have the booking status mentioned above.

## Benefits

- Keeping the state as queryParams allow the user to do a page refresh and keep the table filtered
- Not deleting the appointment / slot allows us to have a tracking history of previously denied or cancelled appointments

## Applicable Issues

### DIFF
--- a/backend/src/appointment/database/repo/slot.py
+++ b/backend/src/appointment/database/repo/slot.py
@@ -56,13 +56,13 @@ def add_for_schedule(db: Session, slot: schemas.SlotBase, schedule_id: int):
 
 
 def exists_on_schedule(db: Session, slot: schemas.SlotBase, schedule_id: int):
-    """check if given slot already exists for schedule of given id"""
+    """check if given slot is already requested or booked for schedule of given id"""
     db_slot = (
         db.query(models.Slot)
         .filter(models.Slot.schedule_id == schedule_id)
         .filter(models.Slot.start == slot.start)
         .filter(models.Slot.duration == slot.duration)
-        .filter(models.Slot.booking_status != models.BookingStatus.none)
+        .filter(models.Slot.booking_status.in_([models.BookingStatus.requested, models.BookingStatus.booked]))
         .first()
     )
     return db_slot is not None
@@ -113,4 +113,22 @@ def delete(db: Session, slot_id: int):
     db_slot = get(db, slot_id)
     db.delete(db_slot)
     db.commit()
+    return db_slot
+
+
+def reject(db: Session, slot_id: int) -> models.Slot | None:
+    """update booking status for slot of given id as declined"""
+    db_slot = get(db, slot_id)
+    db_slot.booking_status = models.BookingStatus.declined
+    db.commit()
+    db.refresh(db_slot)
+    return db_slot
+
+
+def cancel(db: Session, slot_id: int) -> models.Slot | None:
+    """update booking status for slot of given id as cancelled"""
+    db_slot = get(db, slot_id)
+    db_slot.booking_status = models.BookingStatus.cancelled
+    db.commit()
+    db.refresh(db_slot)
     return db_slot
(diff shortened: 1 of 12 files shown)


backend/src/appointment/migrations/versions/2023_10_19_1535-2b1d96fb4058_extend_slots_table_for_.py
```diff
@@ -33,7 +32,7 @@ def upgrade() -> None:
         sa.Column('booking_tkn', StringEncryptedType(sa.String, secret, AesEngine, 'pkcs5', length=512), index=False),
     )
     op.add_column('slots', sa.Column('booking_expires_at', DateTime()))
-    op.add_column('slots', sa.Column('booking_status', sa.Enum(BookingStatus), default=BookingStatus.none))
+    op.add_column('slots', sa.Column('booking_status', sa.Enum('none', 'requested', 'booked'), default='none'))
```
davinotdavid, line 35: Since I am changing the `BookingStatus` enum itself, I believe we can't use it here anymore for setting it up from scratch unfortunately. Open to suggestions though?

backend/src/appointment/migrations/versions/2025_07_17_1924-b2f516cb40cc_update_booking_status_enum_.py
```diff
@@ -0,0 +1,30 @@
...
+revision = 'b2f516cb40cc'
+down_revision = 'fb27de54e731'
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.alter_column('slots', 'booking_status',
+               existing_type=sa.Enum('none', 'requested', 'booked'),
+               type_=sa.Enum('none', 'requested', 'booked', 'declined', 'cancelled'),
```
davinotdavid, line 22: Similar here as https://github.com/thunderbird/appointment/pull/1124/files#r2214445347

backend/src/appointment/migrations/versions/2023_10_19_1535-2b1d96fb4058_extend_slots_table_for_.py
devmount: Well, since this is only a migration file, this should be fine imo

backend/src/appointment/database/repo/slot.py
```diff
@@ -114,3 +114,21 @@ def delete(db: Session, slot_id: int):
...
+    return db_slot
+
+
+def cancel(db: Session, slot_id: int) -> models.Slot | None:
+    """update booking status for slot of given id as cancelled"""
+    db_slot = get(db, slot_id)
+    db_slot.booking_status = models.BookingStatus.cancelled
+    db.commit()
+    db.refresh(db_slot)
+    return db_slot
```
devmount, line 134: This is fine, I was just thinking if we want to unify those into a `set_status(db: Session, slot_id: int, status: models.BookingStatus)` function and call this instead of duplicating the save logic for each status.

frontend/src/composables/useQueryParamState.ts
devmount, line 1: That's convenient! Thanks for putting that into a reusable component!

devmount (approved):
Awesome stuff, no blockers found. Thanks!

backend/src/appointment/database/repo/slot.py
davinotdavid: Good point! Unifying would reduce the amount of code but would also allow for updating the booking status to _any_ booking status – for instance, someone could potentially use the `.update()` to change a slot from `cancelled` to `declined` at will which shouldn't be possible. 

Someone could still call `reject` and `cancel` to forcibly change these status in the current code but one idea of the extra validations for business logic here could be adding checks to the `reject` and `cancel` methods to further tight the business logic.

Open to both though! What do you think?

devmount: Valid point. Although normally we used to keep the repository functions rather atomic and put the business logic in the controllers. So any misuse of a possible set_status function should be handled/validated before calling it.

But I'm also open for both!

davinotdavid: Makes sense! Since the code is fine for now, I will do the refactor of the booking status methods to create the `update` method and change the usage in the Part 3 PR that I am working on 🙇

PR #1130, opened 2025-07-21
Transfer Pulumi IaC into release env

ryanjjung:
The deploy steps are failing because we are not bringing the Pulumi code into the workflow environment we're running the deploy out of.

### DIFF
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -8,7 +8,6 @@ on:
   push:
     branches:
       - main
-      - fix-releases
 
 permissions:
   id-token: write # required for OIDC connectiong to AWS
@@ -33,7 +32,6 @@ jobs:
               - 'tofu/modules/**'
               - 'tofu/environments/stage/**'
               - '.github/workflows/deploy-staging.yml'
-              # TODO: Remove prod listings from here
               - '.github/workflows/deploy-production.yml'
             deploy-backend:
               - 'backend/**'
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-staging.yml
```diff
@@ -33,7 +32,6 @@ jobs:
               - 'tofu/modules/**'
               - 'tofu/environments/stage/**'
               - '.github/workflows/deploy-staging.yml'
-              # TODO: Remove prod listings from here
```
ryanjjung, line 36: On second thought, I think we should leave them. The only way to trigger the prod deploy workflow so that it uses new changes to the prod deploy workflow is to actually merge such a change into main and then complete a release. So if we need to change the prod deploys, we would need all of these jobs to execute to produce that draft release anyway.

.github/workflows/deploy-staging.yml
```diff
@@ -528,6 +526,16 @@ jobs:
...
         run: zip -r iac.zip tofu -x "tofu/environments/stage/*" "tofu/environments/prod/*/*/.terragrunt-cache/*"
 
+      - name: Zip Pulumi
+        run: |
+          # Remove any lingering build pieces that may exist
+          pushd pulumi
+          rm -rf appointment_pulumi.egg-info build __pycache__ .ruff_cache venv
+          popd
+
+          # Tarbizz the Pulumi IaC
```
ryanjjung, line 536: Use tar.bz so we can feel superior and stuff.

MelissaAutumn (approved):

PR #1135, opened 2025-07-22
Job deps

ryanjjung:
The thinking here is that the production deployment steps require a whole bunch of artifacts to be present in order to succeed. The stage deployment steps end (when they are successful) with a `create-release` step that gathers the artifacts necessary for a prod release to succeed. Because we need all of these artifacts to be built, we cannot do itemized builds on merge where only certain artifacts get prepared.

So this PR removes the entire `detect-changes` step and every dependency other steps have on it. In this way, the only thing that would prevent a release is if something went wrong during this merge process.

### DIFF
--- a/.github/workflows/deploy-production.yml
+++ b/.github/workflows/deploy-production.yml
@@ -9,7 +9,7 @@ on:
     types: [published]
 
 permissions:
-  id-token: write # required for OIDC connectiong to AWS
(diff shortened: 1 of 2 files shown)


davinotdavid (approved):
🚀

MelissaAutumn (approved):
lgtm!

PR #1136, opened 2025-07-22
Don't run backend deploy on failure/noncompletion

ryanjjung:
This prevents a case where we try to do a backend image deployment before we have completed a backend image build.


ryanjjung:
> Do we need this on every subsequent job or just ones that "need" the output?

The other jobs have their own "if" clauses on them. In particular, the `create-release` job has a complex condition where it checks for success of any jobs listed as "needs". With that model, just listing the dependency is enough to require success to proceed. I checked the other jobs before submitting this PR, and they have good conditions set up on them.

PR #1125, opened 2025-07-21
Switch to use Services-UI components

devmount:
## Description of the Change

This change replaces all local tbpro components with their services-ui counterparts. We still use Buttons and other components from earlier designs, those will be replaced in later PRs.

## Benefits

Centralized management of design elements.

## Applicable Issues

Closes #862

Issue #1094, opened 2025-07-08
Availability page overhaul


malini:
From the updated UI wireframes:

<img width="705" height="741" alt="Image" src="https://github.com/user-attachments/assets/a52efd92-86fc-4ba4-b181-4a9fab36cdaf" />

* Move current Availability settings panel from Dashboard to new Availability view / route, keep the same functionality as is for now
* Split Availability details in Availability, Booking Page Details and Booking Page Link sections, keep the same functionality as is for now
* Update Set Your Availability section to match design changes
* Update Booking Page Details section to match design changes
* Update Booking Page Link section to match design changes

again, thanks Davi for breaking up this task!

MelissaAutumn:
Page should be a 50:50 split between components. We can collapse the flex div to columns for [small breakpoints](https://bolt.thunderbird.net/8b179dbfd/p/3599eb-layout/b/2821c0)

davinotdavid:
PRs have been merged and code is up on staging! Closing this but feel free to re-open if needed.

Issue #644, opened 2024-09-10
Add the ability to request a different meeting duration


malini:
**Is your feature request related to a problem? Please describe.**
Your calendar is set up to accept bookings of a fixed duration, and that one duration applies to anyone booking you. It would be nice to give the user who is booking you an option to request a different duration. For example, if Melissa gives me her booking url to book her for an hour, but I only need a half hour, I'd like there to be some way that apmt.day handles that situation

**Describe the solution you'd like**
Good question! I'm not sure what it should look like.

Perhaps an input field like this, but label it "Propose a different duration" or the like:
![Screenshot 2024-09-10 at 4 29 28 PM](https://github.com/user-attachments/assets/31630dac-885a-42d7-b062-06e810ad3edd)

On the booking confirmation page?
![Screenshot 2024-09-10 at 4 25 35 PM](https://github.com/user-attachments/assets/dde58d00-95c9-419e-ba35-cc92cc44e0c5)

PR #1138, opened 2025-07-22
Bookings page UI overhaul - Part 3 (SlidingPanel steps + hidden Modify Appointment feature)

davinotdavid:
### Frontend
- `AppointmentSlidingPanel` is the core change here. It has the logic to render the different steps that the panel can have through its button interactions and it hosts the `title` and `cta` slot coming from the `SlidingPanel` component. The initial state is being rendered by the `AppointmentSlidingPanelDetails`:
  - When the appointment is not yet confirmed, the user can select `Confirm` or `Deny`:
    - Either button moves the step to `APPOINTMENT_SLIDING_PANEL_STEPS.CONFIRMATION`, rendering the `AppointmentSlidingPanelBookingConfirmation.vue` component and passes a boolean prop to indicate the choice. Like the previous `BookingConfirmation.vue` component, it makes the API call and shows the appropriate result during the `onMounted`.
  - When the appointment is confirmed, the user can select `Modify booking`:
    - The `Modify booking` button moves the step to `APPOINTMENT_SLIDING_PANEL_STEPS.MODIFY`, rendering the `AppointmentSlidingPanelModify.vue` component which has the buttons `Cancel booking` and `Save`:
      - The `Cancel booking` moves the step to `APPOINTMENT_SLIDING_PANEL_STEPS.MODIFY`, rendering the `AppointmentSlidingPanelCancel.vue` component which shows a confirmation prompt to cancel the appointment. 
- `BookingConfirmation.vue` and its route have been deleted since it is now being rendered within the `AppointmentSlidingPanel` as one of the steps 
- Added `Modified` filter to the Bookings table

### Backend
- Added migration to include `modified` in the `BookingStatus` enum
- Added a `GET /schedule/availability_for_day?date=YYYY-MM-DD` endpoint so that we can retrieve slot availability for a specific day to populate a booking slot dropdown in the frontend
- Added a `PUT /apmt/{id}/modify` that can modify the appointment title and the slot start. It also changes the `BookingStatus` to `modified`
- Removed cancel reason since it is not present in the design anymore
- Mini-refactor on the `repo.slot.update` method as it was being used in a single place as a way to create and add attendees to a slot. This has now been renamed to `repo.slot.add_attendee_to_slot`
- Changed `repo.slot.update` to accept a partial `SlotUpdate` object to replace methods that had a single purpose (e.g. cancel only changed the booking status to cancel, reject only changed the booking status to reject but both were duplicated methods). Based off of [this comment thread](https://github.com/thunderbird/appointment/pull/1124#discussion_r2219766656).

## Benefits

- New feature of allowing a booking to be modifiable

## TODOs

- [TODO] Upon appointment modification, we need to send an email to the booker for confirmation with links / buttons in the email to confirm or cancel. The email flow itself has not been implemented yet.
- [TODO] Final design of each step is pending and it is currently uh... not great 😅 

## Applicable Issues


rwood-moz (changes requested):
Wow, cool PR! When attempting to run the E2E tests against this branch I found a few issues:

- The `share my quick link` button `data-testid` no longer exists and breaks the E2E tests as they can't find the button
- Similar issue with the profile save changes button
- Same issue with the account settings refresh link button
- Same issue with account settings disconnect google calendar button
- Same issue with calendar settings edit calendar button

I guess because these buttons are now `tbpro-button` class which doesn't have a `data-testid` value.

Is it possible to add a `data-testid` to the `tbpro-button` class with a name that matches the button label? Then the E2E test locators can be updated to match the new data-testid values. If not the E2E test button locators can be changed to something like this:

`this.shareMyLink = this.page.getByRole('button', { name: 'quick link'});`

rwood-moz:
Regarding E2E test locators (my previous comment) more specifically here are the locators that are broken and should be updated:

Account settings profile save changes button:
https://github.com/thunderbird/appointment/blob/main/test/e2e/pages/settings-page.ts#L91

Account settings refresh link button:
https://github.com/thunderbird/appointment/blob/main/test/e2e/pages/settings-page.ts#L92

Connected accounts settings google disconnect button:
https://github.com/thunderbird/appointment/blob/main/test/e2e/pages/settings-page.ts#L118

Connected accounts settings caldav connect button:
https://github.com/thunderbird/appointment/blob/main/test/e2e/pages/settings-page.ts#L121

Dashboard page availability edit link button:
https://github.com/thunderbird/appointment/blob/main/test/e2e/pages/dashboard-page.ts#L59

Calendar settings edit calendar button:
https://github.com/thunderbird/appointment/blob/main/test/e2e/pages/settings-page.ts#L97

With the above locators updated the E2E tests should then pass on this branch. Hope that helps :)

rwood-moz:
This might be a design question, but to me the `cancel booking` link is kind of small and doesn't really match the other buttons:

Screenshot 2025-07-24 at 3 05 43 PM

And on the modify dialog:
Screenshot 2025-07-24 at 3 05 30 PM

Should the `cancel booking` link be a similar (red) button maybe? Just my 2 cents haha

davinotdavid:
@rwood-moz 

About the E2E, thanks for letting me know! To be honest, I completely forgot to run them 😅 also, it seems that in `main` the tests are also not passing... I suspect it is related to the recent migration from the `tbpro` components to `services-ui` but I am still investigating (if that's the case, I might upgrade the `services-ui` repo and bump its package version here).

image

About the `cancel booking`, it is a bit odd indeed but I tried to follow the wireframes that we got so far (reference above). There is a "north star" UI design still under development that will hopefully polish all the visuals! Also note that we don't have yet the double booking capabilities nor the "busy" details block.

rwood-moz:
Ok thanks @davinotdavid! One small thing I do notice, the wireframe has upper case `Cancel Booking` for the link/button but the PR link has lowercase `booking` in the same link; just a tiny nit :)

devmount:
@rwood-moz @davinotdavid Surprisingly there were no fails on #1125 though, or do we run e2e separately?

backend/src/appointment/routes/schedule.py
```diff
@@ -128,6 +128,56 @@ def read_schedules(db: Session = Depends(get_db), subscriber: Subscriber = Depen
...
+
+    for schedule in active_schedules:
+        # calculate theoretically possible slots from schedule config for the given day
+        available_slots = Tools.available_slots_from_schedule(schedule, day)
+        all_available_slots.extend(available_slots)
+
+    actual_slots = Tools.events_roll_up_difference(all_available_slots, existing_slots)
+
+    if not actual_slots or len(actual_slots) == 0:
+        return []
```
MelissaAutumn: I think we can just do: 
`return actual_slots or []`

backend/src/appointment/routes/schedule.py
```diff
@@ -128,6 +128,56 @@ def read_schedules(db: Session = Depends(get_db), subscriber: Subscriber = Depen
...
+
+    calendars = repo.calendar.get_by_subscriber(db, subscriber.id, False)
+
+    if not calendars or len(calendars) == 0:
+        raise validation.CalendarNotFoundException()
+
+    try:
+        day = datetime.strptime(date, '%Y-%m-%d')
+    except ValueError:
+        raise validation.ValidationException(f"Invalid date format for: {date}")
```
MelissaAutumn: This doesn't seem to exist 🤔

backend/src/appointment/routes/schedule.py
```diff
@@ -128,6 +128,56 @@ def read_schedules(db: Session = Depends(get_db), subscriber: Subscriber = Depen
     return repo.schedule.get_by_subscriber(db, subscriber_id=subscriber.id)
 
 
+@router.get('/availability_for_day', response_model=list[schemas.SlotBase])
+def read_schedules_availability_for_day(
```
MelissaAutumn: This is a nice function but we should probably just grab the month on the frontend store it for this view and slice it up into days. This would remove the individual calls to the freebusy api.

frontend/src/router.ts
```diff
@@ -82,15 +81,6 @@ const routes: RouteRecordRaw[] = [
...
   },
-  {
-    path: '/user/:username/:signature/confirm/:slot/:token/:confirmed',
-    name: 'confirmation',
-    component: BookingConfirmationView,
-    meta: {
-      isPublic: true,
-      maskForMetrics: true,
-    },
-  },
```
MelissaAutumn, line 93: Unused?

davinotdavid:
@devmount ah, that's because E2E on CI is turned off on PRs but it should have ran when it was merged to `main`. After checking the logs, [it seems that it is also turned off for main](https://github.com/thunderbird/appointment/actions/runs/16478187074/job/46585663200).

It seems that [they were not working and they were slow](https://github.com/thunderbird/appointment/pull/1123/files#r2216905449) so there's a potential investigation going on about it.

malini:
regarding modifications, let's punt on this for public beta as it's not adequately scoped. I believe it was based off https://github.com/thunderbird/appointment/issues/644, which is a "request a different duration" vs "request 1 block vs 2 blocks of time" feature.

In any case, we should more thoroughly scope out how this should act and what happens after the request is made (ex: the email flow, and actions in that email, do modification requests expire, etc.).

MelissaAutumn:
> @devmount ah, that's because E2E on CI is turned off on PRs but it should have ran when it was merged to `main`. After checking the logs, [it seems that it is also turned off for main](https://github.com/thunderbird/appointment/actions/runs/16478187074/job/46585663200).
> 
> It seems that [they were not working and they were slow](https://github.com/thunderbird/appointment/pull/1123/files#r2216905449) so there's a potential investigation going on about it.

Ah sorry, I'll make a ticket to re-enable them.

Edit: Ticket https://github.com/thunderbird/appointment/issues/1142

frontend/src/router.ts
davinotdavid: Yep! This became `AppointmentSlidingPanelBookingConfirmation` and instead of doing a route navigation, it is now prop-based instead of url-based (still making the API request during `onMounted` though)

backend/src/appointment/routes/schedule.py
davinotdavid: Ah good point but just to make sure if I understand the solution here, do you mean that we should instead have a `GET /availability_for_month?month=` or the frontend should just call the existing `POST /public/availability`?

EDIT: Removed the whole method for now since we are hiding the modify feature on the frontend anyways

backend/src/appointment/routes/schedule.py
davinotdavid: Whoops, reminiscent of a previous code trial. Will be removed alongside this whole method

davinotdavid:
@rwood-moz after updating the `services-ui` package (https://github.com/thunderbird/services-ui/pull/85), I can verify that I can run the E2E tests locally 🎉  could you re-check on your end just in case?

rwood-moz:
> @rwood-moz after updating the `services-ui` package ([thunderbird/services-ui#85](https://github.com/thunderbird/services-ui/pull/85)), I can verify that I can run the E2E tests locally 🎉 could you re-check on your end just in case?

Yep, I pulled your latest branch and ran the E2E tests and they all passed now. Awesome, thanks @davinotdavid.

davinotdavid:
Removed `read_schedules_availability_for_day` method without replacement for now since we are not enabling the modify appointment feature anyways on the frontend.

Probably https://github.com/thunderbird/appointment/issues/1139 will need to be done first before enabling modify.

backend/src/appointment/routes/schedule.py
MelissaAutumn: Frontend just call the existing route, cache the result in a store, and then slice that data into the exact day you need.

This removes the 400ms or so delay on clicking each day, and switching to another month only requires 1 request.

backend/src/appointment/routes/api.py
```diff
@@ -393,6 +393,29 @@ def public_appointment_serve_ics(slug: str, slot_id: int, db: Session = Depends(
...
+        raise validation.AppointmentNotAuthorizedException()
+
+    modified_appointment = repo.appointment.update_title_and_slot(
+        db=db, appointment_id=id, appointment_data=appointment_data
+    )
+
+    # TODO: Send email to booker with link to confirm or cancel
+    #       use appointment_data.notes as the reason for the modification
+
+    return modified_appointment
```
MelissaAutumn, line 421: If we're not going to remove this function can you put a raise a 403 or 404 to prevent someone with a token from calling it?

frontend/src/views/BookingsView/components/AppointmentSlidingPanelModify.vue
```diff
@@ -0,0 +1,363 @@
+<script setup lang="ts">
```
MelissaAutumn, line 5: Can you add a comment mentioning it's currently unused and then the future ticket number?

rwood-moz:
Hey @davinotdavid one issue that I noticed while checking out your PR - however note I confirmed this also exists on the `main` branch but not on the currently deployed production Appointment.

The dashboard booking settings panel `booking confirmation` button on the dashboard is not being saved - when you change the setting it doesn't enable the save button. 

Not sure if that's something easy that you want to fix in this PR? I can create a separate issue for it if you like.

davinotdavid:
@rwood-moz Oh good catch! Perhaps that would be better suited for the [`Availability UI Overhaul` issue](https://github.com/thunderbird/appointment/issues/1094), I believe.

I started that to take a little break from Bookings but the [first PR](https://github.com/thunderbird/appointment/pull/1144) only moves components around.

I've opened a bug issue to track this:
https://github.com/thunderbird/appointment/issues/1145

backend/src/appointment/routes/api.py
davinotdavid: Definitely, good call! Added in latest commit

frontend/src/views/BookingsView/components/AppointmentSlidingPanelModify.vue
davinotdavid: Added! Here's the new ticket for reference: https://github.com/thunderbird/appointment/issues/1146

rwood-moz (approved):

PR #1141, opened 2025-07-24
Quick document on some (if not now, then soon) facts about appointment and datetimes.

MelissaAutumn:
Since we tend to run into the same issues I figured I'd create a source of truth for datetime guidelines/facts/etc...

It's meant to be a living document and not all of it is true right now, but we can workshop it as we go.

### DIFF
--- a/docs/datetime-guidelines.md
+++ b/docs/datetime-guidelines.md
@@ -1,0 +1,37 @@
+# Datetime Guidelines
+
+This is a short document intending to inform the reader of how we handle (or should be handling) datetimes in
+Appointment.
+
+## Facts
+
+### All server datetime and time fields should be stored as UTC time
+
+This normalizes datetime / time fields so we don't have to worry about not including timezone information,
+or doing any timezone comparisons when simply looking up a time or datetime.
+
+### All schedule comparisons should be converted and compared in the schedule owner's timezone
+
+This eliminates any day overflow errors when comparing a start and end time.
+
+For example, if someone wants to set their availability to `9 to 17` (`9am to 5pm`) America/Vancouver
+(Daylight savings - GMT-7) the server should not do this check in UTC as that would convert to `16 to 0` and the check
+would fail.
+
+### The server should be the one converting dates
+
+To avoid having to re-convert datetime fields during comparisons the server / backend should be tasked with converting
+datetime fields to UTC not the frontend.
+
+### Day comparisons should be calculated via the `toordinal()` function of a datetime object
+
+If you're comparing two dates against each other, do not use the current day of the month. Use the datetime fields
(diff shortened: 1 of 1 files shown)


docs/datetime-guidelines.md
```diff
@@ -0,0 +1,37 @@
...
+
+### All schedule comparisons should be converted and compared in the schedule owner's timezone
+
+This eliminates any day overflow errors when comparing a start and end time.
+
+For example, if someone wants to set their availability to `9 to 17` (`9am to 5pm`) America/Vancouver
+(Daylight savings - GMT-7) the server should not do this check in UTC as that would convert to `16 to 0` and the check
+would fail.
+
+### The server should be the one converting dates
```
MelissaAutumn, line 21: I believe this isn't true right now. But we should aim to make it be.

docs/datetime-guidelines.md
```diff
@@ -0,0 +1,37 @@
...
+
+This is a short document intending to inform the reader of how we handle (or should be handling) datetimes in
+Appointment.
+
+## Facts
+
+### All server datetime and time fields should be stored as UTC time
+
+This normalizes datetime / time fields so we don't have to worry about not including timezone information,
+or doing any timezone comparisons when simply looking up a time or datetime.
```
MelissaAutumn, line 11: Unsure if we should clarify that these datetimes should be timezone naive or timezone aware.

davinotdavid: I think it would be nice to clarify, yes. At least for me, it is useful to have it back as a reference when in doubt.

If the server datetime and time fields are UTC timezone naive, it would mean that we shift the responsibility on the client to do the conversions since it would be the only one timezone aware, no?

docs/datetime-guidelines.md
```diff
@@ -0,0 +1,37 @@
...
+
+To avoid having to re-convert datetime fields during comparisons the server / backend should be tasked with converting
+datetime fields to UTC not the frontend.
+
+### Day comparisons should be calculated via the `toordinal()` function of a datetime object
+
+If you're comparing two dates against each other, do not use the current day of the month. Use the datetime fields
+`toordinal()` function. This value is the absolute integer and incremented from by day starting from Jan 1st Year 1.
+
+This avoids having to take into account the month and year when the dates are cross-month or cross-year.
```
davinotdavid, line 31: Great tip!

docs/datetime-guidelines.md
devmount: I'm all for clarification too. Currently I think we have a mix of both: The client doing the timezone conversion itself in the frontend, and the client sending their current timezone information to the backend to get a converted datetime (e.g. on the booking request page).

Just a dumb question: How can a UTC datetime field be timezone aware? If we store in UTC, we'll always know how to convert to a different TZ, right?

PR #1144, opened 2025-07-25
Move schedule creation panel from Dashboard to new /availability route

davinotdavid:
## Description of the Change

- Created a new `/availability` route
- Moved schedule creation panel from Dashboard -> Availability page
- Updated E2E tests and data-testid to reflect the changes

Dashboard page
image

Availability page
image

PS: Functionality-wise it should be the exact same still, only moved components and files around at this point to prepare for the upcoming changes (trying to reduce PR size). It doesn't look super pretty atm but I left it the schedule creation panel as 50% width to represent a glimpse of the future.

## Benefits

- Better file structure as co-located components are easier to find / delete

## Applicable Issues

https://github.com/thunderbird/appointment/issues/1094


frontend/src/router.ts
```diff
@@ -75,7 +76,7 @@ const routes: RouteRecordRaw[] = [
   },
   {
     path: '/user/:username/:signatureOrSlug?',
-    name: 'availability',
+    name: 'booking-view',
```
davinotdavid, line 78: The named route here is unused and it can be confused with `bookings` so changed to `booking-view` for now but happy to change to whatever else for now.

Also, route name + view name combo is quite confusing IMHO. Perhaps we could standardize in the future for reducing cognitive load.

rwood-moz:
This looks really cool! \o/

FYI there is something up with the `Set general availability` slider/button. Note it is off by default however availability is turned on. Also if you change it, it doesn't enable the save changes button. Similar to the other slider/button issue I noted [here](https://github.com/thunderbird/appointment/pull/1138#issuecomment-3119747476). I checked production and it works fine there; guessing this is already in main too but not yet deployed.

Yep I switched back to main and confirmed this issue exists there already too.

davinotdavid:
Thank you @rwood-moz ! I've opened up a bug ticket to track this https://github.com/thunderbird/appointment/issues/1145

frontend/src/router.ts
MelissaAutumn: Yep, please normalize this when you come across it. Most of these occurences were from older versions of appointment.

rwood-moz:
Thanks for updating the E2E tests! I ran them locally against your branch and all passed for me too. 🥇

frontend/src/views/AvailabilityView/index.vue
```diff
@@ -0,0 +1,67 @@
...
+  scheduleStore.fetch();
+  schedulesReady.value = true;
+});
+</script>
+
+<script lang="ts">
+export default {
+  name: 'AvailabilityView'
+}
+</script>
```
MelissaAutumn, line 40: I'm guessing we need this because the file is named index.vue?

rwood-moz:
> Thank you @rwood-moz ! I've opened up a bug ticket to track this #1145

Great, thank you @davinotdavid

MelissaAutumn (approved):
lgtm!

frontend/src/views/AvailabilityView/index.vue
davinotdavid: Yep! That and the [lint rule for using multi word component names only](https://eslint.vuejs.org/rules/multi-word-component-names)

Issue #1092, opened 2025-07-08
Add sidebar with notification section and "What do you want to do today?" content


malini:
From the updated UI wireframes:

<img width="1289" height="731" alt="Image" src="https://github.com/user-attachments/assets/8313fa74-0fcd-4abe-b79e-c909313f75ce" />

davinotdavid:
Closing this as it has been merged to staging

PR #1150, opened 2025-07-25
Add quick actions sidebar to dashboard

davinotdavid:
## Description of the Change

Now that the schedule creation panel / availability setting has been moved to the dedicated `/availability` page, the calendar was sad by itself. This adds a quick actions sidebar!

As part of this PR I've also changed `ScheduleView.vue` to `DashboardView/index.vue` to be more representative of the route that it actually renders.

image

Zeplin screenshot for reference
image

Note: the "UI north star" is still under development so minor changes might follow

## Benefits

## Applicable Issues

https://github.com/thunderbird/appointment/issues/1092

### DIFF
--- a/frontend/src/views/DashboardView/components/QuickActionsSideBar.vue
+++ b/frontend/src/views/DashboardView/components/QuickActionsSideBar.vue
@@ -8,7 +8,7 @@ const user = useUserStore();
 
 const props = defineProps<{
   pendingBookingRequestsCount: number
-}>()
+}>();
 
 const copyBookingUrlLabel = ref(t('label.copyBookingUrl'));
 const hasCopied = ref(false);
@@ -30,7 +30,7 @@ const copyLink = async () => {
   <aside>
     <router-link v-if="props.pendingBookingRequestsCount" data-testid="link-pending-requests" class="link-buton pending"
       :to="{
-        path: '/bookings',
+        name: 'bookings',
         query: {
           unconfirmed: 'true'
         }


frontend/src/locales/de.json
devmount, line 1: Yes, that works well 👏🏻

frontend/src/views/DashboardView/components/QuickActionsSideBar.vue
```diff
@@ -0,0 +1,122 @@
...
+
+<template>
+  <aside>
+    <router-link v-if="props.pendingBookingRequestsCount" data-testid="link-pending-requests" class="link-buton pending"
+      :to="{
+        path: '/bookings',
+        query: {
+          unconfirmed: 'true'
+        }
+      }">
```
devmount: Do we maybe want to make this a named route? Or put it into a const? Since it is rather static at this place.

frontend/src/views/DashboardView/components/QuickActionsSideBar.vue
```diff
@@ -0,0 +1,122 @@
...
+import { ref } from 'vue';
+import { useI18n } from 'vue-i18n';
+import { useUserStore } from '@/stores/user-store';
+
+const { t } = useI18n();
+const user = useUserStore();
+
+const props = defineProps<{
+  pendingBookingRequestsCount: number
+}>()
```
devmount: I think we explicitly set `;` (maybe we should really run the frontend linter at some point 😂 )

devmount (approved):
Just some nits, thank you for the greate work!

frontend/src/views/DashboardView/components/QuickActionsSideBar.vue
davinotdavid: oh speaking of which, I think this is because my code editor is not setup properly 🤔 

I realized I was using the Vue (Official) but perhaps I should be using something else instead? I see that the run script in `package.json` has eslint being ran but when I run `npm run lint` it has no output

Screenshot 2025-07-28 at 11 29 01 AM

which extension / linter do you use so I can fix my env? in the future, it would be nice to have a pre-commit hook to run a `--fix` of sorts to prevent this from happening

frontend/src/views/DashboardView/components/QuickActionsSideBar.vue
davinotdavid: Good point! I've changed to the `bookings` named route now.

Did you mean just the path or the path + query param? I tried moving the the path + query param to a specific named route (e.g. 'bookings-unconfirmed-first') but it wasn't being interpreted as a `route.query` for the `useQueryParamState` logic for that page. Open to suggestions here 🙇

frontend/src/views/DashboardView/components/QuickActionsSideBar.vue
devmount: Hm, if `npm run lint` doesn't do it, then I think I was imagining things 😂  I have the same formatter like you. So guess it was another project then. Personally I like to set semicolons for better readable code. I'll take a look at the eslint config and see if that is configured correctly. I'm not so much of a pre-commit-hook-fan myself, in my experience this breaks the dev flow more than it helps. But good point anyway!

frontend/src/views/DashboardView/components/QuickActionsSideBar.vue
devmount: Oh my bad, I thought this was possible, but yeah, according to the docs it seems it's only possible for routes like `/booking/:unconfirmed`, not for query parameters.

frontend/src/views/DashboardView/components/QuickActionsSideBar.vue
davinotdavid: Ah I don't have strong opinions either way, I only wish that I could turn on `format on save` and leave it to project defaults somehow 😅 For the time being, I will keep in mind about the semicolon preference! Thanks for sharing!

PR #1151, opened 2025-07-28
[WIP] Bookings page UI overhaul - Part 4 (Pagination w/ Filters + Infinite Scrolling)

davinotdavid:
### Frontend changes
- Updated overall height calculation on the whole app so that we can determine the height of elements within pages reliably with `flex`, therefore allowing for scroll manipulation within the bookings table.

### Backend changes
- Added support for pagination and filter query params to `GET /me/appointments`
- Added a new `GET /me/pending_appointments_count` endpoint and a `repo.appointment.count_by_subscriber` method since there are two places in the frontend (dashboard an profile page) that only care about the count of pending appointments and those were fetching _all_ appointments

## Benefits

## Applicable Issues


backend/src/appointment/routes/api.py
```diff
@@ -100,19 +101,86 @@ def read_my_calendars(
...
-def read_my_appointments(db: Session = Depends(get_db), subscriber: Subscriber = Depends(get_subscriber)):
-    """get all appointments of authenticated subscriber"""
-    appointments = repo.appointment.get_by_subscriber(db, subscriber_id=subscriber.id)
+@router.get('/me/appointments', response_model=schemas.ListResponse)
+def read_my_appointments(
+    page: int = 1,
+    per_page: int = 50,
+    status: list[str] = Query(
+        default=[],
+        description='Filter appointments by booking status (requested, booked, declined, cancelled, modified)',
```
devmount, line 110: Do we want localization here (and l.167 too?)?

frontend/src/views/BookingsView/index.vue
```diff
@@ -148,12 +142,28 @@ const ariaSortForColumn = (column: TableColumn) => {
...
+// Handle infinite scrolling
+const handleScroll = async (event: Event) => {
+  const target = event.target as HTMLElement;
+  const { scrollTop, scrollHeight, clientHeight } = target;
+
+  // Load more when user scrolls to bottom (with a small threshold)
+  if (scrollHeight - scrollTop <= clientHeight + 100 && hasMorePages.value && !isLoading.value) {
+    await appointmentStore.loadMore(selectedFilters.value);
+  }
+};
```
devmount, line 154: Can we put that into a utility function (e.g. with a callback function parameter) for reuse on other pages?

frontend/src/App.vue
devmount, line 1: Thanks for starting the Tailwind removal here :)

devmount (approved):
This is awesome, many thanks for providing dedicated endpoints for counts only and all the other improvements! 🎉 

> It would be nice to have a seeds script of sorts to create a bunch of valid appointments to facilitate testing

That is a great idea, I feel that pain (but apparantly it wasn't great enough to build some seeders myself until now 😂) and created a ticket for that (#1160).

> The Dashboard page now fetches only the first page of appointments which may or may not be enough to comprehensively fill the whole calendar (as in, if the user has more than 50 appointments, some would be missing from the page). Unsure if the best way would be to allow someone to fetch all appointments as it was before.

First I think it's good to limit the initial request like this. We could explore a "load more" / "load all" functionality in a later iteration. The only thing I think is important is, that the user knows, they're only seeing the first 50 records (so somewhere should be a hint preventing confusion about missing appointments).

backend/src/appointment/routes/api.py
davinotdavid: Oh I thought this would be internal docs only (like the docstrings from method descriptions throughout the code) but happy to move it to a comment instead if needed. Good point about internationalization of docs though, I am not familiar on how to achieve that but can research 🤔

devmount: No no all good then! If it's just docs it's fine. I don't think we need to think about l10n for docs yet.

frontend/src/views/BookingsView/index.vue
davinotdavid: Hmm while this is generally a good idea, it seems like this is the only page / component in Appointment that will have infinite scrolling even after all the UX overhaul of all pages. 

How about we extract it in case that changes in the future?

devmount: Sure, no problem

PR #1152, opened 2025-07-29
Update dependencies

devmount:
## Description of the Change

The current frontend dependency status is the following:

image

This PR updates all packages with minor changes and 4/5 packages with major changes (We keep Tailwindcss v3 since we want to remove tw in the future).

The current testing dependency status is the following:

image

This PR updates them and hence should fix https://github.com/thunderbird/appointment/security/dependabot/58 which has a surprisingly high CVE.

## Benefits

Stability, security.

## Applicable Issues

None.


davinotdavid (approved):
Awesome, thanks! Just ran the E2E suite locally and it all passes 👍

PR #1155, opened 2025-07-29
Availability page UI overhaul - Part 2 (Component restructure w/ same functionality)

davinotdavid:
## Description of the Change

- This PR essentially breaks down the former [ScheduleCreation.vue](https://github.com/thunderbird/appointment/blob/main/frontend/src/views/AvailabilityView/components/ScheduleCreation.vue) component (now gone) into several components, all connected to the newly created `AvailabilityStore` that handles the form state (as `currentState`).

- We are now leveraging [Pinia's $patch method](https://pinia.vuejs.org/core-concepts/state.html#Mutating-the-state), where each component can now use `computed` properties to get / set form fields directly while keeping the reactivity and without having to resort to `watch` or other workarounds.

Zeplin Wireframe
Appointment_MVP_Availability

Light mode
image

Dark mode
image

## Benefits

- Better DX and separation of components

- There are other places that uses the ScheduleStore, the biggest one being the FTUE (First Time User Experience) flow. This looks similar-ish with the AvailabilityStore so I get that we might feel the urge of refactor and merging them together somehow. However, I am choosing to keep this separate for now as IMHO it looks quite clean to leave the AvailabilityStore to handle only the form management.
- Speaking of the Availability page, I don't like very much that we still require some bits of logic present in the Schedule Store in order to save the form. This could potentially be ported over to the AvailabilityStore in a future PR.
- Add `data-testid`s / update existing E2E tests.
- The main toggle that enables / disables the availability is not keyboard focused. Opened up a ticket for it in `services-ui` https://github.com/thunderbird/services-ui/issues/102.
- This PR does NOT include "buffer time" (https://github.com/thunderbird/appointment/issues/1164) or "virtual meeting details" (unsure of what that means yet).

## Applicable Issues

https://github.com/thunderbird/appointment/issues/1094


frontend/src/locales/de.json
```diff
@@ -113,6 +112,7 @@
         "zoom": "Zoom"
       }
     },
+    "setYourAvailability": "Set Your Availability",
```
MelissaAutumn: :de: :eyes:

frontend/src/views/AvailabilityView/components/AvailabilitySettings/index.vue
```diff
@@ -0,0 +1,247 @@
...
+        :slot-duration="slotDuration"
+        :required="true"
+        v-model="weekDays"
+        @update="onAvailabilitySelectUpdated"
+        :disabled="!currentState.active"
+      />
+    </template>
+
+    <!-- Availability without customization -->
+    <template v-if="!useCustomAvailabilities">
```
MelissaAutumn: Can we use `v-else(-if)` instead of a separate if?

frontend/src/views/AvailabilityView/components/BookingPageDetails/index.vue
```diff
@@ -0,0 +1,161 @@
...
+
+    <!-- Auto generate Zoom link in invites -->
+    <div class="auto-generate-link-container">
+      <checkbox-input
+        name="autoGenerateZoomLink"
+        :label="t('label.autogenerateZoomLinks')"
+        v-model="autoGenerateZoomLink"
+        :disabled="!hasZoom"
+      />
+      <i18n-t v-if="!hasZoom" keypath="text.generateZoomMeetingHelpDisabled.text" tag="span" scope="global" class="text-xs">
```
MelissaAutumn: is text-xs ours or tailwinds?

frontend/src/views/AvailabilityView/index.vue
```diff
@@ -1,36 +1,97 @@
...
+
+function validateSchedule(schedule) {
+  // Schedule name is empty
+  if (schedule.name === '') {
+    return t('error.fieldIsRequired', { field: t('label.pageName') });
+  }
+
+  // All good
+  return null;
+}
```
MelissaAutumn, line 40: Probably not worth it for one field, but you can leverage html5's form validation here by putting a ref on the form element, and adding "required" to pageName's element. Then run form.value.checkValidity() to determine if any fields fail. 

E.g. (The login view is messy so here's a few pointer links)
https://github.com/thunderbird/appointment/blob/main/frontend/src/views/LoginView.vue#L33
https://github.com/thunderbird/appointment/blob/main/frontend/src/views/LoginView.vue#L195
https://github.com/thunderbird/appointment/blob/main/frontend/src/views/LoginView.vue#L236

You can additionally handle custom messages by using `@invalid` event.

frontend/src/locales/de.json
davinotdavid: Hoppla! 😅

frontend/src/views/AvailabilityView/components/AvailabilitySettings/index.vue
davinotdavid: For sure, changed to `v-else`, thanks!

frontend/src/views/AvailabilityView/components/BookingPageDetails/index.vue
davinotdavid: Tailwinds, copy and paste got me here heh! Changed, thanks!

frontend/src/views/AvailabilityView/index.vue
davinotdavid: Good point! [`pageName` already has a `required`](https://github.com/thunderbird/appointment/pull/1155/files#diff-f98cf6f274907bcb981c632e06262193634dd661f0c5f262ff73b82f00abd51aR79) and this was copying over [the previous validation](https://github.com/thunderbird/appointment/blob/main/frontend/src/views/AvailabilityView/components/ScheduleCreation.vue#L258-L266) but indeed this whole page is lacking quite a bit of form validations I believe.

frontend/src/locales/de.json
```diff
@@ -113,6 +112,7 @@
         "zoom": "Zoom"
       }
     },
+    "setYourAvailability": "Legen Sie Ihre Verfügbarkeit fest",
```
devmount: We're mostly using a passive form to keep things short and simple:


```suggestion
    "setYourAvailability": "Verfügbarkeit festlegen",
```

frontend/src/locales/de.json
```diff
@@ -300,9 +309,10 @@
     "logIn": "Anmelden",
     "logOut": "Abmelden",
     "loginToContinue": "Anmelden, um fortzufahren",
-    "meetingDetails": "Meeting Details",
+    "meetingDuration": "Meeting Dauer",
```
```suggestion
    "meetingDuration": "Meeting-Dauer",
```

frontend/src/locales/de.json
```diff
@@ -300,9 +309,10 @@
...
     "meetingTitle": "Meeting Titel",
     "message": "Nachricht",
+    "minimumNotice": "Mindestfrist",
```
```suggestion
    "minimumNotice": "Mindestvorlaufzeit",
```

frontend/src/locales/de.json
```diff
@@ -410,12 +416,16 @@
     "videoLink": "Konferenz-Link",
     "viewAll": "Alle ansehen",
     "viewBooking": "Buchung anschauen",
+    "virtualMeetingLink": "Virtuelle Konferenz-Link",
```
```suggestion
    "virtualMeetingLink": "Online-Meeting-Link",
```

devmount (approved):
Great stuff, Davi! This was a lot of work, thanks for this!

I suggested some changes to the German translation. Also when testing, I found three minor things, no blockers imo. Other than that: approved! 🎉 

### Resetting error messages doesn't work properly

When getting a form error (e.g. availability is invalid), the error message won't disappear when the error gets corrected. I saved the valid schedule config and the error message was still there.

Image

### Styling of Radio Group Pill

The select pills are really nice! I just wondered if this blueish border is intended on an active element:

Image

### Snackish Bar hides content

When the form is dirty, the snackish bar appears correctly, but it seems to overlay the last part of the form, even when scrolled down to the bottom. Maybe we can give the content some more padding bottom when the snackish bar is visible:

Image

davinotdavid:
@devmount 

Thank you so much for the review! I really appreciate it 🙇 From your points, here are the ones addressed:

- i18n for de.json
  - Applied in the commit https://github.com/thunderbird/appointment/pull/1155/commits/428bb6ae29e805582d571b36d36999c14b8dbfa4 (thanks again!!)
  
- Resetting error messages doesn't work properly
  - Ah, I wasn't cleaning up the error on the success case 🤦 . Updated in commit https://github.com/thunderbird/appointment/pull/1155/commits/3cc2a8dcd7ce169b27727cf341faa564fa503aa6.
  
- Styling of Radio Group Pill
  - Thanks for the compliments! This one I prefer not to address in this PR as it will probably change styles very soon when the "north star UI" gets ready to be implemented. For now, it was just me / imagination from the wireframes!
  
- Snackish Bar hides content
  - Great catch! I've added some extra margin to the page to account for the "dirty footer" that shows up. Thanks! Updated in commit https://github.com/thunderbird/appointment/pull/1155/commits/3cc2a8dcd7ce169b27727cf341faa564fa503aa6.

MelissaAutumn (approved):

devmount:
Awesome, thanks Davi!

Issue #1145, opened 2025-07-25
Set General Availability and Booking Confirmation not working


davinotdavid:
**Describe the bug**
In the Schedule Creation Panel, the `Set General Availability` and `Booking Confirmation` toggles are not being saved

**To Reproduce**

1. Go to '/availability'
2. Toggle the `Set General Availability` or `Booking Confirmation` toggle
3. Refresh the page
4. Check that the changes don't persist

**Expected behavior**
The toggled settings should persist and behave according to their text

**Actual behavior**
The toggled settings don't get saved

**Screenshots**

Both toggles can be seen here:

<img width="652" height="678" alt="Image" src="https://github.com/user-attachments/assets/76cc9414-302f-418b-8ded-066d86a0a09a" />

**System (please complete the following information):**

- OS: macOS 
- Browser Version [e.g. Firefox 141.0 (aarch64)]

davinotdavid:
Closing this as PR has been merged

PR #1159, opened 2025-07-31
Fix General Availability SwitchToggles

davinotdavid:
## Description of the Change

Updating `services-ui` fixes half of the problems here as it is now emitting the `change` events on the SwitchToggle per the change in the [newest version fix](https://github.com/thunderbird/services-ui/pull/95).

However, the `booking_confirmation` had a second problem where it wasn't being initialized properly. The `deepClone` call in the `onMounted` hook ([ref](https://github.com/thunderbird/appointment/compare/fix-general-availability-toggles?expand=1#diff-1ba5ada7f435e4d1f83537e333373409187f31fc53195039ab464a36cff55879R118)) is breaking the Vue's reactivity for this field. This will be a non-issue with the upcoming changes for the [Availability page overhaul / rewrite](https://github.com/thunderbird/appointment/issues/1094) so the fix in this PR is enough for now.

## Benefits

Toggles behave as expected

## Applicable Issues

https://github.com/thunderbird/appointment/issues/1145
https://github.com/thunderbird/services-ui/issues/94

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -14,7 +14,7 @@
         "@sentry/vue": "^9.10.1",
         "@tabler/icons-vue": "^3.7.0",
         "@tailwindcss/forms": "^0.5.3",
-        "@thunderbirdops/services-ui": "^0.6.18",
+        "@thunderbirdops/services-ui": "^0.6.20",
         "@types/ua-parser-js": "^0.7.39",
         "@vitejs/plugin-vue": "^6.0.1",
         "@vueuse/components": "^13.0.0",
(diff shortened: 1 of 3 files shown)


devmount (commented):
Lgtm, thanks @davinotdavid for fixing this! One question though: Is the fix in services-ui actually released? Because it's [not part of the release notes](https://github.com/thunderbird/services-ui/releases/tag/v0.6.20)?

davinotdavid:
> Lgtm, thanks @davinotdavid for fixing this! One question though: Is the fix in services-ui actually released? Because it's [not part of the release notes](https://github.com/thunderbird/services-ui/releases/tag/v0.6.20)?

It is released, yes! I can verify that the bug exists if I downgrade and set the version up to 0.6.19 and the bug is fixed when setting the version to 0.6.20.

I am not sure why it doesn't show up in the release notes though very likely something wrong that I did during the release process (again 😅).

Issue #1158, opened 2025-07-30
Database passwords with special characters break Appointment


ryanjjung:
SQLAlchemy has some [odd behavior around quoting passwords](https://docs.sqlalchemy.org/en/14/core/engines.html#escaping-special-characters-such-as-signs-in-passwords). We have found that the way we currently handle passwords does not handle these cases at all, leading to either a broken environment or a restriction on password complexity.

We get this data from strings right now, but the recommended way of presenting database URLs is through the [sqlalchemy.engine.url.URL](https://docs.sqlalchemy.org/en/14/core/engines.html#creating-urls-programmatically) class, which should handle all the necessary special character handling.

Issue #1110, opened 2025-07-14
Remove the normalize_secrets function


ryanjjung:
[Ref](https://github.com/thunderbird/appointment/blob/main/backend/src/appointment/utils.py#L105)

This piece of code reads in secrets which have been pulled from Secrets Manager and stored in environment variables. Those values are JSON-formatted, containing several different values. This necessitates code on the app side to parse those values into individual variables internally. This whole process would be more transparent and less complex if we separated each of these variables out into separate inputs.

So that's the plan:

- Change the secrets so that they are each individual environment variables.
  - This means updates to Pulumi secrets and the PulumiSecretsManager config.
  - These have to be deployed to environments before the new image goes out.
- Port those environment variables directly into the places they need to be.
  - Some of these are DB-related, some Zoom-related, and several others.
- Delete the normalize_secrets function.

ryanjjung:
Actually, we can't remove the old secrets until after we have migrated to this infrastructure. All of this is still "not live" in the sense that folks are still using the old `appointment.day` (Terraform-backed) domain. So the cleanup steps can't be done until #1104 is complete.

However, I can still make some progress and bring the new variables online in the newer `appointment.tb.pro` (Pulumi-backed) domain.

ryanjjung:
# Appointment Normalized Variables Before and After

I did a full inventory of the JSON-formatted secrets we currently rely upon and which of those objects' keys map to which environment variables in the `normalize_secrets` function. I found the versions of these which we rely upon for our Terraform-backed environments (using the `appointment.day` domains, slated to be destroyed when we migrate to the newer infra) and those which the Pulumi environments (using the `appointment.tb.pro` domains, slated to be migrated to) use.

This comment serves as a plan of action and a tracking chart (mostly for my own sanity) for the work of migrating these complex inputs into individual inputs and ultimately removing the `normalize_secrets` function entirely. Many of the values in these secrets are not actually secret at all, and those will be moved straight into the task defintions in plaintext, removing the abstraction in these cases. In other cases, certain values have been lumped in with categories they don't really belong to. These will be nominally separated after these changes.


## DATABASE_SECRETS

This holds database connection information.


### Old Secrets

- appointment/stage/database-connection (Pulumi Stage)
- appointment/prod/database-connection (Pulumi Prod)
- tb-apmt-stage-db-XVbgRSzR (TF Stage)
- tb-apmt-prod-db-OQEwTVTq (TF Prod)


### Secret Fields

- engine
- username
[...]

PR #1161, opened 2025-08-04
Fix a few database setup issues

ryanjjung:
## Description of the Change

For one thing, we would like to ultimately [remove the `normalize_secrets` function](https://github.com/thunderbird/appointment/issues/1110). While responding to the problem of this software throwing database authentication errors constantly, I discovered the problem was that the password contained a `%` sign, and that SQLAlchemy would routinely choke on any password containing this, an `@` symbol, or presumably some other special characters. I tried a number of methods of escaping those characters, primarily trying to use `urllib.parser.quote_plus`, which is what I had seen as a solution online.

But as I dug deeper into the code, I saw we were pulling this value right out of the environment as a string independently in a lot of different places, making it non-trivial to run the `quote_plus` solution (and corresponding "unquotes") all over the place while keeping the pattern and intent obvious to future developers.

I also learned that SQLAlchemy has a built-in URL builder that already handles all of the character-escaping for us. This PR does a few things in this vein:

- Centralizes the URL building in a single cached function
- Adjusts input processing to prefer individual URL components (host, port, password, etc.) over one mega-string, and thus to prefer `sqlalchemy.url.URL` objects over plain connection strings.
- Shims these preferences into the Alembic config, except for "offline" migrations

## Benefits

- Artificial password complexity limitations removed!
- Progress toward removal of `normalize_secrets`!

## Applicable Issues

- #1110
- #1158


backend/src/appointment/commands/update_db.py
```diff
@@ -1,15 +1,16 @@
-import os
 
 from ..database import models
 from alembic.runtime import migration
 
 from ..dependencies.database import get_engine_and_session
-from ..utils import normalize_secrets
+from ..main import _common_setup
```
ryanjjung, line 6: `_common_setup` now calls `normalize_secrets`, so we don't need this import any longer.

backend/src/appointment/commands/update_db.py
```diff
@@ -1,15 +1,16 @@
...
+from ..utils import get_database_url
 
 
 def run():
     print('Checking if we have a fresh database...')
 
+    _common_setup()
```
ryanjjung, line 13: Make sure to call common setup, which gets a bunch of environmental stuff setup.

backend/src/appointment/main.py
```diff
@@ -222,10 +222,10 @@ async def warn_about_deprecated_routes(request: Request, call_next):
         if request.scope.get('route') and request.scope['route'].deprecated:
             app_env = os.getenv('APP_ENV')
             if app_env == APP_ENV_DEV:
-                logging.warning(f"Use of deprecated route: `{request.scope['route'].path}`!")
+                logging.warning(f'Use of deprecated route: `{request.scope["route"].path}`!')
             elif app_env == APP_ENV_TEST:
                 # Stale test runtime error
-                raise RuntimeError(f"Test uses deprecated route: `{request.scope['route'].path}`!")
+                raise RuntimeError(f'Test uses deprecated route: `{request.scope["route"].path}`!')
```
ryanjjung, line 228: I must have `ruff format`ted this file in my IDE. I ran a separate `ruff format` and it wants to change basically every Python file. I will file a separate ticket to `ruff format` the repo and set it up as a validation step in CI.

backend/src/appointment/utils.py
```diff
@@ -7,13 +8,15 @@
 from functools import cache
 
 from argon2 import PasswordHasher
+from sqlalchemy.engine.url import URL as sqlalchemy_url
```
ryanjjung, line 11: We have other variables in this file called `url`, so I opted for a longer import name for clarity.

backend/.env.example
```diff
@@ -20,7 +20,16 @@ FRONTEND_URL=http://localhost:8080
...
 # -- DATABASE --
-DATABASE_URL="mysql+mysqldb://tba:tba@mysql:3306/appointment"
+# The use of DATABASE_URL is being phased out. It should only be used for offline migrations.
+#  DATABASE_URL="mysql+mysqldb://tba:tba@mysql:3306/appointment"
+
+# Instead, use these values to engage the ORM's built-in URL factory
+DATABASE_ENGINE=mysql
+DATABASE_HOST=mysql
+DATABASE_NAME=appointment
+DATABASE_PASSWORD=abcd%efgh
```
ryanjjung, line 30: Changed the dev password to include the `%` sign that tipped us off to the problem.

backend/src/appointment/main.py
MelissaAutumn: I think you'll need to specify the ruff options file as pyproject.toml, as we use single quotes
https://github.com/thunderbird/appointment/blob/main/backend/pyproject.toml#L83

backend/src/appointment/utils.py
```diff
@@ -103,38 +106,76 @@ def chunk_list(to_chunk: list, chunk_by: int):
...
+    dialect = os.environ.get('DATABASE_ENGINE', 'mysql')
+    driver = determine_database_driver(dialect=dialect)
+    port = int(os.environ.get('DATABASE_PORT', '3306'))
+
+    # These settings are secrets and must be set manually.
+    host = os.environ.get('DATABASE_HOST')
+    username = os.environ.get('DATABASE_USERNAME')
+    password = os.environ.get('DATABASE_PASSWORD')
+
+    if not db_name or not dialect or not host or not password or not port or not username:
```
MelissaAutumn:
```suggestion
    if not all([db_name, dialect, host, password, port, username]):
```
Should be a slightly cleaner way of checking this.

MelissaAutumn (changes requested):
Please double check your ruff and fix the lint on that one file, but otherwise looks good.

This PR will cause existing dbs to fail password authentication because of the password change. Maybe change the apmt_db volume name to a different one? Like apmt_db_v2? 

If we encounter more db related changes we just can just increment the version number in the future.

backend/src/appointment/utils.py
ryanjjung: Yeah, I like that better. Will push this change.

MelissaAutumn (approved):
conditional approve on fixing formatting + not all

PR #1162, opened 2025-08-05
Run ruff format

ryanjjung:
This massive PR makes no functional changes to the code. It only applies our existing Ruff config to the files and adds a validation CI flow step to make sure we are applying it consistently. The console output:

```
> ruff format                                                                                                                                                                            
48 files reformatted, 47 files left unchanged
```


ryanjjung:
Wrong branch.

PR #1163, opened 2025-08-05
Fix formatting

ryanjjung:
This massive PR makes no functional changes, only applies our existing Ruff rules to the code in `main` and adds a CI validation check to keep us consistent.


MelissaAutumn (approved):

ryanjjung:
No problem, @devmount. I ran it the other day and it changed all these files and I immediately `git reset --hard HEAD` and made a plan to do this.

Issue #1096, opened 2025-07-08
Settings page overhaul


malini:
From the updated UI wireframes:

<img width="712" height="872" alt="Image" src="https://github.com/user-attachments/assets/2f8f7410-1fa7-453e-8648-f8fff89649fa" />

and show any errors like so:

<img width="662" height="869" alt="Image" src="https://github.com/user-attachments/assets/cd9a2b44-04dc-4a41-9048-a9812b677cd4" />

Other than reorganizing existing data/features, the following will change:
* primary email account is no longer configurable within Appointment. You have to click 'Manage Account' which will take you to accounts.tb.pro instead
* Start of the Week is completely configurable instead of either Monday or Sunday. This would be the result of removing Qalendar and relying on simpler selectors
* Validation feedback: Connected Applications tab will have a warning icon next to it if any authentication fails for linked accounts, and the panel for Connected Applications will have more information about what is failing
* "Save" button. Instead of instantly changing settings as it works now, the user will have to commit their changes with the 'Save' button

MelissaAutumn:
@Jesse-Moz @laurelterlesky since we locked changing primary email account on thundebird accounts for now, should we keep it in Appointment's settings page for now?

laurelterlesky:
Yes, absolutely. I'll add an annotation in Zeplin

laurelterlesky:
^ Question though - how will we migrate our beta users @malini

MelissaAutumn:
Very carefully...but thats something i need to look into. We need to do a db wipe for #855, but we should retain the user emails so we can import them into keycloak.

We wont have to worry about old user prefs.

davinotdavid:
The overhaul has been merged and deployed to production. The only bit missing for the original requirements is related to the external connection errors.

That error handling will be a little bit more involved as it requires some sort of "health check" instead of the binary connected / not-connected that comes from the DB.

Created https://github.com/thunderbird/appointment/issues/1195 to track this missing bit and closing this issue!

PR #1166, opened 2025-08-06
[WIP] Settings page UI overhaul

davinotdavid:
## Applicable Issues

https://github.com/thunderbird/appointment/issues/1096


rwood-moz:
Hey @davinotdavid I have created a separate issue (#1187) to update the E2E tests for these changes, and I can take care of that if you like (I'll wait until you add the `data-testid` elements in this PR or a follow-up one).

frontend/src/router.ts
```diff
@@ -207,6 +207,17 @@ const routes: RouteRecordRaw[] = [
 const router = createRouter({
   history: createWebHistory(import.meta.env.BASE_URL),
   routes,
+  // Ref: https://router.vuejs.org/guide/advanced/scroll-behavior.html
+  scrollBehavior(to) {
+    if (to.hash) {
+      // This ensures that if hash is provided to router.push it works as expected.
+      return {
+        el: to.hash,
+        behavior: 'smooth',
```
MelissaAutumn: Can we make this dependent on prefers reduced motion? If its reduced make it instant.

frontend/src/views/SettingsView/components/AccountSettings.vue
```diff
@@ -0,0 +1,189 @@
...
+    // TODO: show error
+    // console.warn('ERROR: ', error.value);
+    return;
+  }
+
+  // We can't logout since we've deleted the user by now, so just delete local storage data.
+  userStore.$reset();
+  await router.push('/');
+};
+</script>
```
MelissaAutumn, line 57: I'm not sure how we'll be handling this with accounts, but any deletion or cancellation will probably not be handled in appointment anymore. We can keep this for now though.

davinotdavid:
@rwood-moz that would be wonderful, yes! Thanks a lot 🙇

frontend/src/locales/en.json
```diff
@@ -155,13 +156,15 @@
     "MMDDYYYY": "MM/DD/YYYY",
     "accepted": "Accepted",
     "account": "Account",
-    "accountData": "Download your account data",
     "accountDeletion": "Delete your account",
+    "accountSettings": "Account settings",
```
rwood-moz: Should be uppercase "Account `Settings`".

frontend/src/views/SettingsView/components/AccountSettings.vue
```diff
@@ -0,0 +1,189 @@
...
+  <header>
+    <h2>{{ t('label.accountSettings') }}</h2>
+  </header>
+
+  <div class="booking-page-url-container">
+    <label for="booking-page-url">
+      {{ t('label.bookingPageURL') }}
+    </label>
+    <div class="booking-page-url-input-container">
+      <text-input name="booking-page-url" class="booking-page-input" v-model="userStore.data.userLink" />
```
rwood-moz: The booking page URL displayed is missing the optional hash at the end, ie. my share link is:

`http://localhost:8080/user/<username>/91ecbe99/`

but the booking page URL field just shows:

`http://localhost:8080/user/<username>/`

However when I click the `Copy link` button, the hash at the end IS copied and I get the correct full link.

frontend/src/locales/en.json
```diff
@@ -216,6 +221,7 @@
     "cancelled": "Cancelled",
     "cancelBooking": "Cancel booking",
     "cancelReason": "Reason for cancelling",
+    "cancelService": "Cancel service",
```
rwood-moz: Should be uppercase "Cancel `Service`".

frontend/src/locales/en.json
```diff
@@ -254,10 +260,11 @@
     "day": "Day",
     "deactivateSchedule": "Deactivate schedule",
     "declined": "Declined",
+    "default": "Default",
     "defaultFont": "Default font",
+    "defaultTimeZone": "Default Time Zone",
```
rwood-moz, line 265: Not sure it should be called "`Default` Time Zone", on the current production settings page it is called `Primary` time zone, and on the set your availability page it's just called `Time Zone`.

frontend/src/locales/en.json
davinotdavid: Thanks! Actually I've realized that I accidentally duplicated the key, so this is now removed in favor of `headings.accountSettings` which already has the correct casing.

frontend/src/views/SettingsView/components/AccountSettings.vue
davinotdavid: Good point. For consistency, I've added the slug to the end of the link url. Thanks!

frontend/src/locales/en.json
davinotdavid: Changed, thanks!

frontend/src/locales/en.json
davinotdavid: This one is a little tricky. I've had a meeting with @Jesse-Moz last week and this is deliberate – the idea here is that in the (near) future there will be different schedules and each can have its own timezone.

So even though currently the Availability page and the Settings page change the same timezone settings, it won't be once we have multiple active schedules.

davinotdavid:
@rwood-moz thanks for the review! Really appreciate it 🙏 

> After changing the language (ie. From English to German) and saving the settings, the language change doesn't actually take effect - you need to do a manual page refresh for the language to actually change

Ah yes, good point! I've missed the update to the actual i18n composable. Changed here, thanks!

> If you make changes to connected calendars ie. turn OFF the check box for a currently connected calendar, and then click Revert changes, the change to the check box is NOT reverted

Good catch! Updated the checkbox logic to be revertable correctly.

> If you make any changes the Save changes button at the bottom appears but in the wireframe it shows the button label to just be Save not Save changes

Updated both the Availability page and Settings page to be `Save` instead of `Save Changes` to be more aligned with the wireframes. Thanks!

> If you change the theme it would be nice to preview it ie. have it applied to the current settings page even before saving the settings changes; just my opinion - as currently in prod when you change settings including the theme they are applied immediately without having to click a Save button

Ah, this one is deliberate by the design team. Unsure about the specifics on why that is but it was requested to add one more step to actually apply the changes in this case.

frontend/src/locales/en.json
rwood-moz: Ah makes sense, thank you for clarifying that.

frontend/src/views/SettingsView/components/AccountSettings.vue
```diff
@@ -0,0 +1,189 @@
...
+// Link copy
+const copyLink = async () => {
+  await navigator.clipboard.writeText(userStore.myLink);
+
+  copyLinkTooltip.value = t('info.copiedToClipboard');
+
+  setTimeout(() => {
+    copyLinkTooltip.value = t('label.copyLink');
+  }, 2000);
+};
```
MelissaAutumn, line 32: Should we maybe make this a component itself? I use this within keycloak, while it's not a foundational component it could be handy to just have for any service.

MelissaAutumn (approved):
Only real issue I found is the prefers-reduced-motion comment. Rest are good!

frontend/src/views/SettingsView/components/AccountSettings.vue
davinotdavid: Yep, also I assume that the flow for this to happen will probably change, specially wording-wise / options. A good follow up to keep in mind for sure when the time comes!

frontend/src/views/SettingsView/components/AccountSettings.vue
davinotdavid: Ah I'd love to, honestly! The annoying-ish part is that we have it in different flavors at the moment with slightly different presentation / copy / behavior 😅 :

- NavBar, we have the icon button to copy the link and the tooltip's text change
- Dashboard, on the aside, its a text button that its copy change (but doesn't have a tooltip)
- Settings page, also an icon button with tooltip but different icon

How's it used in keycloak? Potentially we could punch this into a separate ticket with all the current usage and make some props / variations to conquer them all!

frontend/src/router.ts
davinotdavid: Ah good point, missed this one! Changed here, thanks!

frontend/src/views/SettingsView/components/AccountSettings.vue
MelissaAutumn: We use it to copy the OTP auth url if they cannot scan the qr code for setting up MFA. I'll bring this up to one of the design channels to see how we want to proceed.

davinotdavid: Linking the issues in `services-ui` that you've created for tracking purposes:

https://github.com/thunderbird/services-ui/issues/109
https://github.com/thunderbird/services-ui/issues/110

PR #1168, opened 2025-08-07
Set up new task def for stage/mysql

ryanjjung:
If Phase One was "Fix DB Connection Strings and the Variables that Construct It", then this is Phase Two: Make the Task Definition Use Those Variables. This keeps the service pointed at the existing MySQL database and gives us a safe rollback point in case the Postgres migration doesn't go well.


ryanjjung:
Pertains to #1067

Issue #322, opened 2024-03-18
Unify the desktop and mobile UX


malini:
**Is your feature request related to a problem? Please describe.**
We've been developing a mobile experience without direct input from our design team. We need a solid story for UI/UX for mobile and desktop, and make them both look good. Ex:

![Thunderbird Appointment](https://github.com/thunderbird/appointment/assets/24374/a59e6d4d-81f2-4ea7-9fde-4f9401ad142a)

**Describe the solution you'd like**
Work with the design team to determine what each workflow should look like and implement it.

laurelterlesky:
Mobile Navigation: https://www.figma.com/design/rgWIxk5tu39vLl0j35IiAA/TBA_2024_Q3_NavBarMobile?node-id=1-6&t=fWTwaWYCOP9hAsW5-1

cc: @solangevalverde

davinotdavid:
Since opening this issue, we have been through a few different UI iterations and this has been reasonably being already addressed. Closing this issue!

PR #1172, opened 2025-08-11
Mobile Menu for Appointment

davinotdavid:
## Description of the Change

- Added a NavBarMobile component!
  - Design has been approved by the design team for now, except that there might be small tweaks to be done in the near future since the Figma that I've based myself off of is a little outdated already.
  - Using `display: none` / `display: show` and separated the Desktop and Mobile navs in their own separate components so that we can avoid using JS window listeners for that and they are significantly different to be together.
  - It doesn't receive the `navItems` as props. I see the value of using it before but that was coupling the navigation order and translation file together which gives us less flexibility. For instance, the mobile navigation has a different navigation item order.

Frame 3

## Benefits

- It was very very hard to use Appointment overall without a mobile navigation. It doesn't solve all the responsive problems but it is still a step into the right direction!

## Applicable Issues

https://github.com/thunderbird/appointment/issues/322

### DIFF
--- a/frontend/src/components/NavBar.vue
+++ b/frontend/src/components/NavBar.vue
@@ -51,12 +51,6 @@ const copyLink = async () => {
 </script>
 
 <template>
-  <!-- Mobile NavBar -->
-  <header class="header-mobile">
-    hi
-  </header>
-
-  <!-- Desktop NavBar -->
   <header class="header-desktop">
     <router-link
       class="appointment-logo"
@@ -142,17 +136,7 @@ const copyLink = async () => {
   display: none;
 }
 
-.header-mobile {
-  height: 64px;
-  background-color: var(--colour-neutral-base);
-}
-
 @media (--md) {
-  /* Hide header mobile in favor of desktop */
-  .header-mobile {
-    display: none;
-  }
-
   .header-desktop {
     position: fixed;
     display: flex;
--- a/frontend/src/components/NavBar.vue
+++ b/frontend/src/components/NavBar.vue
@@ -51,34 +51,21 @@ const copyLink = async () => {
 </script>
 
 <template>
-  <header
-    class="
-      fixed z-50 flex h-16 w-full justify-between overflow-x-auto overflow-y-visible border-b border-gray-300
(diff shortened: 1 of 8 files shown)


frontend/src/components/NavBar.vue
```diff
@@ -51,62 +51,43 @@ const copyLink = async () => {
...
-          stroke-gray-300 dark:stroke-gray-500
-        "
-      />
-      <input
-        class="w-full h-full text-xl pl-20 pr-2 border-none"
-        type="search"
-        name="search"
-        :placeholder="t('label.search')"
-      />
-    </label> -->
```
MelissaAutumn, line 79: Rip search. You never worked but I was always hopeful you someday might. :cry:

frontend/src/components/NavBarMobile.vue
```diff
@@ -0,0 +1,300 @@
...
+  { route: 'availability', i18nKey: 'availability', icon: IconCalendarClock },
+  { route: 'settings', i18nKey: 'settings', icon: IconSettings },
+];
+
+const menuOpen = ref(false);
+const myLinkTooltip = ref(t('navBar.shareMyLink'));
+
+function onMenuOpen() {
+  menuOpen.value = true;
+  document.body.style.overflow = 'hidden'
```
MelissaAutumn, line 38: Why do we need overflow set here?

MelissaAutumn (approved):
Looks good, just confused on the overflow.

frontend/src/components/NavBar.vue
davinotdavid: RIP search 🫡

One day it will make a comeback though, I am sure!

frontend/src/components/NavBarMobile.vue
davinotdavid: Ah, if we are in pages that have vertical scroll in it, the scrollbar shows and its content is scrollable

Aug-25-2025 10-52-15

MelissaAutumn: Ahh I see. Can you throw a comment on it? Thanks!

davinotdavid: Sure! Added, thanks!

Issue #1173, opened 2025-08-11
SQLAlchemy + psycopg fails to compare boolean with integer on Availability page


davinotdavid:
**Describe the bug**
After the migration from MySQL to Postgresql using the `psycopg` adapter, we lost the ability to implicitly convert boolean / integer so comparisons like `some_boolean == 1` or `some_boolean == 0` doesn't work anymore.

```
 File "/app/appointment/routes/schedule.py", line 207, in read_schedule_availabilities
    calendars = repo.calendar.get_by_subscriber(db, subscriber.id, False)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

sqlalchemy.exc.ProgrammingError: (psycopg.errors.UndefinedFunction) operator does not exist: boolean = integer
LINE 3: ...rs.owner_id = $1::INTEGER AND calendars.connected = $2::INTE...
                                                             ^
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
```

**To Reproduce**

1. Go through the FTUE and **connect a Zoom account**
2. Copy your appointment link and paste it in a Private window
3. See error in the backend docker container / console / see the infinite loading

**Expected behavior**
Availability schedule should be loaded correctly.

**Actual behavior**
Infinite loading of death

davinotdavid:
Closed since PR has been merged

PR #1174, opened 2025-08-11
Fix psycopg and sqlalchemy comparison fail between boolean and integer

davinotdavid:
## Description of the Change

We no longer may compare boolean and integers while using sqlalchemy + psycopg. This breaks the inner workings of the SQL layer translation and requires an explicit cast.

```
 File "/app/appointment/routes/schedule.py", line 207, in read_schedule_availabilities
    calendars = repo.calendar.get_by_subscriber(db, subscriber.id, False)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

sqlalchemy.exc.ProgrammingError: (psycopg.errors.UndefinedFunction) operator does not exist: boolean = integer
LINE 3: ...rs.owner_id = $1::INTEGER AND calendars.connected = $2::INTE...
                                                             ^
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
```

## Benefits

- Booker availability page loads correctly!

## Applicable Issues

https://github.com/thunderbird/appointment/issues/1173

### DIFF
--- a/backend/src/appointment/database/repo/calendar.py
+++ b/backend/src/appointment/database/repo/calendar.py
@@ -46,7 +46,9 @@ def get_by_subscriber(db: Session, subscriber_id: int, include_unconnected: bool
     query = db.query(models.Calendar).filter(models.Calendar.owner_id == subscriber_id)
 
     if not include_unconnected:
-        query = query.filter(models.Calendar.connected)
+        # Skipping ruff check here since we need to compare a boolean to an integer in Postgresql
+        # and "models.Calendar.connected is True" always evaluates to false.
+        query = query.filter(models.Calendar.connected == True)  # noqa: E712
 
     return query.all()
 


davinotdavid:
> Would `.filter(models.Calendar.connected is True)` work also?

@MelissaAutumn not really. When I tried locally, it doesn't break in the DB query level as it was breaking before but it fails to match the the correct values in the query. Here's the difference in the queries underneath when I stringify it:

[.filter(models.Calendar.connected)]
```
SELECT [...]
FROM [...]
LEFT OUTER JOIN [...]

WHERE calendars.owner_id = %(owner_id_1)s::INTEGER AND calendars.connected
```

[.filter(models.Calendar.connected is True)]
```
SELECT [...]
FROM [...]
LEFT OUTER JOIN [...]

WHERE false
```

MelissaAutumn:
Last one: How about == True? Just trying to remove truthy where possible.

davinotdavid:
> Last one: How about == True? Just trying to remove truthy where possible.

@MelissaAutumn  That works 🎉  It interprets as:

[.filter(models.Calendar.connected == True)]
```
SELECT [...]
FROM [...]
LEFT OUTER JOIN [...]

WHERE calendars.owner_id = %(owner_id_1)s::INTEGER AND calendars.connected = true
```

The only gotcha is that `Ruff` gets a little upset and the auto-fix suggests to do `is True` which can't be done here. Would you prefer to have Ruff disabled for this line and keeping the `== True` in this case + a comment?

image

MelissaAutumn:
Forgot to reply this morning. Yep skipping ruff check here is fine. Thanks!!

PR #1175, opened 2025-08-12
Cache-control on external connections / Filter out unnecessary fields on Slot-based responses

davinotdavid:
## Description of the Change

- Adds a 'no-cache' [FastAPI metadata tag](https://fastapi.tiangolo.com/tutorial/metadata/#create-metadata-for-tags) and a middleware to add a response header `Cache-control: no-store` for tagged requests.

- Adds the 'no-cache' header for several endpoints:
  - [GET /me/appointments](https://github.com/thunderbird/appointment/pull/1175/files#diff-09299ef06274d0c8e9e54ca9f4f976ca279727291549e451ea0206a9e181dbfdR104) since it returns slot information that contains `meeting_link_id` / `meeting_link_url` in it.
  - [GET /rmt/cal/:id](https://github.com/thunderbird/appointment/pull/1175/files#diff-09299ef06274d0c8e9e54ca9f4f976ca279727291549e451ea0206a9e181dbfdR388) since it returns slot information that contains `meeting_link_id` / `meeting_link_url` in it and after booking an appointment with Zoom, the Zoom link is present in the string description of the appointment.
  - [GET /account/external-connections](https://github.com/thunderbird/appointment/pull/1175/files#diff-5cd18dac1390a742f2891a578efa39be254d94c7ac44ca13e7c17494604f5793R25) since it returns `type_id` and that may be associated with sensitive information on external connections.
  - [PUT /schedule/public/availability/booking](https://github.com/thunderbird/appointment/pull/1175/files#diff-1d810ba22035665e8ce68c193c11abbec1fdc754452ef1f1304f9b0592604c02R455), just in case even though we are not returning sensitive data in these endpoints anymore.
  - [POST /schedule/public/availability](https://github.com/thunderbird/appointment/pull/1175/files#diff-1d810ba22035665e8ce68c193c11abbec1fdc754452ef1f1304f9b0592604c02R178), just in case even though we are not returning sensitive data in these endpoints anymore.
  - [PUT /schedule/public/availability/request](https://github.com/thunderbird/appointment/pull/1175/files#diff-1d810ba22035665e8ce68c193c11abbec1fdc754452ef1f1304f9b0592604c02R241), just in case even though we are not returning sensitive data in these endpoints anymore. 

- On the frontend, fixed a small bug where we were reading the `meeting_link_url` from the `location_url`

- On the backend, changed the `SlotOut` schema  to return only essential information

## Benefits

- Hardened security for potential vulnerabilities on sensitive meeting / external connection information

## Applicable Issues

https://github.com/thunderbird/private-issue-tracking/issues/18
https://github.com/thunderbird/appointment/pull/1174

### DIFF
--- a/backend/src/appointment/database/schemas.py
+++ b/backend/src/appointment/database/schemas.py
@@ -84,8 +84,12 @@ class Slot(SlotBase):
     attendee: Attendee | None = None
 
 
-class SlotOut(SlotBase):
+class SlotOut(BaseModel):
     id: int | None = None
+    attendee_id: int | None = None
+    booking_status: BookingStatus | None = BookingStatus.none
+    duration: int | None = None
+    start: datetime
 
 
 class SlotAttendee(BaseModel):
@@ -143,7 +147,7 @@ class AppointmentWithCalendarOut(Appointment):
 class AppointmentOut(AppointmentBase):
     id: int | None = None
     owner_name: str | None = None
-    slots: list[SlotBase | SlotOut] = []
+    slots: list[SlotOut] = []
     slot_duration: int
     booking_confirmation: bool
 
@@ -398,11 +402,6 @@ class InviteAdminOut(ListResponse):
 
 """ other schemas used for requests or data migration
 """
-
-
-class AppointmentSlots(BaseModel):
-    appointment: AppointmentFull
-    slots: list[SlotBase] = []
 
 
 class AppointmentModifyRequest(BaseModel):
--- a/backend/src/appointment/main.py
+++ b/backend/src/appointment/main.py
@@ -227,7 +227,7 @@ def server():
         """Adds cache-control: no-store response header for tagged requests"""
         response = await call_next(request)
         if request.scope.get('route') and 'no-cache' in request.scope['route'].tags:
-            response.headers['Cache-Control'] = 'no-store'
+            response.headers['cache-control'] = 'no-store'
         return response
 
     @app.middleware('http')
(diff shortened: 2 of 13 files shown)


backend/src/appointment/database/schemas.py
```diff
@@ -400,11 +404,6 @@ class InviteAdminOut(ListResponse):
 """
 
 
-class AppointmentSlots(BaseModel):
-    appointment: AppointmentFull
-    slots: list[SlotBase] = []
-
-
```
davinotdavid, line 407: Not being used

davinotdavid:
@MelissaAutumn TIL, I didn't know about tags and using the middleware makes a lot more sense. Thank you so much for the suggestion / guidance! Changed the implementation to use tags

MelissaAutumn (changes requested):
Looks good, only request is adding server tests for the tag middleware. You can probably test health route and a random post route vs the no-cache tagged routes. 

Thanks!

davinotdavid:
@MelissaAutumn Thanks for reviewing! For sure, added tests!

davinotdavid:
@MelissaAutumn sorry if it is a dumb question but do you mean the the key `Cache-Control` or the `no-store` value?

as in, would

`assert response.headers.get('Cache-Control').lower() == 'no-store'`

be ok?

MelissaAutumn:
Sorry I meant case insensitive for the key, since dicts are case sensitive. (Although I'm not sure if headers is a dict or a dict-like object :thinking: )

davinotdavid:
@MelissaAutumn interesting, I did some digging and it seems like request headers are case insensitive in the [Starlette docs](https://www.starlette.io/requests/) and in the [HTTP RFC 7230](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2). However, interestingly in HTTP/2 it seems that it should be lowercased in the [RFC 7540](https://www.rfc-editor.org/rfc/rfc7540#section-8.1.2) 🤔

Seems like it is better to have it lowercased then? Last commit changes them all to`cache-control`

MelissaAutumn (approved):

Issue #1176, opened 2025-08-12
Postgresql container in docker-compose FATAL: role 'root' does not exist


davinotdavid:
**Describe the bug**
Even though the application works correctly, the postgresql container keeps erroring with `FATAL: role 'root' does not exist`

**To Reproduce**
Run `docker compose up -d --build` and see the logs for `postgresql` container.

**Expected behavior**
No errors

davinotdavid:
Closing as PR has been merged

PR #1177, opened 2025-08-12
Update compose.yml file regarding postgresql container health check

davinotdavid:
## Description of the Change

The health check command done in the postgresql container in Docker Compose attempts to use the role `root` by default but that role doesn't exist. This PR updates the command to actually check the DB health using the user and DB in the env vars.

## Benefits

No more `FATAL: role 'root' does not exist` in the docker container console output.

## Applicable Issues

https://github.com/thunderbird/appointment/issues/1176


rwood-moz (approved):
Verified locally, postgresql container no longer reporting the error. Thanks!

PR #1180, opened 2025-08-13
Set specific playwright version to fix E2E BrowserStack test jobs

rwood-moz:
The E2E tests run fine locally but recently they started failing when they run in BrowserStack. In BrowserStack the session just hangs and the E2E tests timeout as the browsers don't even startup.

The E2E BrowserStack test jobs just use the newest playwright and browser versions however after upgrading to the latest playwright browsers the tests won't run in BrowserStack.

After some investigation the solution is to use a specific playwright version on the client side (test/e2e/package.json) and also tell BrowserStack to use this same playwright version (browserstack-desktop.yml); and use the latest version that BrowserStack specifies in their [documentation here](https://www.browserstack.com/docs/automate/playwright/browsers-and-os?fw-lang=nodejs) which is `1.53.0`. I tried this out, and using this playwright version fixes the BrowserStack issue and the E2E tests startup and run.

### DIFF
--- a/test/e2e/browserstack-desktop.yml
+++ b/test/e2e/browserstack-desktop.yml
@@ -80,6 +80,7 @@ debug: false # <boolean> # Set to true if you need screenshots for every seleniu
 networkLogs: false # <boolean> Set to true to enable HAR logs capturing; off as may contain sensitive info like login API requests
 consoleLogs: info # <string> Remote browser's console debug levels to be printed (`disable`, `errors`, `warnings`, `info`, or `verbose`)
 framework: playwright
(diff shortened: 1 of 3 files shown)


rwood-moz:
I ran the nightly E2E tests job in BrowserStack against this branch and the browsers now startup fine and the tests run, BrowserStack [link here](https://automate.browserstack.com/dashboard/v2/builds/742fb83285c966d31d9d76a0c768f12e57794fbd).

Note that there are some E2E test failures but they are unrelated; they are because the tests were updated for the latest UI changes (and pass locally) but the changes haven't been deployed to production yet; after the next production deployment the E2E tests against production (nightly test job) will pass again.

PR #1185, opened 2025-08-14
Fix deploys

ryanjjung:
This restores a piece of code that we still need to operate our old stage environment. If the `DATABASE_SECRETS` variable is set, we will set the `DATABASE_URL` variable, which then triggers an admonishment of the configuration and the reluctant carrying out of the old way of configuring the DB.

### DIFF
--- a/backend/src/appointment/utils.py
+++ b/backend/src/appointment/utils.py
@@ -185,6 +185,32 @@ def normalize_secrets():
     """Normalizes AWS secrets for Appointment"""
 
     log.info('Normalizing secrets...')
+
+    database_secrets = os.getenv('DATABASE_SECRETS')
+    if database_secrets:
+        secrets = json.loads(database_secrets)
+        host = secrets['host']
+        port = secrets['port']
+
+        # If port is not already in the host var, then append it to hostname
+        hostname = host
+        if f':{port}' not in host:
+            hostname = f'{hostname}:{port}'
+
+        # Determine a "dialect+driver" scheme
+        dialect = secrets['engine']
+        if dialect == 'mysql':
+            driver = 'mysqldb'
+        elif dialect == 'postgresql':
+            driver = 'psycopg2'
+        else:
+            driver = None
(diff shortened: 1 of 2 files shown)


backend/src/appointment/utils.py
```diff
@@ -176,6 +185,32 @@ def normalize_secrets():
     """Normalizes AWS secrets for Appointment"""
 
     log.info('Normalizing secrets...')
+
+    database_secrets = os.getenv('DATABASE_SECRETS')
```
MelissaAutumn, line 190: If `get_database_url` runs after `normalize_secrets` I think we can re-wire them up to the new env vars. Like 

```suggestion
    database_secrets = os.getenv('DATABASE_SECRETS')
    if database_secrets:
        secrets = json.loads(database_secrets)
        os.environ['DATABASE_HOST'] = secrets['host']
        ... etc ...
```

backend/src/appointment/utils.py
```diff
@@ -157,8 +157,17 @@ def get_database_url() -> str | sqlalchemy_url:
...
-        raise ValueError('Missing one or more database configuration value. Review your environment.')
+    requirements = {
+        'db_name': db_name,
+        'dialect': dialect,
+        'host': host,
+        'password': password,
+        'port': port,
+        'username': username,
+    }
+    missing_requirements = [key for key, value in requirements.items() if not value]
```
MelissaAutumn, line 168: Gooood idea.

backend/src/appointment/utils.py
MelissaAutumn: Then we can not worry about database urls anymore, and just focus on the newer better more string escaped method of forming a db url.

ryanjjung: It should. normalize_secrets runs out of __main__ on both API startup and CLI command execution. get_database_engine_and_session is when the various `DATABASE_*` settings get used. I'll see if I can wire that up a bit better.

Issue #1093, opened 2025-07-08
Replace Qalendar and add calendar refresh/last updated info


malini:
From the updated UI wireframes:

<img width="1281" height="727" alt="Image" src="https://github.com/user-attachments/assets/7a92d870-740d-472d-8cfb-c225a1c7c044" />

We're simplifying the design of the calendar, so we can remove Qalendar as it's unsupported now, and Aadd the last updated field and a refresh button

devmount:
For reference #762 (can be closed in favor of this)

MelissaAutumn:
We'll probably want to make our own component just for the week view. For super small sizes it can horizontally scroll for now.

davinotdavid:
PRs have been merged and it is live on staging!

PR #1188, opened 2025-08-18
Replace Qalendar on Dashboard and add calendar refresh/last updated info

davinotdavid:
## Description of the Change

There are three main parts of this change so here are the highlights of each change:

1. ### WeekPicker
- This is a week picker that hides an `<input type="date">` underneath for selecting a date. We _could_ try to adapt and reuse the `CalendarMiniMonth` component but I reckon that it was used for a different purpose and we could rely on the browser defaults to get some goodies for free (e.g. accessibility). Since it was not specified in the Wireframes that the design should follow a specific look, I believe this is for the best as of now.
- Keyboard arrows change the week forwards and backwards.

2. ### Calendar / Remote Events Sync
- This is using the first schedule's calendar. ~~Since it is currently impossible (AFAIK at least) to have a disconnected calendar as your first schedule's calendar, it doesn't have any logic to show a label other than `Connected` there~~
- Clicking the `sync` icon button re-fetches the calendars' events and resets the timer.
- Timer is being automagically handled by the [useTimeAgo](https://vueuse.org/core/useTimeAgo/) composable from @vueuse package that we are already using. I had to update the version because this is very new and it also handles the i18n automagically.

3. ### WeekCalendar 
- This is the big one. It is done with CSS Grid and the time slot placements of the events are done by calculating their row / column combination in their respective descriptive functions. Sounds intuitive but it was surprisingly tricky to get i18n weekday placements correctly ("de" starts on a Monday but "en" starts on a Sunday, for instance).
- Hovering the event now calculates the best popup / tooltip placement according to whether it fits on the screen or not.
- There is an attempt of making the calendar responsive-ish. It doesn't look broken but it also doesn't look good. Need to check with the design team on what to do here.

Wireframes
Appointment_MVP_Dashboard

Light mode
image

Dark mode
image

## Benefits

- Better UX according to the Design team
- Better control over calendar styling
- Fix for event hover popup cropping and / or causing horizontal scroll
- Fairly OK responsiveness

- All `data-testids`, sorry Rob :(
- This replaces Qalendar in the context of the Dashboard but it doesn't remove it from the application completely as it is still being used in the Booker view. This will be extracted and prop-ed in that task.
- I see that we have a `CalendarMiniMonth` and `CalendarMiniMonthDay` components that are currently being unused (the only reference seems to be the `AppointmentModifyPanel` which is a feature currently TBD). Perhaps we can 🪓  these components?
- Hovering the event on mobile does nothing at the moment since we can't actually hover on mobile (heh). Need to check with design in a follow up ticket to know what we want to do with this.

## Applicable Issues

https://github.com/thunderbird/appointment/issues/1093

### DIFF
--- a/frontend/src/locales/de.json
+++ b/frontend/src/locales/de.json
@@ -271,6 +271,7 @@
     "denyBooking": "Buchung ablehnen",
     "disable": "Deaktivieren",
     "disconnect": "Trennen",
+    "disconnected": "Getrennt",
     "displayName": "Anzeigename",
     "download": "Download",
     "downloadInvitation": "Termineinladung herunterladen",
--- a/frontend/src/locales/en.json
+++ b/frontend/src/locales/en.json
@@ -270,6 +270,7 @@
     "denyBooking": "Deny booking",
     "disable": "Disable",
     "disconnect": "Disconnect",
+    "disconnected": "Disconnected",
     "displayName": "Display name",
     "download": "Download",
     "downloadInvitation": "Download invitation",
--- a/frontend/src/views/DashboardView/components/UserCalendarSync.vue
+++ b/frontend/src/views/DashboardView/components/UserCalendarSync.vue
@@ -49,14 +49,19 @@ onMounted(async () => {
 
 <template>
   <div class="user-info-container">
-    <div>
-      <p>{{ t('label.connected') }} - <strong>{{ scheduleCalendar?.user }}</strong></p>
-      <p>{{ t('label.updated') }} {{ timeAgo }}</p>
-    </div>
+    <template v-if="scheduleCalendar?.connected">
+      <div>
+        <p>{{ t('label.connected') }} - <strong>{{ scheduleCalendar?.user }}</strong></p>
+        <p>{{ t('label.updated') }} {{ timeAgo }}</p>
+      </div>
 
-    <button @click="onSyncCalendarButtonClicked" :disabled="loading" :class="{ 'loading': loading }">
-      <icon-reload size="20"/>
-    </button>
+      <button @click="onSyncCalendarButtonClicked" :disabled="loading" :class="{ 'loading': loading }">
+        <icon-reload size="20"/>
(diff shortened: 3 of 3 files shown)


MelissaAutumn:
image

Nice! I'm seeing some odd default component state, at least on dark mode. This is a brand new account with a connected but empty calendar.

MelissaAutumn:
Ah for some reason by default on Aug 20th 2025 it's pulling September 2025 3 times instead. 

image

davinotdavid:
@MelissaAutumn iiiinteresting, I didn't come across the empty calendar use case but I can see it happening in case we don't have the first Schedule set (I could reproduce this only in during the FTUE, behind the setup modal).

I am now loading default time slots / time duration so that the rows appear correctly regardless (https://github.com/thunderbird/appointment/pull/1188/commits/e0dc53da0201da74175460621eddefe7ccb30bf6).

Speaking of FTUE, also changed the UserSync component to show just `Disconnected` (https://github.com/thunderbird/appointment/pull/1188/commits/1823cb90513c3c346e0f96b81e8b51b6c1725cbd).

> Ah for some reason by default on Aug 20th 2025 it's pulling September 2025 3 times instead.

About this one, to reduce the pulls, I've left the previous fetch implementation since we were already doing it monthly (so that September 2025 request is actually from the period of August 1st to September 1st, as a `GET` to `/rmt/cal/1/2025-08-01/2025-09-01`) and it is done per calendar (I assume that you have 3 calendars connected?).

MelissaAutumn:
Ah yes the 3 calendars connected would do it. :sweat_smile: 

Will re-review in a bit!

MelissaAutumn:
Okay I found a separate bug. This looks fine now, I'll have a patch up for the one I ran into in a min.

davinotdavid:
@MelissaAutumn awesome, thanks! Perhaps we can punch it into another ticket for tracking? I will merge this one for now then!

Issue #1196, opened 2025-08-25
Re-add ability to download your data


davinotdavid:
**Is your feature request related to a problem? Please describe.**
As part of the new Settings UX Overhaul, we didn't capture the pre-existing functionality of being able to download your data. This should be added back.

**Describe the solution you'd like**

<img width="893" height="511" alt="Image" src="https://github.com/user-attachments/assets/18e8a0ec-744d-4ba4-9955-184958b31176" />

As part of this, we need to ask for the user to re-authenticate before downloading.

**Additional context**
The download button was previously removed as part of https://github.com/thunderbird/appointment/issues/1096

davinotdavid:
This has been deployed to prod!

PR #1198, opened 2025-08-25
Re-add the download data flow as it was before

davinotdavid:
## Description of the Change

This PR adds back the download data feature as it was before the UX overhaul while re-asking for authentication before the download starts.

Zeplin wireframe
image

Implementation

https://github.com/user-attachments/assets/0ef18036-fb56-4bc9-9634-dded05cf3908

## Benefits

- Re-add ability for users to download their data

## Applicable Issues

https://github.com/thunderbird/appointment/issues/1196


frontend/src/stores/user-store.ts
```diff
@@ -324,6 +324,22 @@ export const useUserStore = defineStore('user', () => {
...
+    formData.set('password', password);
+    const { error, data: tokenData }: TokenResponse = await call.value('token').post(formData).json();
+
+    if (error.value || !tokenData.value.access_token) {
+      return { error: tokenData.value ?? error.value };
+    }
+
+    data.value.accessToken = tokenData.value.access_token;
+
+    return { error: null };
```
MelissaAutumn: This will only actually work with the username, password auth meant for self-hosting / development work. We don't store a password in the backend for fxa (deprecated) or oidc (keycloak) auth. 

I think you'll need set a flag in localStorage (so you can detect the user is returning from a download request login) and then trigger a keycloak login request via signInRedirect({prompt: 'login'}). 

You'll need to do something similar for FXA (that handles oauth over the backend via redirects), as we'll need this to work before we switch servers. :disappointed: 

Lmk if you need help with this.

davinotdavid: Cool, definitely need help with setting up a testing environment for both `fxa` and `oidc` as my attempts of connecting to the stage env failed :( I will send a message to you to sync tomorrow or whenever you are available!

Also, IIRC we didn't have a re-auth part to download the data before. I don't think that this is super urgent so we can definitely work this flow out but just wandering if it would be best to have something shipped without re-auth first.

frontend/src/views/SettingsView/components/AccountSettings.vue
```diff
@@ -23,12 +23,12 @@ const copyLinkTooltip = ref(t('label.copyLink'));
 const cancelAccountModalOpen = ref(false);
 const downloadAccountModalOpen = ref(false);
 
-const closeModals = () => {
+function closeModals() {
```
MelissaAutumn: Did our linter mess up, we tend to use arrow functions for reasons that are mostly just  @devmount and I programming it that way. :sweat_smile:

frontend/src/stores/user-store.ts
MelissaAutumn: Sounds good, we can skip re-auth for now. Can you punt it to another ticket?

frontend/src/views/SettingsView/components/DownloadDataModal.vue
```diff
@@ -0,0 +1,141 @@
...
+const loginError = ref<Alert>(null);
+
+const props = defineProps<{
+  onLoginSuccessful: () => void,
+}>();
+
+async function onSubmit() {
+  isLoading.value = true;
+
+  if (isOidcAuth) {
```
MelissaAutumn: I must have commented on the wrong function, but feel free to punt this to another tickets so we can bring this functionality back.

frontend/src/views/SettingsView/components/AccountSettings.vue
davinotdavid: Oh sorry, I don't think that is covered by the linter atm but I can add a hard rule there of course!

I personally prefer to use the `function` keyword as it is visually easier to distinct from other constants that are not functions / non-callable bits and tend to use arrow functions for callbacks but of course happy to adjust for better code standards!

frontend/src/stores/user-store.ts
davinotdavid: Great, thanks! Punt it here: https://github.com/thunderbird/appointment/issues/1208

PR #1199, opened 2025-08-25
Overhaul the settings E2E tests to reflect the new settings page

rwood-moz:
Now that the Appointment settings page has been overhauled nicely, this PR overhauls the E2E tests for the new settings. Tested against local stack, stage, and prod. BrowserStack link can be [found here](testing/settings-page-e2e-overhaul) showing all of the tests passing (against prod) with this branch.


test/e2e/pages/settings-page.ts
```diff
@@ -1,199 +1,162 @@
...
   }
 
   /**
-   * Navigate directly to the account settings page
+   * Navigate to settings, preferences section
    */
-  async gotoAccountSettingsPage() {
-    await this.page.goto(APPT_ACCOUNT_SETTINGS_PAGE);
+  async gotoPreferencesSettings() {
+    await this.page.goto(APPT_SETTINGS_PAGE);
```
davinotdavid, line 101: I see that we have defined specific page sections as constants here: https://github.com/thunderbird/appointment/pull/1199/files#diff-9329dd4646721e7bcfe42d3fbc26e5ba895f28b865e9031448408e954c904ee0R12-R14

Can't we directly use `.goto` using the pre-defined consts and skip the scrolls and `.waitForTimeout` to speed up the test navigation?

test/e2e/pages/settings-page.ts
```diff
@@ -1,199 +1,162 @@
...
   }
 
   /**
-   * Navigate directly to the connected accounts settings page
+   * Navigate to settings, connected applications section
    */
-  async gotoConnectedAccountsSettingsPage() {
-    await this.page.goto(APPT_CONNECTED_SETTINGS_PAGE);
+  async gotoConnectedAppSettings() {
+    await this.page.goto(APPT_SETTINGS_PAGE);
```
davinotdavid, line 112: Same here

test/e2e/pages/settings-page.ts
```diff
@@ -1,199 +1,162 @@
...
   }
 
   /**
-   * Navigate directly to the calendar settings page
+   * Navigate to settings, account settings section
    */
-  async gotoCalendarSettingsPage() {
-    await this.page.goto(APPT_CALENDAR_SETTINGS_PAGE);
+  async gotoAccountSettings() {
+    await this.page.goto(APPT_SETTINGS_PAGE);
```
davinotdavid, line 90: Same here

davinotdavid (approved):
Nice! Thanks for working on the e2e here after the overhauls 🙏 

Pre-approving here as I don't want to hang this PR over my comments here but it would be nice to reduce some of the timeouts to speed up the testing process whenever we can.

test/e2e/pages/settings-page.ts
rwood-moz: Thanks for the review, appreciated!

Hey, that's a great question! The reason why I didn't go to the settings page sections URLs directly is because I want the E2E tests to perform as an end user would, and our end-users will click on the section buttons and not change the URL. This way the section link buttons will also get exercised.

About the scroll into views, the reason these are here is for when I port these tests over to mobile browsers, when only pats of the pages are displayed it's best to scroll to the area we are exercising (especially so we have screenshots of the areas being tested if tests fail).

davinotdavid: Ah makes sense, if that's the best practice here perhaps we can delete the unused constants to avoid misuse in the future?

rwood-moz: > Ah makes sense, if that's the best practice here perhaps we can delete the unused constants to avoid misuse in the future?

Good idea, I'll do that in my next PR that I am currently working on 👍

PR #1200, opened 2025-08-26
[Hotfix?] First schedule may not exist when accounts are disconnected / reconnected

davinotdavid:
## Description of the Change

This PR fixes a few problems but also opens up new questions on how to handle some scenarios:

1. In the Settings page, we were coupling the calendar updates with the first / default schedule since the `default` calendar can be changed in this page to any other connected calendar. This is a problem because if a user removes all of the accounts / has no calendars, they would also have no schedule anymore. [Adding a check here solves it](https://github.com/thunderbird/appointment/pull/1200/files#diff-f09242ff792588c8b0041051932ba73857622a2d2057991ca39ae2db73c09e12R139-R143). [I've also added a placeholder message](https://github.com/thunderbird/appointment/pull/1200/files#diff-fdd0b20f6fd56a7487b6aa97d100fe585694f85bd7db1c531ee100dc8a8d942aR247) (copy not approved yet) for the empty state.

2. After the FTUE, we were also not accounting for the possibility of not having a first / default schedule at all. This means that the Availability page was only doing an update when clicking `save` instead of being able to create a schedule. Now, [if there is no schedule id, we are creating a schedule instead](https://github.com/thunderbird/appointment/pull/1200/files#diff-69283f1479f03baa0f0bd773b09e7a82cf9621d38cca5941b0d44d16777204c5R72-R75) using pre-populated default values (similar logic as the FTUE).

3. Booking page slug was not working as expected / not reflecting changes instantly. We are syncing backend data with local storage here and there were some caching problems. Now both the auto-generated refreshed slug and the custom one get synced correctly.

> [!IMPORTANT]  
> The core issue that started it all is still not fixed (accepting booked appointments is still failing the `verify_link`) but this PR should fix the "blank screen of death"

## Testing
- After going through the FTUE, go to the Settings page and disconnect all your calendar accounts.
- See that there is an empty state message `**[NEW]**`
- Connect a Google Account (haven't tested with CalDav)
- See that there are no connected calendars. Click on the checkbox next to the calendar and click "Save" to apply the change. `**[NEW - This was failing before per point number 1 described above]**`
- Go to the Availability page, toggle the "You're not bookable" switch and select your connected calendar. Note that the schedule settings here has default values. Click on "Save" to apply the change. At this point a schedule will be created instead of being updated. `**[NEW - This did not exist before]**`
- Still in the Availability page, customize your booking page slug to any text and hit "Save". The change should be reflected instantly. `**[NEW - This was not working before]**`
- Still in the Availability page, click on the "refresh" icon in the booking page slug and confirm on the modal. The change should be reflected instantly `**[NEW - This was not being reflected instantly before and required a page refresh]**`

## Benefits

- App works again and folks can disconnect all accounts and upon re-connection of one, it would create a schedule.
- Slugs edits are reflected correctly.

## Applicable Issues

https://github.com/thunderbird/appointment/issues/1201

### DIFF
--- a/frontend/src/locales/de.json
+++ b/frontend/src/locales/de.json
@@ -603,6 +603,9 @@
             "title": "Zoom trennen"
           }
         }
+      },
+      "calendars": {
+        "noCalendars": "Keine Kalenderkonten verbunden. Füge ein Konto hinzu, um zu beginnen."
       }
     },
     "timesAreDisplayedInLocalTimezone": "Die Zeiten werden in deiner lokalen Zeitzone {timezone} angezeigt.",
--- a/frontend/src/locales/en.json
+++ b/frontend/src/locales/en.json
@@ -602,6 +602,9 @@
             "title": "Disconnect Zoom"
           }
         }
+      },
+      "calendars": {
+        "noCalendars": "No calendar accounts connected. Add an account to get started."
       }
     },
     "timesAreDisplayedInLocalTimezone": "Times are displayed in your local timezone {timezone}.",
--- a/frontend/src/models.ts
+++ b/frontend/src/models.ts
@@ -205,7 +205,7 @@ export type AvailabilityFormFields = {
   end_time?: string;
   farthest_booking?: number;
   id?: number;
-  link_slug?: string;
+  slug?: string;
   location_url?: string;
   meeting_link_provider?: string;
   name?: string;
--- a/frontend/src/stores/availability-store.ts
+++ b/frontend/src/stores/availability-store.ts
@@ -75,7 +75,7 @@ export const useAvailabilityStore = defineStore('availability', () => {
     hasZoom.value = !!externalConnectionStore.zoom[0];
 
     // Booking Page Link data
-    initialState.value.link_slug = userStore.mySlug;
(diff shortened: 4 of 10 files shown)


rwood-moz:
When you don't have any calendars connected, on the Settings => Connected Applications section it does say `No calendar accounts connected. Add an account to get started.` which is great. However if you're on the `Availability` tab/page it's not made clear that you need to connect a calendar i.e. the set your availability section options are greyed out (including the 'you're not bookable' slider) but it doesn't indicate why you aren't able to change any of those settings or become bookable. Maybe somewhere there it should also indicate that you need to connect a calendar first? Just my opinion :)

MelissaAutumn (approved):

davinotdavid:
@rwood-moz ah I agree, the greyed out / disabled controls without any feedback is something that is probably going to be changed ... including the overall error handling / form validation of all the availability form fields (currently it is quite wacky).

The empty state message for the calendar was added by me including copy so I am sure that these will be prone to be changed soon too. I will leave that for a next ticket after design give their considerations (as in, should it be tooltips / alert info boxes / should it have a different UI / flow altogether, etc). Good point though!

Issue #1049, opened 2025-06-03
Can't accept/decline bookings


MelissaAutumn:
On stage, I can't seem to accept or decline bookings.

MelissaAutumn:
This was caused by https://github.com/thunderbird/appointment/pull/994 (It's been a while since we could deploy 😅) 

We are erronously signing the short url and then checking that signature against a signed long url. I didn't catch it because we don't have short urls setup on local dev. 

Looking into a fix, but it's a little complicated.

devmount:
Thanks for the fix @MelissaAutumn! Maybe we can find a way to automatically test this or set up short urls in our dev envs too.

PR #1207, opened 2025-08-27
Update verify_link to always compare with long urls

davinotdavid:
## Description of the Change

When we have `SHORT_BASE_URL` env var set, we are returning a signed url for the user with a `short_url`. This gets stored in frontend's user data and it is currently used to confirm / decline bookings. This is a problem because the verification of the signed url always compares it with the long url and not the short ones.

This used to work before because the frontend was sending **`window.location.href` while being in a route that had the long url** but it is now broken because, instead, we are now sending the `signedUrl` back to the backend for verification (and the initial signage of the url was using the short url and not the long one).

- On the backend, if the `SHORT_BASE_URL` env var is set, we are now replacing it with the long url upon link verification so that we are comparing long urls with long urls.

- Removed previous print debug statements

## Benefits

- Folks who have the `SHORT_BASE_URL` env var set (like us, in stage/prod) will be able to confirm / decline bookings again

## Applicable Issues

https://github.com/thunderbird/appointment/issues/1049

### DIFF
--- a/backend/src/appointment/database/repo/subscriber.py
+++ b/backend/src/appointment/database/repo/subscriber.py
@@ -5,6 +5,7 @@ Repository providing CRUD functions for subscriber database models.
 
 import datetime
 import secrets
+import os
 
 from sqlalchemy.orm import Session
 from .. import models, schemas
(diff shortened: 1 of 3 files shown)


backend/src/appointment/database/repo/subscriber.py
```diff
@@ -132,25 +133,21 @@ def verify_link(db: Session, url: str):
     """Check if a given url is a valid signed subscriber profile link
     Return subscriber if valid.
     """
-    print(f'==== verify_link called with {url} ====')
 
-    username, signature, clean_url = utils.retrieve_user_url_data(url)
+    if os.getenv('SHORT_BASE_URL'):
+        # If the url is a short url, we need to reconstruct the long url
+        # otherwise it will fail the signature check as we always sign with the long url
+        url = url.replace(os.getenv('SHORT_BASE_URL'), f'{os.getenv("FRONTEND_URL")}/user')
```
MelissaAutumn: We should define frontend_url/user as a constant somewhere and pull that in.

backend/test/unit/test_subscriber_repo.py
```diff
@@ -0,0 +1,26 @@
+import os
+
+from appointment.database.repo.subscriber import verify_link
+from appointment.controller.auth import signed_url_by_subscriber
+
+
+class TestSubscriber:
+    def test_verify_link_uses_long_url(self, with_db, make_basic_subscriber):
```
MelissaAutumn, line 8: Can we add another test that uses the BASE_URL and ensure that it doesn't change?

davinotdavid: Sure, moved it to `defines.py`!

backend/src/appointment/database/repo/subscriber.py
davinotdavid: Sure, moved it to `defines.py`! Thanks!

backend/test/unit/test_subscriber_repo.py
davinotdavid: Wow took me a shameful amount of time to figure out why tests were failing when I moved the constant to `defines.py` :( monkeypatch to the rescue!

Added a another test!

Issue #1211, opened 2025-08-28
CalDAV calendars require page refresh to be listed and events not being retrieved


davinotdavid:
**Describe the bug**

There are two parts of this bug:

### Events not being retrieved
- After the work on allowing multiple google account connections to be added, we made changes in the modelling between `calendars` and `external_connections` tables.
 
- CalDAV connections, however, are not linking their external_connection_id to their calendars upon sync, leaving the calendar with a broken link with their respective external_connection.

### CalDAV connection does not refresh the calendar list in Settings

- Self-explanatory, after a CalDAV connection, it is required to do a page refresh to load the calendars from the connection.

**To Reproduce**

1. Go to Settings > Connected Applications
2. Click on Add CalDAV and add a valid CalDAV connection
3. Notice that CalDAV calendars are not populated
4. Refresh the page, note that CalDAV calendars are populated
5. Connect a CalDAV calendar that has events in it
6. Navigate to the Dashboard and verify that the events don't show up

**Expected behavior**
CalDAV calendar events should be retrieved correctly and each calendar should be tied to their specific external_connection_id.

**Actual behavior**
CalDAV calendars are created with `NULL` in external_connection_id and events are not being retrieved.

**Additional context**

https://github.com/thunderbird/appointment/issues/394

https://github.com/thunderbird/appointment/issues/1193

https://github.com/thunderbird/appointment/pull/1078/files#diff-834b7644fd5d58716ff67b03e6e988ff3a7abfe48c6be456a25c7d6e5aacf9f9R253

Issue #1193, opened 2025-08-21
caldav list_events not working with fastmail caldav


MelissaAutumn:
For some reason on a fresh install the fastmail caldav calendar I'm using is skipping events.

This might be due to a library update, I have a fix ready though.

MelissaAutumn:
I can't actually do this fix atm because it requires upgrading both caldav and icalendar. Will tag it for next week.

kewisch:
I believe this is fixed in #1212 ?

Issue #741, opened 2024-11-15
Check before accessing event attributes on caldav


MelissaAutumn:
Sentry bug: https://thunderbird.sentry.io/issues/6031913346/?project=4505428124827648&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&statsPeriod=14d&stream_index=2

On https://github.com/thunderbird/appointment/blob/main/backend/src/appointment/controller/calendar.py#L403 we assume the calendar event has a summary, in some cases it doesn't! It seems like only start and end are required fields, so we should treat everything else as optional.

devmount:
I'll look into this. I found the [following spec](https://icalendar.org/iCalendar-RFC-5545/3-6-1-event-component.html):

```plain
eventprop  = *(
            ;
            ; The following are REQUIRED,
            ; but MUST NOT occur more than once.
            ;
            dtstamp / uid /
            ;
            ; The following is REQUIRED if the component
            ; appears in an iCalendar object that doesn't
            ; specify the ["METHOD"](https://icalendar.org/iCalendar-RFC-5545/3-7-2-method.html) property; otherwise, it
            ; is OPTIONAL; in any case, it MUST NOT occur
            ; more than once.
            ;
            dtstart /
            ;
            ; The following are OPTIONAL,
            ; but MUST NOT occur more than once.
            ;
            class / created / description / geo /
            last-mod / location / organizer / priority /
            seq / status / summary / transp /
            url / recurid /
            ;
            ; The following is OPTIONAL,
            ; but SHOULD NOT occur more than once.
            ;
            rrule /
            ;
            ; Either 'dtend' or 'duration' MAY appear in
            ; a 'eventprop', but 'dtend' and 'duration'
            ; MUST NOT occur in the same 'eventprop'.
            ;
            dtend / duration /
            ;
            ; The following are OPTIONAL,
            ; and MAY occur more than once.
            ;
            attach / attendee / categories / comment /
            contact / exdate / rstatus / related /
            resources / rdate / x-prop / iana-prop
            ;
            )
```

So we might indeed handle everything as optional except for `dtstart`.

MelissaAutumn:
I think we can ignore events that don't have a dtend/duration. Unless there's some existing behaviour that's expected for end-users.

PR #1212, opened 2025-08-28
Fix CalDAV connect / disconnect, list calendar refresh and external_connection_id

davinotdavid:
## Description of the Change

There were several small bugs connected to CalDAV which this PR aims to fix:

[**Problem**]
In Settings, after successfully connecting a CalDAV server, it didn't refresh the calendar list automatically

[**Fix**]
Upon successful CalDAV connection, manually refresh the calendar list (nothing was being done other than closing the connection modal)

--

[**Problem**]
In Settings, it wasn't possible to disconnect a CalDAV account. Clicking "disconnect" would do nothing.

[**Fix**]
We were confusing enums in the comparison that opens the CalDAV disconnect modal: we were comparing a calendar provider with an external connection provider. This wasn't a problem with Google because they were both coincidentally equal to 2.

--

[**Problem**]
CalDAV events were not showing up properly in the Dashboard

[**Fix**]
Essentially revert [this commit](https://github.com/thunderbird/appointment/commit/8a1b5e3fcdecee8cd0ebf12c5332372912f14dae) since apparently `vevent.components()` return nothing and we are reading the start date / end date directly from the vevent object anyways.

--

[**Problem**]
CalDAV calendars were not being associated with their `external_connection_id`

[**Fix**]
Upon calendar sync after CalDAV server auth, make the connection between the Calendar <-> External Connection

## Benefits

- Ability to connect / disconnect CalDAV accounts in Settings
- User don't need to refresh to see CalDAV calendars upon connection
- CalDAV calendar events show up properly

## Applicable Issues

https://github.com/thunderbird/appointment/issues/1211
https://github.com/thunderbird/appointment/issues/1193

### DIFF
--- a/backend/test/integration/test_appointment.py
+++ b/backend/test/integration/test_appointment.py
@@ -1,6 +1,7 @@
 import pytest
 import dateutil.parser
 from unittest.mock import patch, MagicMock
+from datetime import datetime
 
 from defines import DAY1, DAY3, auth_headers, TEST_USER_ID
 from appointment.database.repo import appointment as appointment_repo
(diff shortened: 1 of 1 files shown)


MelissaAutumn:
Here's the context for the vevent stuff 
https://github.com/thunderbird/appointment/issues/741

I think accessing summary as a prop caused it to raise a KeyError or something like that, additionally we need a way to guard against every attribute as we found out everything but dtstart was optional lol.

frontend/src/views/SettingsView/components/ConnectedApplications.vue
```diff
@@ -77,17 +79,22 @@ async function disconnectAccount(provider: ExternalConnectionProviders, typeId:
 function displayModal(
   provider: ExternalConnectionProviders | CalendarProviders,
```
MelissaAutumn, line 80: Oh yeah, this is not ideal haha. The fix works for now, thanks!

MelissaAutumn (commented):
Just need to look into making sure events don't crash the app like they use to.

davinotdavid:
@MelissaAutumn ah thanks for the context! I think we are reasonably guarding against `KeyError` while erroring on the side of ignoring the vevents that don't have the information that we require to display the events in the calendar so I think we are good to go? Happy to change it here if you think that we need extra guarding / reporting of course!

backend/src/appointment/controller/calendar.py
```diff
@@ -495,18 +499,14 @@ def list_events(self, start, end):
...
-                'dtend' not in event_components and 'duration' not in event_components
-            ):
+            if not vevent.dtstart or (not vevent.dtend and not vevent.duration):
                 continue
 
             # Mark tentative events
             tentative = status == 'tentative'
 
-            title = vevent.summary.value if 'summary' in event_components else l10n('event-summary-default')
+            title = vevent.summary.value if vevent.summary else l10n('event-summary-default')
```
MelissaAutumn: If summary is not included in the vevent this will error out, even if we check for none iirc. The only value that it won't potentially error out on is dtstart. 

I'd recommend checking the docs for a way to determine if the object key exists and applying that logic to every vevent.<attribute>.

MelissaAutumn: Thanks! Can you give a shot with an event thats as empty as possible. (radicale might be a good way, or adding a test for that.) 

This caused a lot of headaches, so I want to make sure it won't die on us in the future 😅

davinotdavid: Ah that makes a lot of sense! Thanks for double checking here, too much JS brain here haha

Added a bunch of `hasattr` all around and some test cases to cover that. Manual testing with radicale seems to be good too on my end!

PR #1214, opened 2025-09-02
[WIP] Booker page UI overhaul

davinotdavid:
## Description of the Change

- Renamed `BookingView` to `BookerView` so that it is not confused with `BookingsView` (plural, related with the Bookings page in the main navigation)
- Removed the footer that showed up when a slot was selected in favor of enabling / disabling a `Confirm` button in the aside
- Added an `aside` that hosts the schedule / availability / meeting details

## Benefits

- Better UX according to the Design team

- Given that this was based on wireframes, colors are a little wacky at the moment. Specially, the card color that shows up when a slot is selected is hardcoded as I couldn't find a good suitable color after some attempts and this will be decided soon anyways

## Applicable Issues
https://github.com/thunderbird/appointment/issues/1097


davinotdavid:
@devmount sending a message here since Matrix is down (PR is not ready for review atm)

As part of the Qalendar replacement, I've made a [WeekCalendar component](https://github.com/thunderbird/appointment/blob/main/frontend/src/views/DashboardView/components/WeekCalendar.vue) to show the events on the Dashboard but can't use it as is in the Booker page since the it is quite coupled with the data it displays. Also, there are already hi-fi designs for the calendar itself with slightly different requirements than the wireframes (for example, there's a new horizontal line that correspond to `All day` events now).

I've heard that you've worked on creating calendar components before so I'd really appreciate some help on how to best design / architecture this component better to be more flexible to be used in both contexts. Or since I would need to probably do a rewrite of the component anyways, you can take this part over and start from a cleaner slate?

This PR adds the `aside` aspect of the Booker page but didn't replace the Qalendar just yet because of the reasons mentioned so it could be a potential standalone improvement unrelated with the calendar replacement itself.

devmount:
@davinotdavid Thanks for the background info! Since both are more or less the same week views, I would go with the same WeekCalendar component. But you're right, this might make things a bit more complicated.

If you want, you can finish this PR without the week view on the booker page and I'll take care of that after. My first idea would be, to make it a bit more general by adding a prop for events to render and two modes for read-only and selectable events. But I might have to take a closer look first how things are currently implemented.

MelissaAutumn:
For colour issues, try raising the background colour by 20-40%, or by lowering the blue until it passes all contrast checks https://webaim.org/resources/contrastchecker/ (you can also do this in firefox dev tools.)

I'll review this hopefully this week!

davinotdavid:
> For colour issues, try raising the background colour by 20-40%, or by lowering the blue until it passes all contrast checks https://webaim.org/resources/contrastchecker/ (you can also do this in firefox dev tools.)
> 
> I'll review this hopefully this week!

@MelissaAutumn ah thank you for the link (and thanks in advance for the future review heh)! I meant that I tried the colours from the CSS variables that are currently defined and the pairings didn't look great so I left it as the same colour as the wireframes for now (which doesn't pass contract checks at all). The blue that would pass the contract checks for the text color would be `#0439D7` but that also looks wacky 😅 so I left it for the hi-fi design time to sort it out.

image


image


image


image

frontend/src/stores/booking-view-store.ts
```diff
@@ -18,7 +18,9 @@ export const useBookingViewStore = defineStore('bookingView', () => {
   // Data
   const selectedEvent = ref<Appointment & Slot>(null); // The selected slot also needs some data from its parent
   const appointment = ref<Appointment>(null);
-  const attendee = ref<Attendee>(null);
+  const attendee = ref<Attendee>(null); // Attendee can either be a guest or an actual user
+  const guestUserInfo = ref<{ name?: string, email: string }>({ name: '', email: '' }); // Used in the SlotSelectionUserInfo form
```
devmount: Suggestion: Can we make this a `Guest` type?

frontend/src/views/BookerView/components/SlotSelectionUserInfo.vue
```diff
@@ -0,0 +1,155 @@
...
+  userLoginError.value = '';
+  userLoginLoading.value = true;
+
+  if (userLoginFormRef.value.checkValidity()) {
+    const { error } = await userStore.login(userLoginEmail.value, userLoginPassword.value);
+
+    if (error) {
+      // FIXME: The return of the login function is not standardized at the moment
+      // and currently inferred as string | boolean but it does return an object
+      // with a detail key in this case
```
devmount, line 58: Good catch, I also had troubles with this in the past. Maybe we can create a ticket for improving the error types / error handling.

frontend/src/views/BookerView/index.vue
devmount, line 1: Bye bye tailwind 👋🏻 Thanks for this

frontend/src/router.ts
```diff
@@ -75,8 +75,8 @@ const routes: RouteRecordRaw[] = [
   },
   {
     path: '/user/:username/:signatureOrSlug?',
-    name: 'booking-view',
-    component: BookingView,
+    name: 'booker',
+    component: BookerView,
```
devmount, line 79: Thanks, this makes communication much easier 👏🏻

devmount (approved):
Thanks Davi! Great work. Some minor comments below and I have two questions:

1. Maybe it's a good idea to make the sidebar sticky? When having a long list of time slots to the left, the selected slot on the right disappears on the top:
    Image
2. For some reason, one of my timeslots is shown but not available. I don't think this is caused by this PR, but I wanted to report this anyway. Might be something on my local db state. Just if you experience something similar.
    Image

If  that PR is merged, I take care of the week view on the booker page. Thanks again!

frontend/src/views/BookerView/components/SlotSelectionUserInfo.vue
davinotdavid: Yes! That would be wonderful to review / refactor all functions but perhaps the `login` specific one is a good one to start a good pattern. Created the ticket here: https://github.com/thunderbird/appointment/issues/1223

frontend/src/stores/booking-view-store.ts
davinotdavid: For sure! Added in this commit https://github.com/thunderbird/appointment/pull/1214/commits/e179a59e6088494f80d2c0aa9e5fdba0e9bb0d38 thanks!

davinotdavid:
@devmount Thanks a lot for the review / suggestions!

About the first point, I've added the sticky behaviour to the aside and it does look much better, thanks for the suggestion!


https://github.com/user-attachments/assets/f50075e4-e413-4cf7-82ec-e5d06a07dd69


About the second point, I didn't experience this on my end but that's interesting... I wonder if the calendar is showing a slot that is not supposed to be there or if the selection of a valid slot is failing with the new code. If this is happening with this particular slot, perhaps the slot shouldn't show up in the calendar in the first place? 🤔

devmount:
Great stuff, Davi! The sticky sidebar looks awesome now! Also thanks for always providing those beautiful screencasts 👌🏻 

I will investigate a bit more about that time slot issue. But that's really no blocker here, if I find the cause, I'll create another issue. This can be merged now imo!

PR #1220, opened 2025-09-04
E2E settings test updates

rwood-moz:
More additions and improvements to settings tests:

- Update account settings test to check for display name and download data button
- Add a test to change the display name, verify changed on book appointment (share link) page
- Add some `expect.soft`'s to some of the settings tests to ensure the test continues to run (and set the settings back to their original state) even if the test assert fails (test is still marked as a failure in that case)
- In the `able to change start of week` test temporarily remove the start of week check on the dashboard, as this is currently failing because the dashboard calendar UI is being overhauled; will add this back after that is complete

### DIFF
--- a/test/e2e/pages/settings-page.ts
+++ b/test/e2e/pages/settings-page.ts
@@ -24,11 +24,13 @@ export class SettingsPage {
   readonly startOfWeekMondayBtn: Locator;
   readonly startOfWeekSundayBtn: Locator;
   readonly accountSettingsHeader: Locator;
+  readonly displayNameInput: Locator;
   readonly bookingPageURLInput: Locator;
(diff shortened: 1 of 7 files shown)


rwood-moz:
BrowserStack [link is here](https://automate.browserstack.com/dashboard/v2/builds/c13b7ef05649440c66443c7884435e5d04f13d0b) showing the new tests green on production.

devmount (approved):
Looks reasonable, thanks Rob for your great work!
Also wise to comment some dashboard tests, since it's still wip 👍🏻

rwood-moz:
> Looks reasonable, thanks Rob for your great work! Also wise to comment some dashboard tests, since it's still wip 👍🏻

Thank you @devmount :)

PR #1225, opened 2025-09-05
Expand E2E nigthly test BrowserStack job platforms

rwood-moz:
Previously the nightly test suite runs in BrowserStack on OSX Firefox and OSX Chromium only. This PR updates the nightly test job to now run on the following platforms/browsers:

- Firefox on OSX Tahoe (upgrade from Sequoia)
- Safari (Webkit) on OSX Tahoe
- Chromium on Win 11
- Edge on Win 11

Here is [a BrowserStack link](https://automate.browserstack.com/dashboard/v2/builds/2c59f81ea13c59a6a36f1750604bdd09ce601d4a) showing the updated nightly test suite running on all of the above platforms, and all tests passing successfully. \o/

Notes:

- These expanded platforms are only for the nightly test job;
- The post-production deployment E2E test suite will continue to only run on Firefox and Chromium (for time sake)
- This PR also updates the E2E suite that runs on stage after stage deployments to only run on Firefox (and not also on Chromium) that will reduce the test time by half


davinotdavid (approved):
Nice! Thanks for working on this 🙏 

I believe macOS `Tahoe` is still in beta, right? It seems to be due to a release in the Fall, probably next week-ish with the new Apple event but just wondering if we should test it against Sequoia instead for now and do a fast follow up once its out?

rwood-moz:
> Nice! Thanks for working on this 🙏
> 
> I believe macOS `Tahoe` is still in beta, right? It seems to be due to a release in the Fall, probably next week-ish with the new Apple event but just wondering if we should test it against Sequoia instead for now and do a fast follow up once its out?

Ah good point, ok I'll revert back to Sequoia and update to Tahoe later this fall. Thanks for the review!

Issue #1224, opened 2025-09-05
Update desktop NavBar & Dashboard sidebar with hi-fi designs


davinotdavid:
**Is your feature request related to a problem? Please describe.**
Given that we received the light mode for the Dashboard sidebar and the desktop NavBar, update the code to reflect the design changes.

**Describe the solution you'd like**
Follow the Zeplin hi-fi mockup
<img width="1280" height="1066" alt="Image" src="https://github.com/user-attachments/assets/b6d17810-1284-40e1-8dd9-a881fc79c5bc" />

davinotdavid:
Closing as PR has been merged and code is in staging now.

PR #1226, opened 2025-09-05
Updates desktop NavBar & QuickActionsSideBar styling + page height fixes

davinotdavid:
## Description of the Change

- Added @phosphor-icons/vue package as we are prioritizing icons from here instead of @tabler-icons when moving into hi-fi designs.
- Updated the `NavBar` on Desktop with "forced dark mode" as we don't have a light mode yet.
- Fixed Footer in the Bookings page. This was because when the body content is shorter than the window, it was not growing to push the Footer to the bottom so added some baseline styles to fix that in `main.css`.

Dark mode
image

Light mode
image

## Benefits

- NavBar more aligned with the hi-fi designs
- Page height fix

- Achieving the same colour / shadow accuracy from Zeplin to CSS is a challenge 😓 the gradients don't produce the same effect as apparently there are some hidden layers at work so unsure just now. Perhaps there's another layer of detail that can be done afterwards but IMHO this is ok for now.
- The background color of the whole app is not `white` anymore, it is `#f7f7f7` for light mode and the surfaces are `white` which requires us to flip the color logic pretty much everywhere in the app.

## Applicable Issues

https://github.com/thunderbird/appointment/issues/1224

### DIFF
--- a/frontend/src/components/NavBar.vue
+++ b/frontend/src/components/NavBar.vue
@@ -145,7 +145,7 @@ const copyLink = async () => {
     padding-inline: 1rem;
     box-shadow: 0 10px 15px -3px rgb(0 0 0 / 0.1), 0 4px 6px -4px rgb(0 0 0 / 0.1);
     overflow: visible;
-    background-color: #1A202C; /* TODO: --colour-neutral-base but we don't have light mode for NavBar yet */
+    background-image: linear-gradient(to top, #1a202c, #24a39d 300%); /* TODO: update to point to a var when we have light mode defined */
     z-index: 50;
   }
 


frontend/src/components/NavBar.vue
```diff
@@ -140,14 +138,14 @@ const copyLink = async () => {
...
     justify-content: space-between;
     width: 100%;
-    height: 64px;
+    height: 68px;
     padding-inline: 1rem;
     box-shadow: 0 10px 15px -3px rgb(0 0 0 / 0.1), 0 4px 6px -4px rgb(0 0 0 / 0.1);
     overflow: visible;
-    border-block-end: 1px solid var(--colour-neutral-border);
-    background-color: var(--colour-neutral-base);
+    background-color: #1A202C; /* TODO: --colour-neutral-base but we don't have light mode for NavBar yet */
```
MelissaAutumn: It's not perfect, but try:
`background: linear-gradient(0deg, var(--colour-ti-base) -40%, #566a92 200%);`

MelissaAutumn: Thats my temp value for the accounts hi-fi that I will someday get back to.

frontend/package.json
```diff
@@ -12,6 +12,7 @@
     "test": "vitest"
   },
   "dependencies": {
+    "@phosphor-icons/vue": "^2.2.1",
```
devmount, line 15: Just wondering: Do we have any decision on icon libraries that we use? When starting to implement Appointment, I added the tabler icons just for convenience to fill in design gaps. So we can totally replace them with design SVGs or icons from another lib.

frontend/src/elements/NavBarItem.vue
```diff
@@ -10,18 +10,31 @@ defineProps<Props>();
...
+    color: #EEEEF0; /* TODO: --colour-ti-base but we don't have light mode for nav bar yet */
+
+    &.active {
+      color: #D9D9DE; /* TODO: --colour-ti-secondary but we don't have light mode for nav bar yet */
+    }
+
+    a {
+      text-align: center;
+      min-width: 120px;
+    }
```
devmount: Can we give the actual clickable element a bit more height? Imho, I'm used to not being forced to actually click on the text on menu items 😇  (but might just be me)

Image

frontend/src/components/NavBar.vue
davinotdavid: Before:
image

After:
image

Zeplin mockups:
image

@MelissaAutumn Thanks! That was a good try I think! Not close to the Zeplin mockups though, do you prefer the non-gradient or the gradient one for now?

frontend/package.json
davinotdavid: Ah apparently the decision (at least for now) is to use `phosphor-icons`. The design team recommended to use it as it is the same one they use for their designs.

Unfortunately we will have two icon libraries in the app until we make a full transition but the idea is to phase out / remove tabler icons, yes

frontend/src/elements/NavBarItem.vue
davinotdavid: Oh 100%, great catch! That's definitely not ideal to have a slim nav item, changed it to be the full height! Thanks for pointing it out!

image

frontend/package.json
devmount: Perfect, thanks for this information! Do we have a ticket for that? I can file one if needed.

frontend/src/elements/NavBarItem.vue
devmount: Awesome, thanks! 😍

frontend/src/components/NavBar.vue
MelissaAutumn: Ooh i see. They might be different colours than the accounts mockup. Ill double check in a bit.

frontend/package.json
davinotdavid: Just created one for tracking https://github.com/thunderbird/appointment/issues/1232 but feel free to add / edit any information there! Thanks!

frontend/src/components/NavBar.vue
MelissaAutumn: Ah they are different colours. Replace `#566a92` colour with `#24a39d`. Should be good until design can fix it for us :smile:

MelissaAutumn: Oh wait the gradient does seem to work from zepplin. 
`background-image: linear-gradient(to top, #1a202c, #24a39d 300%);`. I would just roll with that.

davinotdavid: Zeplin
image

Code
image

Ah it is close but it doesn't seem accurate either :/ close enough for now though. Thanks! Updated with the above.

MelissaAutumn: I think theres some additional effects it does with background blurring. Works well enough for now.

frontend/package.json
devmount: Great, thank you!

PR #1227, opened 2025-09-05
Delete old automation user

ryanjjung:
## Description of the Change

We have switched over to the new `StackAccessPolicies` model of permissions, and we no longer need this old CI config.

## Benefits

Unnecessary path to environmental access revoked.


devmount (approved):
Every PR with more red than green is a good PR! Thanks for this!!

Issue #1229, opened 2025-09-08
Protocol and host missing from copied booking share link when run on local stack


rwood-moz:
**Describe the bug**
When running appointment on the latest main on my local stack, the booking share link copied is missing the protocol and host (and has `None` instead).

**To Reproduce**
1. Start local stack on latest main
2. Go to the dashboard, click the 'copy link' button on the top right
3. Past the link
4. -OR-
5. Go to the `Availability` page
6. Scroll down to the `Booking page link` section
7. Notice the link displayed is missing the protocol and host

**Expected behavior**
Expect share link to be correct.

**Actual behavior**
Share link is missing the protocol and host i.e. looks like this:

`None/user/<myuser>/`

davinotdavid:
This is a bug that was created by this recent change that I've made in how we are evaluating short url / long url signature: https://github.com/thunderbird/appointment/pull/1207/files#diff-e4f9db8436e168a46e5c5cf753a232f188ae14dcd31fcf6c61b4230a0cfbc331R39

When `defines.py` gets imported into `main.py` the `.env` vars are not loaded yet so it evaluates to `None`.

I will provide a fix shortly to lazy load the constant.

davinotdavid:
@rwood-moz just a heads up, PR has been merged! Would you mind checking if the issue persists / close if solved?

rwood-moz:
Yep looks good, thanks for this fix!

PR #1231, opened 2025-09-08
Lazy load the LONG_BASE_SIGN_URL from defines.py

davinotdavid:
## Description of the Change

This bug came from https://github.com/thunderbird/appointment/pull/1207

`defines.py` gets imported in `main.py` before the `load_dotenv` is ran which means that the unless you specify the `FRONTEND_URL` var through your shell / as an arg, it won't read from the `.env.*` files.

This change now "lazy loads" the value so that it correctly respects the `.env.*` file settings as it was before.

Tests were passing before because I was monkeypatching the env var but that's no longer needed since it is now correctly lazy loading the long base sign url value from the `.env.test`.

> [!IMPORTANT]
> If you pull this branch for testing while your previously existing containers are up, it might not work so I'd recommend starting fresh to avoid false positives / false negatives.

## Benefits

Local environments will now correctly read the former `LONG_BASE_SIGN_URL` when `SHORT_BASE_URL` is not set up.

## Applicable Issues

https://github.com/thunderbird/appointment/issues/1229

### DIFF
--- a/backend/test/unit/test_subscriber_repo.py
+++ b/backend/test/unit/test_subscriber_repo.py
@@ -6,11 +6,11 @@ from appointment.controller.auth import signed_url_by_subscriber
 
 
 @pytest.fixture
-def clear_url_cache(monkeypatch):
+def clear_url_cache():
     """Clear the get_long_base_sign_url cache and set test FRONTEND_URL"""
     from appointment.defines import get_long_base_sign_url
     get_long_base_sign_url.cache_clear()
-    monkeypatch.setenv('FRONTEND_URL', 'https://example-long.org')
+    os.environ['FRONTEND_URL'] = 'https://example-long.org'
 
 
 class TestSubscriber:


backend/src/appointment/defines.py
```diff
@@ -35,9 +35,10 @@
 # Resolves to absolute appointment package path
 BASE_PATH = f'{sys.modules["appointment"].__path__[0]}'
 
-# Base URL used to sign / verify subscriber's signature in requests
-LONG_BASE_SIGN_URL = f'{os.getenv("FRONTEND_URL")}/user'
-
+# This has to be lazy loaded because the env vars are not available at import time in main.py
+def get_long_base_sign_url():
```
MelissaAutumn, line 41: Good candidate for `@cache`.

davinotdavid: Ah TIL about `@cache`, thanks! Added here and updated tests since now the value is cached and we can't redefine the `FRONTEND_URL` on the go (doesn't take away the merit of the tests though since it is controller by the `.env.test` anyways)

MelissaAutumn: You can probably mock the defines.get_long_base_sign_url function instead now if you'd like. Or punt it to another ticket.

backend/test/unit/test_subscriber_repo.py
```diff
@@ -5,41 +5,49 @@
...
+        monkeypatch.setenv('FRONTEND_URL', 'https://example-long.org')
 
         subscriber = make_basic_subscriber()
 
         with with_db() as db:
             signed_url = signed_url_by_subscriber(subscriber)
 
             # When SHORT_BASE_URL is not set, the signed_url should contain the long URL format
-            assert os.environ['FRONTEND_URL'] in signed_url
+            assert 'https://example-long.org' in signed_url
```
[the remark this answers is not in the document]
davinotdavid: Wow that's new, I applied the suggestion here to see how it works!

backend/src/appointment/defines.py
davinotdavid: Oh [how about this](https://github.com/thunderbird/appointment/pull/1231/commits/3ebd57884fa3f8c504cd8ca1431a8b46ecec379b)? I created a fixture to clear the cache for that function and mocking the env var instead!

Also thanks for the patience, always learning here 🙏

backend/test/unit/test_subscriber_repo.py
```diff
@@ -1,21 +1,25 @@
...
 from appointment.database.repo.subscriber import verify_link
 from appointment.controller.auth import signed_url_by_subscriber
 
 
+@pytest.fixture
+def clear_url_cache(monkeypatch):
+    """Clear the get_long_base_sign_url cache and set test FRONTEND_URL"""
+    from appointment.defines import get_long_base_sign_url
+    get_long_base_sign_url.cache_clear()
+    monkeypatch.setenv('FRONTEND_URL', 'https://example-long.org')
```
MelissaAutumn: Do we need this monkeypatch if `@cache` gives us a cache clear function?

davinotdavid: No we don't, good catch! Removed!

PR #1234, opened 2025-09-08
Update E2E tests for latest bookee page UI changes

rwood-moz:
Update the E2E tests to coincide with the latest bookee (booking request) page UI updates.


rwood-moz:
Note: This makes all of the E2E tests pass against the latest main when run on the local stack, but the nightly E2E tests job will fail until after the next production deployment (when the latest bookee page UI changes are deployed).

Tested against latest main local stack on Firefox, Chromium, and Safari. Also tested against stage [on BrowserStack](https://automate.browserstack.com/dashboard/v2/builds/f52de14d0ecf13b44ccae38de2cc4c74dc2ad2ec).

PR #1235, opened 2025-09-08
Dashboard Hi-FI Updates + App Colour Refactor

davinotdavid:
## Description of the Change

Originally, this PR was supposed to do style updates to the `WeekPicker` and `UserCalendarSync` only, according to the hi-fis but since then was later expanded to:

- Added a new color variable called `--colour-ti-black` because the Zeplin mockups refered to a `--tf-android-scrim` for the "What do you want to do today?" except that it is black (#000000) and applied to a text instead of a scrim. The dark version of it is white (#FFFFFF).

- Change the `body` background color to match hi-fi designs. This change alone required me to change the following (otherwise it light mode / dark mode would look broken in other pages):
  - Availability page
    - Some layout updates to adhere to the new 1 column layout proposed on the hi-fi
    - Updated the section header colours to be the `-ti-highlight`
    - Flipped the sections to be `--colour-neutral-base` on the background and `--colour-ti-secondary` as text
  - Bookings page
    - This looks quite bad IMHO and there's no responsive design at all but at least also flipped the colors to at least be readable in both light / dark
  - Settings page
    - Updated the section header colours to be the `-ti-highlight`
    - Flipped the sections to be `--colour-neutral-base` on the background and `--colour-ti-secondary` as text

- Updated the Dashboard `WeekCalendar` component to be a closer match with the hi-fi designs. I am avoiding a bit to change it too much (e.g. all day events are not there, calendar events are not stylized with the border) since unfortunately there's gotta be a bigger refactor to take the Booker Page usage into account. Looker nicer for now IMHO while that's not done.

Before
image

Zeplin
image

Code
image

## Benefits

- Better alignment with design team's hi-fis

- ~~We need to change the background of the `body` of the application and _all_ of the surfaces. This has to be done in one go otherwise all the surfaces won't be surfaces anymore colour-wise (heh). Also, the variables that are currently used in the hi-fis have only been considered for light mode because the dark mode using them looks... quite odd. Therefore, this has not been tackled just yet.~~

## Applicable Issues

https://github.com/thunderbird/appointment/issues/1093

### DIFF
--- a/frontend/src/stores/availability-store.ts
+++ b/frontend/src/stores/availability-store.ts
@@ -53,7 +53,7 @@ export const useAvailabilityStore = defineStore('availability', () => {
     }
 
     if (firstSchedule.value) {
-      initialState.value = firstSchedule.value;
+      initialState.value = deepClone({ ...firstSchedule.value });
 
       // Calculate utc back to user timezone
       initialState.value.start_time = scheduleStore.timeToFrontendTime(initialState.value.start_time, initialState.value.time_updated);
--- a/frontend/src/views/DashboardView/components/WeekCalendar.vue
+++ b/frontend/src/views/DashboardView/components/WeekCalendar.vue
@@ -138,16 +138,20 @@ const timeSlotsForGrid = computed(() => {
 
   // For FTUE for example, we don't have a firstSchedule yet,
   // so we can initialize the time with a default 9-5 in 24 hrs format, 30 min duration
-  const { start_time, end_time, slot_duration } = firstSchedule.value || {
+  const { start_time, end_time, slot_duration, time_updated } = firstSchedule.value || {
     start_time: '9:00',
     end_time: '17:00',
-    slot_duration: 30
+    slot_duration: 30,
+    time_updated: '1970-01-01T00:00:00',
   };
 
   if (!start_time || !end_time || !slot_duration) return [];
 
-  const startTime = dj(start_time, 'H:mm');
-  let endTime = dj(end_time, 'H:mm');
+  const timezoneAwareStartTime = scheduleStore.timeToFrontendTime(start_time, time_updated)
+  const timezoneAwareEndTime = scheduleStore.timeToFrontendTime(end_time, time_updated)
+
+  const startTime = dj(timezoneAwareStartTime, 'H:mm');
+  let endTime = dj(timezoneAwareEndTime, 'H:mm');
 
   // If the end time is before the start time (slot spans midnight), add a day to the end time
   if (endTime.isBefore(startTime) || endTime.isSame(startTime)) {


MelissaAutumn:
Feel free to do the background colour switch for light mode!

Also is the weighting correct? It might just be the bg colour difference, but the zeros look thicker. It's possible the font weights / font style we have defined in appointment doesn't line up with metroplis in zeplin 1:1.

rwood-moz:
Tried it out locally, looks good!

davinotdavid:
@MelissaAutumn alright, so I will move the PR in draft for now while I "flip" all surfaces app-wide so that I don't break dark mode with this PR.

Speaking of the weights, it looks thicker yes :( Zeplin says `font-weight: 500` and the screenshots are also using `font-weight: 500`. Using `400` looks closer in my opinion but not 100% match:

image

It looks closer with 400 though, so I will change it to that, thanks!

davinotdavid:
@MelissaAutumn @devmount @rwood-moz 

Alright, I did lots of color changes (and some layout changes as well) throughout the app so that light is like the hi-fis and dark looks less broken while we don't have everything defined for dark mode as well.

Please pull while testing and look around at all pages / interactions both in light mode and dark mode please (with a focus on the light mode since that's the one we have defined for now)! 🙏

rwood-moz:
I find it kind of difficult to read the text in new appointments created when in dark mode.

Screenshot 2025-09-09 at 10 05 30 AM

rwood-moz:
Just a question, on the `Availability` page the sections are now lined up one below the other instead of beside eachother like they are now in production. Personally I think it looks better as it is now in production because I find it easier to find what I'm looking for without having to scroll down to see the other sections. Totally just my opinion though (and I understand if you're just following the UI design spec).

Update: Sorry I just saw your comment above `Some layout updates to adhere to the new 1 column layout proposed on the hi-fi`.

rwood-moz:
When you're in dark mode and you sign out, on the sign in page if you scroll down you cannot read the text beside the screenshots:


Screenshot 2025-09-09 at 10 40 44 AM

Same screen on current dark mode in prod:


Screenshot 2025-09-09 at 10 43 40 AM


I also found another issue, if you change to dark mode then sign out, then refresh the sign in screen, it changes to light mode haha then when you sign i n the settings say it is set to dark mode but it is showing light mode. I notice this is also happening on production now so I will create a separate issue for that.

davinotdavid:
@rwood-moz thank you so much for reviewing and catching those problems! I really appreciate it!

> I find it kind of difficult to read the text in new appointments created when in dark mode.

This has been updated to look like the screenshot below. We don't have an official dark mode colour to set here yet as we don't have hi-fis for that but for now I've put it as white.

image


> On the Availability page the sections are now lined up one below the other instead of beside each other like they are now in production [...]

Ah yes, this is just following the hi-fis. Hopefully with the upcoming updated form fields / typography for the Availability page (not included in this PR) will make it better but thanks for raising the concern! Perhaps we can [add comments directly in Zeplin](https://app.zeplin.io/project/66dd0f2b2fe8f1b63cda36f1/screen/68b0a2ac22e4dfa7a1dce010) so they can fix it if needed?

> When you're in dark mode and you sign out, on the sign in page if you scroll down you cannot read the text beside the screenshots

Ah 🤦 I completely forgot to do the non-signed in home page. My bad! I have since updated the colours and converted the page from TailwindCSS to vanilla CSS. You might notice that the background SVGs don't look as nice but given that we don't have hi-fis nor wireframes for this, perhaps we can leave as is at least for this PR / yeet it in another ticket when time comes?

> I also found another issue, if you change to dark mode then sign out, then refresh the sign in screen, it changes to light mode

Thanks for creating an issue about this! Hopefully we can get to it soon 🙇

rwood-moz:
> @rwood-moz thank you so much for reviewing and catching those problems! I really appreciate it!
Yw!

> This has been updated to look like the screenshot below. We don't have an official dark mode colour to set here yet as we don't have hi-fis for that but for now I've put it as white.

I pulled the latest, yes it looks much better now, thank you!

> > On the Availability page the sections are now lined up one below the other instead of beside each other like they are now in production [...]
> 
> Ah yes, this is just following the hi-fis. Hopefully with the upcoming updated form fields / typography for the Availability page (not included in this PR) will make it better but thanks for raising the concern! Perhaps we can [add comments directly in Zeplin](https://app.zeplin.io/project/66dd0f2b2fe8f1b63cda36f1/screen/68b0a2ac22e4dfa7a1dce010) so they can fix it if needed?
>

I just realized that the new availability view though makes good sense for mobile browsers... 

> Ah 🤦 I completely forgot to do the non-signed in home page. My bad! I have since updated the colours and converted the page from TailwindCSS to vanilla CSS. You might notice that the background SVGs don't look as nice but given that we don't have hi-fis nor wireframes for this, perhaps we can leave as is at least for this PR / yeet it in another ticket when time comes?
> 

Looks much better now (can read the text when signed out in dark mode) thank you!

devmount (commented):
First of all: Thanks for all the work on this! It's nice to see Appointment getting more and more aligned with the great designs! 🎉 

When browsing through appointment and trying things out, I found the following issue:

For some reason, I got multiple event divs overlapping each other:

Image

I purged the slots and appointment tables locally before testing. When I create a booking, it would render a pending appointment as well as a remote event. It doesn't matter if I'm a logged in user or a guest. It seems however to only occur, when I checked the "Automatically confirm bookings if time is available" setting (which is funny, since this setting should normally avoid setting bookings as pending?).

Can anybody reproduce any of this?

davinotdavid:
@devmount thank you so much for reviewing / trying it out!

Oh I didn't catch that bug before but I can confirm that I can reproduce it! It happens on stage as well too :( so this is not a bug from this PR

image

I am not entirely sure why / how that is just yet but I created a ticket to tackle this separately: https://github.com/thunderbird/appointment/issues/1246

I think it shouldn't be creating the HOLD events at all if its auto-confirm, right? 🤔

MelissaAutumn (changes requested):
Hiii finally getting around to this, I'm getting this on my frontend loading into the dashoboard after login:
```
Uncaught (in promise) TypeError: can't access property "gridRowStart", $setup.timeSlotsForGrid[($setup.timeSlotsForGrid.length - 1)] is undefined
```

devmount:
@davinotdavid Good thing you can reproduce that! And thanks for filing a ticket! Yes, I agree: Normally there shouldn't be any hold events if confirmation was disabled / not activated.

MelissaAutumn:
image

Can we also add a max-width for the calendar? Either way looking much prettier!! Will finally dig into code review.

davinotdavid:
@MelissaAutumn

> I'm getting this on my frontend loading into the dashoboard after login

Oh weeeird, I cannot reproduce it here but I can force it to happen if `timeSlotsForGrid` is an empty array. I wonder if there's a race condition somewhere that makes the first render pass be an empty array somehow 🤔 It shouldn't be the case but I've added a `.?` for it not to break. Could you pull and try again?

> Can we also add a max-width for the calendar?


How about a max-width on the Dashboard page container itself, like so? This is with `max-width: 80rem` which matches with the `--xl` in `custom-media.pcss` (and margin 0 auto):

image

MelissaAutumn:
Will give it a try in a sec, thanks!

MelissaAutumn:
Sorry for the delay!

No errors now, but I still get an initial empty dashboard. Also if I switch to bookings and back to dashboard gives me 0 height even though I have calendar events

image

davinotdavid:
@MelissaAutumn no worries! thank you for reviewing! you actually spotted two bugs that I didn't catch before (they should also be happening on staging / production) but I believe they are now fixed.

[Calendar with 0 height]
- Problem here was there even though the schedule existed and it was valid, the `start_time` of your schedule was `16:00` whereas the `end_time` was `00:00` and I was generating the time column (which makes the calendar height) based off of a while loop `start_time < end_time`. Now that is fixed in this commit https://github.com/thunderbird/appointment/pull/1235/commits/aa746f52b044f14218a20ff4c115efbbb8f8fd46 by adding a +1 day if it crosses midnight.

[Events not showing up upon first load]
- This is a little more mysterious. If the calendar was 0 height, it wouldn't be able to place the event there upon first load so _maybe_ this will be solved by the first issue directly. I did change the usage of the calendar Pinia store though as the store instances could be potentially be conflicting.

Sorry to ask again and thanks for the patience but could you give it another try? 🙇

MelissaAutumn:
Ah I think I see the problem. When I first load in it's defaulting me to 4pm - 12am. When I go back and forth from Availability it fixes it (9-5). 

I think the default schedule times might be in UTC? Might need a tweak there to fix it up. 

The 0 height is fixed now though!

davinotdavid:
@MelissaAutumn 

> When I go back and forth from Availability it fixes it (9-5).

That was an interesting one and a good hint of what was happening. Turns out that in the `availability-store.ts` we were referencing the `firstSchedule` object of the `schedule-store` so when we were doing [this line here](https://github.com/thunderbird/appointment/blob/main/frontend/src/stores/availability-store.ts#L56), whenever the `initialState` changed, it was changing the `firstSchedule` in the `schedule-store` as well 😵‍💫

After fixing that, it was consistently wrong in UTC format (which is good), then it was just a matter of converting it using the handy `timeToFrontendTime`. Thanks a ton for spotting this, really!

MelissaAutumn:
Works now :tada:

MelissaAutumn (approved):

Issue #1230, opened 2025-09-08
In Availability page, the page description should enforce 250 max length


davinotdavid:
**Describe the bug**
In the Availability page while setting up a schedule, the page description has a max length validation of 250 characters. Currently this is not enforced / there's no feedback to the user and the error that shows up while trying to save is a generic "failed to save schedule, please try again".

**To Reproduce**

1. Go to Availability page
2. Fill in the Page Description field with more than 250 characters
3. Click on Save
4. See the error (check the network tab to see that the backend validation that fails is related with the max length of the page description).

**Expected behavior**
There are two things to fix here:

1. There should be a client side validation for the field. Potentially with a character count on the bottom right of the text field (TBD of the exact interaction here).
2. There should be a better error description on the error alert bar that shows up in case there's a backend error validation.

**Screenshots**

<img width="1458" height="484" alt="Image" src="https://github.com/user-attachments/assets/7f815c0e-61a8-43d9-87e8-e9731c5bc751" />

<img width="475" height="124" alt="Image" src="https://github.com/user-attachments/assets/8267026f-13c7-4562-a230-e223500b5759" />

davinotdavid:
Merged and available on staging!

Issue #270, opened 2024-02-15
Simplify pydantic validation errors


MelissaAutumn:
We don't really do a lot of form data validation mostly because everything is just loosey goosey text fields. But as we tighten things up a little more we should look into overriding the default pydantic/fastapi validation exception to be localized, and simplified. We can pretty easily override the exception class (https://stackoverflow.com/a/62937832). 

Here's an example of what it looks like now:
```
{
  detail: [{
      type: 'greater_than_equal',
      loc: ['body', 'slot_duration'],
      msg: 'Input should be greater than or equal to 10',
      input: 5,
      ctx: { ge: 10 },
      url: 'https://errors.pydantic.dev/2.5/v/greater_than_equal',
    }],
}
```

On the backend we should grab the approriate values (like type, context, input, field name (loc[1])) and generate a localized error message. We'll also have to add some extra meta data to include things like field units (like minutes for slot_duration for example.) We don't have to do every possible variation of error just the ones we use. Moving forward if we use a new type of field validation, we'll make sure to cover that in the PR. 

Here's some useful info on pydantic's error messages. 
https://docs.pydantic.dev/latest/errors/errors/

PR #1240, opened 2025-09-09
Add MAX_LENGTH help & check to Availability Page Description

davinotdavid:
## Description of the Change

- (Re?) Added the character count to the Page Description text area in the Availability page.
- Even though it shouldn't be possible due to the `max-length` definition on the field, in case somehow it gets sent to the backend, parse the error type and show a more meaning error message. We probably need to refactor this soon but this has already been captured in this issue: https://github.com/thunderbird/appointment/issues/270.
- Even though it should prevent adding more than 250 characters, add a standard HTML form validation to the whole form during submission to show the 'red border' in case it is invalid.

https://github.com/user-attachments/assets/2b96f091-ccc7-4763-984a-5127a63ce5f3

## Benefits

- Better guardrails for users in terms of Availability page description
- Better error handling / error message for this type of error

## Applicable Issues

https://github.com/thunderbird/appointment/issues/1230
https://github.com/thunderbird/appointment/issues/270

### DIFF
--- a/frontend/src/views/AvailabilityView/components/BookingPageDetails/index.vue
+++ b/frontend/src/views/AvailabilityView/components/BookingPageDetails/index.vue
@@ -27,6 +27,9 @@ const pageDescription = computed({
   }
 })
 
+const MAX_DESCRIPTION_LENGTH = 250;
+const pageDescriptionHelpLabel = computed(() => `${pageDescription.value?.length.toString() ?? 0}/${MAX_DESCRIPTION_LENGTH}`)
+
 const autoGenerateZoomLink = computed({
   get: () => currentState.value.meeting_link_provider === MeetingLinkProviderType.Zoom,
   set: (value) => {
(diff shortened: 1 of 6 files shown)


frontend/src/views/AvailabilityView/components/BookingPageDetails/index.vue
```diff
@@ -27,6 +27,9 @@ const pageDescription = computed({
   }
 })
 
+const MAX_DESCRIPTION_LENGTH = 250;
```
devmount, line 30: Just wondering: Do we have any way to share config between frontend and backend? This would be a good example of a value that should be always the same on backend and frontend...

davinotdavid: Thanks for the review! Good question, as far as I know we don't at the moment but I totally agree that this would be a great candidate to move it there. One way that I can think of is if the schema / fields were sent over on a request in order to build the form "dynamically" based on the backend but that would be quite of a heavy lifting

MelissaAutumn: I want to say this information should be defined in the schema for the route. `Field(max=250)` (or whatever the parameter is), and then process the backend's openapi.json file and generate the schemas to typescript types. 

I'm sure there's packages we could pull to automate this. :thinking: 

That's a different ticket though.

davinotdavid: Ah cool! That approach reminds me of [tRPC](https://trpc.io/) perhaps, then we could even hook up some [Zod](https://zod.dev/) to it for form validations 🤩

devmount: Totally a different ticket, yes! Just thought we might want to investigate that in the future. Thanks for your thoughts!

PR #1243, opened 2025-09-10
Add the first E2E test that runs on a real android device on BrowserStack

rwood-moz:
Add the first E2E test that runs on a real android device on BrowserStack. This involves:

- Moving the existing desktop E2E tests into a new folder (`/test/e2e/tests/desktop`) and creating a new folder for mobile tests (`/test/e2e/tests/mobile`)
- Adding code for mobile authentication; note that the mobile browsers don't support saving auth storage state like we do on the desktop tests, so we have to sign into Appointment before every test unfortunately
- Adding a new mobile theme setting test (`test/e2e/tests/mobile/settings-theme.spec.ts`)
- Adding new BrowserStack configs for mobile jobs
- Updating the version of playwright used (as must be compatible with both desktop and mobile)
- Updating the versions of desktop browsers used in BrowserStack (must be compatible with playwright version)
- Adding a new mobile nightly tests job (that runs on BrowserStack via GHA)
- Updating the E2E tests README accordingly

### DIFF
--- a/.github/workflows/nightly-tests-mobile.yml
+++ b/.github/workflows/nightly-tests-mobile.yml
@@ -32,11 +32,11 @@ jobs:
         with:
           node-version: 20
           cache: 'npm'
-          cache-dependency-path: 'test/e2e-mobile/package-lock.json'
+          cache-dependency-path: 'test/e2e/package-lock.json'
 
       - name: Install dependencies
         run: |
-          cd ./test/e2e-mobile
+          cd ./test/e2e
           npm install
 
       - name: BrowserStack Env Setup
@@ -51,6 +51,6 @@ jobs:
         # don't send GHA failure email if any of the E2E tests fail, can be annoying (I check the jobs each day in BrowserStack)
         continue-on-error: true
         run: |
-          cd ./test/e2e-mobile
+          cd ./test/e2e
           cp .env.prod.example .env
           npm run prod-nightly-tests-mobile-browserstack-gha


rwood-moz:
BrowserStack link and video running the theme test on a Google Pixel 10 [is here](https://automate.browserstack.com/dashboard/v2/builds/d1f01ca39585299c8bf8b077e5f64a2bb3629dde/sessions/43f5e9dc9ee08403b4b5105287707643aa4a01bb).

.github/workflows/nightly-tests-mobile.yml
```diff
@@ -0,0 +1,56 @@
...
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: 'npm'
+          cache-dependency-path: 'test/e2e-mobile/package-lock.json'
+
+      - name: Install dependencies
+        run: |
+          cd ./test/e2e-mobile
```
davinotdavid: Same here?

.github/workflows/nightly-tests-mobile.yml
```diff
@@ -0,0 +1,56 @@
...
+      APPT_MY_SHARE_LINK: ${{ secrets.E2E_APPT_PROD_MY_SHARE_LINK }}
+      APPT_BOOKEE_EMAIL: ${{ secrets.E2E_APPT_PROD_BOOKEE_EMAIL }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: 'npm'
+          cache-dependency-path: 'test/e2e-mobile/package-lock.json'
```
davinotdavid: It seems that we don't have the `e2e-mobile` folder under the `test` folder.

Should this be `test/e2e/package-lock.json`instead?

.github/workflows/nightly-tests-mobile.yml
```diff
@@ -0,0 +1,56 @@
...
+          username:  ${{ secrets.BROWSERSTACK_USERNAME }}
+          access-key: ${{ secrets.BROWSERSTACK_ACCESS_KEY }}
+          project-name: 'Thunderbird Appointment'
+          build-name: 'TB Appointment Nightly Tests (Mobile): BUILD_INFO'
+
+      - name: Run E2E Tests on Production on Browserstack (Mobile)
+        # don't send GHA failure email if any of the E2E tests fail, can be annoying (I check the jobs each day in BrowserStack)
+        continue-on-error: true
+        run: |
+          cd ./test/e2e-mobile
```
davinotdavid: Same here?

test/e2e/playwright.config.ts
```diff
@@ -50,56 +49,48 @@ export default defineConfig({
...
 
     /* Test against mobile viewports. */
-    // {
-    //   name: 'Mobile Chrome',
-    //   use: { ...devices['Pixel 5'] },
-    // },
-    // {
-    //   name: 'Mobile Safari',
-    //   use: { ...devices['iPhone 12'] },
-    // },
```
davinotdavid, line 92: Does this mean that we are only testing on Pixel 7? Not sure how much overhead it would be but perhaps it is nice to test on both iOS and Android?

rwood-moz: Yes this issue is just to add support for the first android device, I will be adding iOS (and other android devices too) in the future.

.github/workflows/nightly-tests-mobile.yml
rwood-moz: Good catch! During dev I had used a different folder, but now it's in /test/e2e, I'll update that (and the other 2 places you noted) thanks :)

test/e2e/playwright.config.ts
rwood-moz: The test is running on a real Google Pixel 10 device in BrowserStack - as specified in the [browserstack-mobile.yml](https://github.com/thunderbird/appointment/pull/1243/files#diff-06b2dc0c010f55040940bb3d1883fd7c8c544e6d126935c1c559da40ce1b8d5e) config file.

The Pixel 7 reference here (in the playright.config) is only used when run on a local machine and not BrowserStack - it uses the playwright emulated mobile browser view. It's handy while working on the tests to run them on a view that is the same size as a mobile device screen. As with the real devices, in the future I will add more emulated viewports (including for iOS).

The Pixel 7 is the latest Pixel series device that Playwright offers emulated views for. :)

davinotdavid (approved):
Awesome! Thanks for working on this 🙏

PR #1249, opened 2025-09-11
Autoscale the backend

ryanjjung:
## Description of the Change

This makes use of [this tb_pulumi change](https://github.com/thunderbird/pulumi/pull/215) to enable autoscaling of our Fargate clusters.

This PR also adds a `dev` environment, which is now destroyed, but it was useful for testing this without interfering with stage. The configs can stay.

## Benefits

Within certain boundaries, ECS should respond automatically to CPU and memory load.

### DIFF
--- a/pulumi/config.prod.yaml
+++ b/pulumi/config.prod.yaml
@@ -217,6 +217,16 @@ resources:
               - name: REDIS_USE_CLUSTER
                 value: "True"
 
+  tb:autoscale:EcsServiceAutoscaler:
+    backend:
+      cpu_threshold: 80
+      ram_threshold: 80
+      cooldown: 180
+      disable_scale_in: False
+      min_capacity: 2
+      max_capacity: 4
+      suspend: False
+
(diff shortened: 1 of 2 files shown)


pulumi/config.dev.yaml
```diff
@@ -0,0 +1,273 @@
...
+              - name: REDIS_USE_SSL
+                value: "True"
+              - name: REDIS_USE_CLUSTER
+                value: "True"
+              - name: OIDC_TOKEN_INTROSPECTION_URL
+                value: "https://auth-stage.tb.pro/realms/tbpro/protocol/openid-connect/token/introspect"
+              - name: OIDC_EXP_GRACE_PERIOD
+                value: "60"
+
+  tb:autoscale:EcsServiceAutoscaler:
```
MelissaAutumn, line 228: Should we add this to stage and prod?

ryanjjung: Yes, I'm working on it right now. Will update this PR soon. Sorry for the brain fog.

Issue #1233, opened 2025-09-08
Timezone settings between Availability <-> Booker page mismatch


davinotdavid:
**Describe the bug**
Writing this on behalf of a user that lives in `Europe/Berlin`, it seems that there is a mismatch between the slots available derived from the Availability settings vs. the ones shown in the Booker page.

This is the availability setting from the user:

<img width="917" height="895" alt="Image" src="https://github.com/user-attachments/assets/8766ba42-b6c2-4d91-b627-4f80b8c88fe2" />

This is the booker page (using the same computer, in a Private Window):

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/2995204f-df06-408b-b0a5-57abb9754328" />

The user tried to book an appointment on September 10th (Wednesday) on the first slot (9:00 AM ~ 9:30 AM) but there was an error saying that the time slot is not found / not available.

How can this be, if the slot is being displayed correctly and, per the Availability settings, it says that the Wednesday availability is from `9 AM to 6 PM` ?

The user then selected a new random slot from their availability and it sent a booking request successfully.

However, when they tried to accept the booking via the confirmation email it showed an error:

<img width="395" height="424" alt="Image" src="https://github.com/user-attachments/assets/2f75c30a-1d86-474b-9ab3-a37e17ad8569" />

Sentry error:
https://thunderbird.sentry.io/issues/6209054483/?project=4505428124827648&query=is%3Aunresolved&referrer=issue-stream

**To Reproduce**
Unclear yet on the exact steps to reproduce the error but the assumption at the time of writing this ticket is that there are problems with the Timezone settings, slot display times and Booker calendar view.

**Expected behavior**
The slots shown to the Booker should respect the Availability set (including Timezone) and any available slot in ideal conditions (as in, not counting on race conditions and such) should be bookable.

**Actual behavior**
Not all slots shown seems to be bookable and the ones that are bookable, can't be accepted. Possibly due to a timezone check error / mismatch.

davinotdavid:
@devmount I believe this is similar with what you've experienced here https://github.com/thunderbird/appointment/pull/1214#pullrequestreview-3188555895 as this report also comes from a user in Europe/Berlin timezone 🤔

devmount:
Hm, although the scenario described here works without issues locally on my end... I'll do some more testing on this.

devmount:
We did some fundamental changes in #1259, so hopefully this is fixed now. Please reopen if it still persists!

Issue #525, opened 2024-07-09
Busy Indicators aren't good for privacy


ryanleesipes:
**Is your feature request related to a problem? Please describe.**
I understand why we show busy times, so folks can have some idea of what state they are getting the person they are booking is in (did they just come out of two back-to-back meetings?)

But ultimately, I think this is bad for the privacy of the calendar owner. I don't want people to know I have four meetings on the day they are getting me, or one another day. Appointment gives random people on the tool too much insight into my schedule over a very broad timespan.

**Describe the solution you'd like**
Much like other tools in the scheduling space, we should have simplified views that show just the available times. Probably in an agenda view versus a calendar view, such as:

# Wednesday, 17

- 11:00am-11:30am
- 11:30am -12:00pm
- 1:30pm - 2:00pm

(The points above would be big buttons to select)

thunderbecca:
"Busy" indicators were not carried over into the updated designs for the Booker’s view. Time slots that fall within the active availability settings (e.g., Monday 9:00 AM to 5:00 PM) that are unavailable due to calendar events will appear with a shared solid “Unavailable” background.

There is an exception exception since most calendars offer a setting that allows individual events to be marked as either “Busy” or “Free” on the user’s primary calendar, which is raised in another issue [#995](https://github.com/thunderbird/appointment/issues/995).

davinotdavid:
Closing this issue since there were design iterations on it that have been implemented and shipped. If this is still a concern, please feel free to reopen the ticket so we can continue discussing!

PR #1259, opened 2025-09-17
Booker page week calendar and Qalendar removal

devmount:
## Description of the Change

This PR reuses the new WeekCalendar component for the Booker page too and removes Qalendar as a dependencies.

image

## Benefits

We finally get rid of a deprecated dependency.

## Known issues

- The Booker page is still fully functional, but the styles need some adjustments to look like design.
- The WeekCalendar and WeekPicker components should be moved to a more general place (its currently located under the DashboardView)

## Applicable Issues

Related to #1097


devmount (commented):
@davinotdavid I wasn't sure if we can actually remove the calendar-mini-month components too? It seems to be in use for the appointment-sliding-panel-modify component.

davinotdavid:
@devmount Ah yes, in my opinion you can remove `calendar-mini-month` too. Thanks for catching that!

I think it is still quite a long way for us to reinstate the modify appointment flow (not tracked as an Early Bird feature) and we are not sure yet if the mini month interaction will be in the hi-fis at this point (or if it is going to be replaced by something else entirely like a standard HTML date picker input for example).

It would be nice to add to the comment in the [AppointmentSlidingPanelModify.vue](https://github.com/thunderbird/appointment/blob/main/frontend/src/views/BookingsView/components/AppointmentSlidingPanelModify.vue#L1-L3) that we are removing the `CalendarMiniMonth` and perhaps make the necessary code adjustments / add TODOs just so it doesn't break on linting.

frontend/src/views/DashboardView/components/WeekCalendar.vue
```diff
@@ -438,6 +499,20 @@ const filteredPendingAppointmentsForGrid = computed(() => {
...
+  border: none;
+  margin: .125rem;
+  width: 90%;
+
+  &.dark {
+    color: var(--colour-ti-black);
+  }
+
+  &.selected {
+    background-color: #58C9FF !important;
```
davinotdavid: Not for this PR but commenting to remind myself for the next PR about matching the hi-fi design once we have a hi-fi design for it

rwood-moz:
Tried it out locally, looks good!

rwood-moz:
Oh hey I just noticed when looking at the bookee page the starting time is 4AM (shows 4AM to 11AM only) even though I have my availability set to 9AM to 5PM America/EST timezone:

Screenshot 2025-09-17 at 3 18 28 PM

Whereas on production now it shows 8AM to 6PM in week view on the bookee page when I have the same availability set.

rwood-moz:
Also note this is going to require an update to the request booking E2E test (the locators that the test uses to select an appointment slot are changed with this PR). I can take care of updating the E2E test for you after this lands, I have created #1260.

devmount:
> It would be nice to add to the comment in the [AppointmentSlidingPanelModify.vue](https://github.com/thunderbird/appointment/blob/main/frontend/src/views/BookingsView/components/AppointmentSlidingPanelModify.vue#L1-L3) that we are removing the `CalendarMiniMonth` and perhaps make the necessary code adjustments / add TODOs just so it doesn't break on linting.

Thanks, I did that. I also deleted the CalendarManagement component. It seems we replaced that with the new settings page too.

devmount:
@davinotdavid Oh that's a very good find!

1. We could, for now, just calculate the time range shown in the week view based on the given events. E.g. show the grid from min(start)-1h until max(end)+1h. Would that be sufficient for now? Oh I think that's what you meant with "window it" just to the time frame that can contain the available slots.
2. Thanks for bringing this up. I think there was a discussion going on about those blockers giving too much insight into a users schedule. But yeah, we'll have to wait for the hi-fi designs. For now I would just keep the gaps.

devmount:
@rwood-moz Thanks for catching that! There's definitely a timezone issue going on (haha would've been a disappointment if there wasn't one 😂 ), I'll look into that. Might be connected to the time window not being properly calculated (as Davi pointed out). Or even to #1233 

And you're totally right! I didn't have the time to look at the e2e tests yet - so if you could update that after this PR landet, that would be great 🙏🏻

davinotdavid:
@devmount awesome! Thanks for working on this one 🙏 

> We could, for now, just calculate the time range shown in the week view based on the given events. E.g. show the grid from min(start)-1h until max(end)+1h. Would that be sufficient for now? Oh I think that's what you meant with "window it" just to the time frame that can contain the available slots.

Yep I think that would be sufficient for now! I think this is what is happening in @rwood-moz 's end as well!

> Thanks for bringing this up. I think there was a discussion going on about those blockers giving too much insight into a users schedule. But yeah, we'll have to wait for the hi-fi designs. For now I would just keep the gaps.

Ah cool! I didn't know there were discussions about this. Happy to move on without the busy blocks then!

devmount:
> Ah cool! I didn't know there were discussions about this. Happy to move on without the busy blocks then!

I think this was the ticket: https://github.com/thunderbird/appointment/issues/525

devmount (commented):
@davinotdavid @rwood-moz Feel free to retest this again. I hope I fixed the issue with timezones. There was also a nasty bug with events over midnight. I tested this with a schedule set in Europe/Berlin and a Booker page shown in America/Los_Angeles.

rwood-moz:
> @davinotdavid @rwood-moz Feel free to retest this again. I hope I fixed the issue with timezones. There was also a nasty bug with events over midnight. I tested this with a schedule set in Europe/Berlin and a Booker page shown in America/Los_Angeles.

Yep looks great! I also changed the timezone and it also worked great. Thanks for the fix!

davinotdavid (approved):
Awesome, thanks for the great work & cleanups 🎉

Issue #1262, opened 2025-09-22
Update `services-ui` to `v1.0.2`


davinotdavid:
**Is your feature request related to a problem? Please describe.**
We should update the `services-ui` package to `v1.0.2` to benefit from the new buttons and various other components added (VisualDivider, UserAvatar, etc).

However, there is a breaking change in `services-ui` where we should replace the `SecondaryButton` usage with a `PrimaryButton` with the `outline` variant.

davinotdavid:
PR has been merged!

PR #1263, opened 2025-09-22
Update services-ui package to v1.0.2 + small fixes

davinotdavid:
## Description of the Change

- The version v1.0.2 of the `services-ui` introduced a breaking change regarding the `SecondaryButton`. So this PR replaces all its usage with the `PrimaryButton` with the `variant="outline"`.

- Tailwind uses `.outline` to add an `outline-style` CSS declaration that visually doubles the border of our new outline variant buttons. This PR overrides it to be `none` instead (doesn't seem to affect focus ring / hover states).

- Added a baseline `color` to body so that the text color can be inherited throughout all the elements as I've noticed that in FTUE there were some text that was not legible in dark mode (e.g. Google Account setup step / Zoom step).

- Fixed the non-signed in footer nav links. Since they were not prefixed with a `/` it tried to get a relative path which didn't exist.

- Removes the `SecondaryButton` from the `elements` folder so that there's no confusion on whether or not it exists anymore (it doesn't).

- Replaced `UserAvatar` from the `elements` folder with the one from `services-ui`.

- Using `StandardFooter` instead of re-implementing the footer.
 

## Benefits

- `services-ui` is updated and we can proceed further with the UI overhaul.
- NavBar uses `UserAvatar` (the one from the `elements` folder is deleted)!
- Footer links for non-signed users now work.
- Footer uses `StandardFooter`!
- FTUE in dark mode is now legible.
- No more `SecondaryButton`, not even in the elements folder!

- Even though I have replaced some instances of the `PrimaryButton` that comes from `elements` with the one that comes from `services-ui`, I didn't replace them all since I believe we some of them are better suited for the `brand` variant instead and this is TBD in the future.
- The [StandardFooter has a bug](https://github.com/thunderbird/services-ui/issues/145) on the translation aspect which forces us to have specific i18n keys for it to render properly. These can be removed once the bug is fixed!
- The `UserAvatar` component from `services-ui` in dark mode doesn't look very legible. Perhaps we should update that soon-ish, opened a ticket here: https://github.com/thunderbird/services-ui/issues/146

## Applicable Issues

https://github.com/thunderbird/appointment/issues/1262

### DIFF
--- a/frontend/src/components/FooterBar.vue
+++ b/frontend/src/components/FooterBar.vue
@@ -2,6 +2,7 @@
 import { storeToRefs } from 'pinia';
 import { useI18n } from 'vue-i18n';
 import { useUserStore } from '@/stores/user-store';
+import { StandardFooter } from '@thunderbirdops/services-ui';
 
 const { t } = useI18n();
 const userStore = useUserStore();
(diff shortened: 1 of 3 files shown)


davinotdavid:
@devmount Thank you so much for the review 🙇 I really appreciate it!

Ah indeed, sorry about the inconsistency. In the Storybook, the `iconLeft` / `iconRight` slots in the button don't increase the height of the button but since we can add any arbitrary component inside the main slot of the button, it could increase it.

I've created a ticket in `services-ui` as I believe it would be nice to constraint / enforce the height on the `services-ui` side instead of the user of the button to have to deal with it. https://github.com/thunderbird/services-ui/issues/148

EDIT: FYI, there's an upcoming change in the Settings page that will look quite different (potentially also using your new link component)

image

Issue #1257, opened 2025-09-16
Update SlidingPanel component with new hi-fi designs


davinotdavid:
**Is your feature request related to a problem? Please describe.**

As the Design Team prepared the hi-fi designs for the SlidingPanel component, we should apply those to Appointment for a cohesive look throughout the app.

Zeplin link: https://app.zeplin.io/project/66dd0f2b2fe8f1b63cda36f1/screen/68b0ba851111cea7a552f99e

Note that this might require some component updates in services-ui as well since the standard components we were using before differ from the hi-fi designs.

**Describe the solution you'd like**
Match the design as close as possible with this:

<img width="1280" height="1066" alt="Image" src="https://github.com/user-attachments/assets/fbab9abd-8ee0-493e-a4f9-d45d9310f480" />

davinotdavid:
PR has been merged!

PR #1264, opened 2025-09-23
Appointment Sliding Panel hi-fi design updates

davinotdavid:
## Description of the Change

- Updated `SlidingPanel` and the `AppointmentSlidingPanel` variants according to the hi-fi designs.
- Added a new `Download ICS` button functionality to the `AppointmentSlidingPanel`.
- Added the user themselves into the attendees list.
- Changed the `GET /apmt/serve/ics/{slug}/{slot_id}` or `public_appointment_serve_ics` route in the backend so that we can download the file through an anchor tag.

## Screenshots
Zeplin
image

Code

(Unconfirmed)
image

(Confirmed)
image

(Cancelled)
image

## Benefits

- Better alignment with high fidelity Designs.

- It would be very nice to have the badges from `services-ui` so there's a TODO for that in the code to replace the logic once its ready.
- `UserAvatar` component needs to have a `xs` variant for `24px / 24px` to be the same size as the designs. Can be easily updated / cleanup once we cut a version that has https://github.com/thunderbird/services-ui/pull/150

## Applicable Issues

https://github.com/thunderbird/appointment/issues/1257

### DIFF
--- a/frontend/src/locales/de.json
+++ b/frontend/src/locales/de.json
@@ -311,7 +311,7 @@
     "guest": "{count} Gäste | {count} Gast | {count} Gäste",
     "hardDelete": "Endgültig Löschen",
     "home": "Start",
-    "host": "Gastgeber",
+    "host": "Organisator",
     "id": "ID",
     "immediately": "Sofort",
     "inPerson": "Vor Ort",


frontend/src/locales/de.json
```diff
@@ -310,6 +311,7 @@
     "guest": "{count} Gäste | {count} Gast | {count} Gäste",
     "hardDelete": "Endgültig Löschen",
     "home": "Start",
+    "host": "Gastgeber",
```
devmount: Better wording in this context would be:


```suggestion
    "host": "Organisator",
```

davinotdavid: Thank you so much! 🙇

Issue #1255, opened 2025-09-16
Update Availability page with new hi-fi designs


davinotdavid:
**Is your feature request related to a problem? Please describe.**
As the Design Team prepared the hi-fi designs for the Availability page, we should apply those to Appointment for a cohesive look throughout the app.

Zeplin link: https://app.zeplin.io/project/66dd0f2b2fe8f1b63cda36f1/screen/68b0a2ac22e4dfa7a1dce010

Note that this might require some component updates in `services-ui` as well since the standard components we were using before differ from the hi-fi designs.

**Describe the solution you'd like**
Should match as close as possible the hi-fi below:

<img width="1280" height="2356" alt="Image" src="https://github.com/user-attachments/assets/d8970bb9-5aaa-44f4-bd74-f0e17c2af0d8" />

davinotdavid:
There's still some UI details to be changed in `services-ui` but the overall reorganization of components / functionality / layout is here. New updates to the `services-ui` will make this look better so all good closing it for now!

PR #1269, opened 2025-09-26
Update Availability page with new hi-fi designs

davinotdavid:
## Description of the Change

- Updated the Availability page to be compliant with the new designs.
- Added `Timezone` as an editable field in the Availability page (cc @rwood-moz)
- Updated the `AlertBox` to leverage the `NoticeBar` updates in `services-ui`.
- Changed the pattern for the `Saving Changes` footer that showed up. Now it is not fixed and it shows up at the very top and at the very bottom only.
- Small fix on Settings > Connected Applications import of the `BaseBadge`. 
- Removed the Tailwind overrides in `main.css` regarding the focus ring as we are dealing with it in services-ui itself.

Zeplin link for reference: https://app.zeplin.io/project/66dd0f2b2fe8f1b63cda36f1/screen/68b0a2ac22e4dfa7a1dce010

Availability

Code: 
image

## Benefits

- Better alignment with the Design Team.

- There a few divergent things / fixes that will be fixed on the `services-ui` side for a pixel perfect approach:
  - `SegmentedControl` options have the wrong font-size (should be smaller) and the vertical padding should be slightly larger
  - Label of `TextInputs` have the wrong letter-spacing (should be slightly larger)
  - `TextArea` should have the counter within the textarea itself, located on the bottom right instead of hacking the `help` prop for achieving this

- The hi-fi design is not considering `one day` as an option for the `Minimum Notice` control even though that's currently the default schedule setting. I've asked a question a few days ago but didn't get an answer yet but I think this is not blocking as I just stacked them (otherwise they look too cramped side-by-side).

- The `Start Time` and `End Time` are currently `input[type="time"]` but in the hi-fi design it is a dropdown. I believe that this is a mistake to be a dropdown as it would need to have LOTS of options within it so left as is for now.

## Applicable Issues

https://github.com/thunderbird/appointment/issues/1255


frontend/src/components/ConfirmationModal.vue
```diff
@@ -57,7 +57,9 @@ const emit = defineEmits(['close', 'confirm', 'error']);
         :label="confirmLabel"
```
devmount: I think the label prop isn't needed anymore on the button components, or is it?

frontend/src/components/ConfirmationModal.vue
```diff
@@ -57,7 +57,9 @@ const emit = defineEmits(['close', 'confirm', 'error']);
...
         @click="emit('confirm')"
         :title="t('label.confirm')"
-      />
+      >
+        {{ t('label.confirm') }}
```
devmount: Do we want to use the `confirmLabel` prop here or can we remove that too?

frontend/src/views/AvailabilityView/components/BookingPageLink/index.vue
```diff
@@ -96,10 +96,13 @@ export default {
     :model-value="userStore.myLink"
   >
```
devmount, line 97: Can we make that readonly? It might be confusing that you can actually change the value here:

Image

frontend/src/views/AvailabilityView/components/BookingPageLink/index.vue
```diff
@@ -96,10 +96,13 @@ export default {
...
-    <button @click="copyLink">
-      <icon-copy size="12" />
+    <link-button @click="copyLink">
+      <template #iconLeft>
+        <ph-copy size="16" />
+      </template>
+
       {{ copyButtonLabel }}
-    </button>
+    </link-button>
```
devmount, line 105: I think this is totally fine for now. Maybe in the future we want to make that button a bit more suiting into the input. Or extend the input with an action slot or similar. And another note for us: a `user-select:none` would be nice for buttons 😅 

Image

frontend/src/views/AvailabilityView/index.vue
```diff
@@ -182,6 +212,14 @@ export default {
...
   font-size: 2.25rem;
   color: var(--colour-ti-base);
+
+  &.isDirty {
+    margin-block-end: 1.5rem;
+  }
+}
+
+.notice-bar {
+  margin-block-end: 2rem;
```
devmount, line 222: The new notice bar is awesome, thanks for implementing this! I just wondered, if we want to make that sticky? I feel like that would improve UX a lot with just 3 lines of CSS:

```suggestion
  margin-block-end: 2rem;
  position: sticky;
  top: 5rem;
  z-index: 50;
```

Image

frontend/src/views/DashboardView/index.vue
```diff
@@ -97,7 +97,9 @@ export default {
         </i18n-t>
       </li>
     </ul>
-    <primary-button class="dismiss" size="small" @click="dismiss">Dismiss</primary-button>
+    <template #cta>
+      <primary-button class="dismiss" size="small" @click="dismiss">Dismiss</primary-button>
```
devmount: l10n?

frontend/src/views/AvailabilityView/components/AvailabilitySettings/components/AvailabilitySelect.vue
```diff
@@ -327,20 +326,26 @@ const removeAvailability = (option: SelectOption, index: number) => {
   display: flex;
   justify-content: center;
   align-items: center;
-  align-self: flex-start;
```
devmount, line 330: I wonder if this was intended or if we want to get the `flex-start` back? Personally I find it more clear if the select option is on top of all availabilities, especially if we have multiple days with multiple availability entries next to each other.

Image

frontend/src/views/AvailabilityView/components/AvailabilitySettings/components/AvailabilityTimezoneSelect.vue
devmount, line 1: It's really great that we can select the timezone right on the Availability page now 🙌🏻 
For some reason, the saving won't work on my end:

1. Select another timezone
2. Click save on the banner
3. (optional) Reload page
4. See timezone is still the one from before saving

Image

devmount (changes requested):
The availability page looks so much more beautiful now, thanks a lot for all the changes!!
I have some suggestions below and one potential blocker (the timezone saving won't work on my end, but maybe I forgot something).

frontend/src/components/ConfirmationModal.vue
davinotdavid: Indeed, not needed anymore! Thanks!

frontend/src/components/ConfirmationModal.vue
davinotdavid: Great point, I changed that to be the label on both the text content of the button and the `:title` prop, thanks!

frontend/src/views/AvailabilityView/components/BookingPageLink/index.vue
davinotdavid: Oh my, indeed that would be wonderful! However, currently it is not possible unfortunately as we don't forward HTML attribute props to the `input` (they land on the `span` that goes around it). This needs to be changed in `services-ui`.

Opened up a bug ticket here: https://github.com/thunderbird/services-ui/issues/170

frontend/src/views/AvailabilityView/components/BookingPageLink/index.vue
davinotdavid: This should have been fixed by https://github.com/thunderbird/services-ui/issues/101 but it seems that having `user-select: none` on the `button` itself doesn't do it. I think we need to manually apply to the span within it. I've reopened the issue to solve it in `services-ui`. Thanks for catching this!

frontend/src/views/AvailabilityView/index.vue
davinotdavid: You're right, that improves the UX a lot. I think that even if we don't have design confirmation at the moment, we should sneak this in and remove if not needed later on haha thanks!

frontend/src/views/DashboardView/index.vue
davinotdavid: Oops, seems like we didn't have it there. Added, thanks!

frontend/src/views/AvailabilityView/components/AvailabilitySettings/components/AvailabilitySelect.vue
davinotdavid: Absolutely! Not sure why I removed in the first place 😓

frontend/src/views/AvailabilityView/components/AvailabilitySettings/components/AvailabilityTimezoneSelect.vue
davinotdavid: Ah my bad, I was wrongly getting the default timezone from the user but this should be the timezone from the schedule itself. Corrected, thanks!

rwood-moz:
Looks amazing! 🥇 

Just a few small things I noticed while trying it out locally:

- In Booking Page Details the `Page description` input box label looks larger to me than the `Page name` label in the field above it; and in Zeplin the label shows as `Add Page Description`
- Also in Booking Page Details, I'm assuming we just don't have the `Add details` and `Buffer time` features yet right (that are shown in Zeplin)
- Zeplin has `Your Booking Page Link` for the link section header but it actually is `Booking Page Link`
- In the booking page link section beside `Customize your link` Zeplin shows an info button but I don't see it locally (here's from Zeplin):
Screenshot 2025-10-02 at 4 39 11 PM

- I am seeing red asterisks on some headers and am wondering what they mean (and I don't see the asterisks on zeplin):
Screenshot 2025-10-02 at 4 26 49 PM
Screenshot 2025-10-02 at 4 26 56 PM

rwood-moz:
One other thing I just noticed, some strange behaviour with the `Set custom times for each day ` check box. This happens for me:

- Go to the Availability page
- Click the `Set custom times for each day` check box to turn it ON
- The custom days options appears below for each day, as expected
- Turn the `Set custom times for each day` check box OFF again
- The custom days options no longer appear below, as expected
- Now on the `Available days and times` section above, turn ON Sundays
- Then click the `Set custom times for each day` checkbox to turn it ON and notice the options don't appear
- Even if I save changes (after turning on Sundays) clicking the `Set custom times for each day` to turn it on still doesn't show the custom day options below

frontend/src/views/AvailabilityView/components/BookingPageLink/index.vue
devmount: Perfect, thanks for that!

frontend/src/views/AvailabilityView/index.vue
devmount: Nice!

frontend/src/views/DashboardView/index.vue
devmount: Great, thank you!

davinotdavid:
@rwood-moz thanks for the review! Really appreciate it!

> In Booking Page Details the Page description input box label looks larger to me than the Page name label in the field above it; and in Zeplin the label shows as Add Page Description

Ah yes, there is still a bit of a visual difference to be tackled on the `services-ui` side for the inputs (the `TextArea` being the one that needs most of the work which is the Page Description field). The label should be corrected in this PR though, thanks!

> Also in Booking Page Details, I'm assuming we just don't have the Add details and Buffer time features yet right (that are shown in Zeplin)

Yep, we don't have these features yet so that's TBD.

> Zeplin has Your Booking Page Link for the link section header but it actually is Booking Page Link

Good catch, changed! Thanks!

> I am seeing red asterisks on some headers and am wondering what they mean (and I don't see the asterisks on zeplin):

Hmm, good point. I think I can remove the `required` here to remove the asterisks. It is theoretically mandatory but also it is always going to be filled anyways.

frontend/src/views/AvailabilityView/components/AvailabilitySettings/components/AvailabilitySelect.vue
devmount: No worries, I wasn't sure if this was intended by design.

devmount (approved):
Awesome! Thanks for taking care of all my complaints so fast!! 🚀

davinotdavid:
@rwood-moz wow good catch! This seems to also be happening on `stage` at the moment so I suspect this issue already existed before. I opened up a ticket for it so we can tackle it separately: https://github.com/thunderbird/appointment/issues/1284

rwood-moz:
> @rwood-moz wow good catch! This seems to also be happening on `stage` at the moment so I suspect this issue already existed before. I opened up a ticket for it so we can tackle it separately: #1284

Ahh ok great, thank you

rwood-moz:
> Hmm, good point. I think I can remove the `required` here to remove the asterisks. It is theoretically mandatory but also it is always going to be filled anyways.

Thanks! I still see the required on the `Meeting duration` control FYI

davinotdavid:
@rwood-moz oh no sorry about that, I thought it was good to go! I removed it in this commit here: https://github.com/thunderbird/appointment/pull/1283/commits/47033f549075e52593372e9c6e13ae5538ed35a5 for the Bookings Page 🙇

rwood-moz:
> @rwood-moz oh no sorry about that, I thought it was good to go! I removed it in this commit here: [47033f5](https://github.com/thunderbird/appointment/commit/47033f549075e52593372e9c6e13ae5538ed35a5) for the Bookings Page 🙇

haha no worries, perfect, thank you :)

PR #1280, opened 2025-10-01
Delete normalize secrets

ryanjjung:
## Description of the Change

This ought to be the last of the removal of the `normalize_secrets` function.

## Benefits

Tidier code, more granular configuration options.

## Applicable Issues

#1110


ryanjjung:
I ran a `ruff format` on this code and it fixed a few other issues unrelated to my changes. Not sure how they got there, though. We have a validation check on PR creation that runs both `ruff format` and `ruff check`. It should fail when there are linter problems.

PR #1281, opened 2025-10-01
Revert "Delete normalize secrets"

ryanjjung:
Reverts thunderbird/appointment#1280

Pertains to #1110


davinotdavid (approved):

davinotdavid:
Just for tracing, this is because the old TF code is using this so it keep error-ing in Sentry.

Issue #1275, opened 2025-09-29
Change authentication to point to accounts.tb.pro


malini:
**Is your feature request related to a problem? Please describe.**
For TB Pro, we will be using accounts.tb.pro instead of FXA.

Update login/signup flow accordingly, and update tests, but do no merge. This will sit in the wings until we're ready to do the Neon migration

davinotdavid:
Ok good news is that this seems to be already implemented by @MelissaAutumn 

I had a hunch that I've seen this working before at least on staging and here was the PR:

https://github.com/thunderbird/appointment/pull/1105

Also, [appointment.tb.pro](https://appointment.tb.pro) is already pointing to Keycloak but the client has not been created yet so I went ahead and created it! 

So now it is possible to login using @thundermail.com but there's currently some CORS errors around it. Will continue to investigate!

davinotdavid:
@malini Since the infrastructure behind `appointment.day` and `appointment.tb.pro` is different, I went ahead and already merged the authentication change to `oidc` / `keycloak` / `accounts.tb.pro`.

So `appointment.day` is still using `fxa` and `appointment.tb.pro` is now using `oidc`. Here's the changes made:
- Changed `AUTH_SCHEME` env var to `OIDC`
- Created clients for both frontend and backend in Keycloak `tb.pro` realm in production
- Added related env vars and secrets to the new clients
- Fixed a Google Auth env var that prevented the connection

I can verify that I am able to login to `appointment.tb.pro`, connect my Google Account and connect Zoom!

Closing this issue as this is live in prod but feel free to reopen if needed.

PR #1282, opened 2025-10-01
Change AUTH_SCHEME to OIDC

davinotdavid:
## Description of the Change

For `prod` / `appointment.tb.pro`, I've created a Keycloak client id for Appointment Frontend but I can't actually login since the `AUTH_SCHEME` is set to `fxa` instead of `oidc`.

All the code to login through oidc seems to be in place as I can verify that pointing my local to the stage realm works well at the moment.

Can this be added only for the `appointment.tb.pro` and not to `appointment.day`?

## Benefits

- Hopefully be able to login in `appointment.tb.pro` using keycloak.

## Applicable Issues

https://github.com/thunderbird/appointment/issues/1275

### DIFF
--- a/pulumi/config.prod.yaml
+++ b/pulumi/config.prod.yaml
@@ -227,7 +227,7 @@ resources:
               - name: APP_ENV
                 value: prod
               - name: AUTH_SCHEME
-                value: "fxa"
+                value: oidc
               - name: DATABASE_ENGINE
                 value: postgresql
               - name: DATABASE_PORT


ryanjjung (approved):
That should do it. The `appointment.day` stuff is managed in the `tofu` directory, so this should only affect the `.tb.pro` infrastructure.

davinotdavid:
@ryanjjung thanks! I realized that I have to add some other things as well:

- 2 secrets to Pulumi: `oidc-client-id` and `oidc-client-secret`
- 2 secrets to the task definition / container definition: OIDC_CLIENT_ID and OIDC_CLIENT_SECRET
- 2 env vars to the task definition / container definition: `OIDC_EXP_GRACE_PERIOD` and `OIDC_TOKEN_INTROSPECTION_URL`

If I run `pulumi config set --secret oidc-client-id CLIENT_ID` and `pulumi config set --secret oidc-client-secret CLIENT_SECRET` it does alter `Pulumi.prod.yaml` file with secrets but which values should I populate the `config.prod.yaml` files / task definitions with? Should I create the secrets in SecretsManager directly? if so which KMS key should I use?

Sorry to ask again for a walk through 😅 just want to make sure that I don't mess it up on prod!

davinotdavid:
@ryanjjung @aatchison 

It seems that the `pulumi preview` has quite a bit of changes regarding `IAM` somehow and given that we had some IAM permission issues recently, I am not sure if I did something wrong:

https://app.pulumi.com/mzla-services/appointment/prod/previews/e6efccab-1787-4c96-8b6d-d684879bb7db

ryanjjung:
Two things...

First, the same names you used in the `pulumi config set --secret` commands should go in the `PulumiSecretsManager` list. [Ref](https://thunderbird.github.io/pulumi/patterns.html#handling-secrets)

Second, you have a lot of extra stuff in the list because you have not disabled resource protection in the prod environment (meant to be a catch/statement of intent against accidentally changing prod). I just looked, and this is not a documented feature, which is my fault. I'll write something up about that in the near future. However, the output shows a lot of this message:

```
Resource protection has been enabled on [secret]-prod-autoscl-backend. To disable, export TBPULUMI_DISABLE_PROTECTION=True
```

So try again after running `export TBPULUMI_DISABLE_PROTECTION=True`.

You may still see a bunch of "sap" related stuff getting deleted. This is a known issue right now. You can use `--target` and `--target-dependents` options to filter down to the task definition, and the end result should be that applying that change by itself will cause Pulumi to not want to delete those other items. [Here is the command](https://github.com/thunderbird/appointment/blob/main/.github/workflows/deploy-production.yml#L350-L353) to run in prod.

ryanjjung:
More info: The `--target` option wants a URN, which is Pulumi's internal reference for each resource. For resources that have already been created, you can use `pulumi stack --show-urns` to see them. For example:

```sh
pulumi stack --show-urns | grep PulumiSecretsManager
```

Grab the top level item: `urn:pulumi:stage::appointment::tb:secrets:PulumiSecretsManager::appointment-stage-secrets`

And use that with `--target` and `--target-dependents` to target all secrets managed by the secrets manager.

One you create the secret, you can grab the ARN of that secret from the web console or AWSCLI and set that as the `valueFrom` part of the secret in the task definition.

davinotdavid:
@ryanjjung thank you so much, that's super helpful! I did the `export TBPULUMI_DISABLE_PROTECTION=True` and ran a `pulumi preview --diff` again and it looks much more sane now (although still lots of sap things):

https://app.pulumi.com/mzla-services/appointment/prod/previews/e4f90723-4b47-4a64-82df-ef6d5bdb658f

Also, I've already added the secrets in the `PulumiSecretsManager` list in the previous commit here: https://github.com/thunderbird/appointment/pull/1282/files#diff-7e56ca5a413822970c716a405bebd8a0dfb525a181e29ff672cb7dc8d681ff8aR104-R105

After this gets approved do I run a `pulumi up` so that it actually creates the secrets in AWS and I get the URNs to fill in the task definition part?

ryanjjung:
Yes, `pulumi up` with targeting so as to only snag the task def. Once that is done, the sap stuff will go away.

davinotdavid:
@ryanjjung ~~hmmm I tried running this:~~

```
pulumi up --target 'urn:pulumi:prod::appointment::tb:secrets:PulumiSecretsManager$tb:secrets:SecretsManagerSecret::appointment-prod-secrets-oidc-client-id' --target 'urn:pulumi:prod::appointment::tb:secrets:PulumiSecretsManager$tb:secrets:SecretsManagerSecret::appointment-prod-secrets-oidc-client-secret'
```

~~And even though it did show only the secrets that I want to add, the update failed:~~

~~https://app.pulumi.com/mzla-services/appointment/prod/updates/49~~

**EDIT**: Nvm, I am dumb and didn't run the correct command 😅 It seems to be created now in Secrets Manager, thank you SO much for the patience again 🙏

Issue #1279, opened 2025-09-30
Can't do booking filtering when browser window isn't full size


ryanleesipes:
**Describe the bug**
Some of the UI is blocked by the top bar when window is not full size. Can't see the "Booking" header or the filtering options.

**To Reproduce**

1. Open Bookings screen in Appointment
2. Make the window half-size
3. You will not be able to see elements at the top of the page.

**Expected behavior**
Be able to see and interact with the filtering options.

**Screenshots**

<img width="1003" height="195" alt="Image" src="https://github.com/user-attachments/assets/79865bbc-f6ee-48dd-bd06-1c6f067831ff" />

**System:**

- OS: Windows 11
- Browser Version: Firefox 143.0.1

**Additional context**
This is what I expected to be able to see at the top of page, but can't scroll up any higher:

<img width="3358" height="62" alt="Image" src="https://github.com/user-attachments/assets/5d4d5e21-0bbd-4ed4-95ae-291e5abe51e4" />

devmount:
Thank you Ryan for reporting this. The whole layout already got a big update, but I think this is not yet in production. The issue you're seeing is already fixed there:

<img width="927" height="685" alt="Image" src="https://github.com/user-attachments/assets/962cc3c7-0f02-4b7d-99e0-47a6ecdad0b0" />

Issue #1256, opened 2025-09-16
Update Bookings page with new hi-fi designs


davinotdavid:
**Is your feature request related to a problem? Please describe.**
As the Design team has updated the designs on the Bookings page, we should apply those to make the app coherent design-wise.

Zeplin link: https://app.zeplin.io/project/66dd0f2b2fe8f1b63cda36f1/screen/68c46c27dbe4b6af5120ba0a

Note that this is not only a fresh coat of paint but also it changes some existing functionality:
- There are no dropdown filters for appointment status anymore
- There doesn't seem to be re-ordering by clicking on table columns anymore
- There is now an omni-search across appointments. TBD on how this works (what can it be searched for, etc)

**Describe the solution you'd like**
Match the design below as close as possible:

<img width="1280" height="1519" alt="Image" src="https://github.com/user-attachments/assets/e55d4655-69ff-4028-92eb-eea1676028a0" />

PR #1283, opened 2025-10-02
Update Bookings page with new hi-fi designs

davinotdavid:
## Description of the Change

This change removes the existing Bookings page completely and replaces it with the newly designed one:

apmt_bookings_03
apmt_bookings_04

apmt_bookings_01
apmt_bookings_02

## Known issues

The badge counter in the SegmentedControl component on the top needs to be implemented in Services UI first:
https://github.com/thunderbird/services-ui/issues/175

## Applicable Issues

Closes https://github.com/thunderbird/appointment/issues/1256
Closes https://github.com/thunderbird/appointment/issues/1279

### DIFF
--- a/frontend/src/views/BookingsView/components/DateRequestedAppointmentItem.vue
+++ b/frontend/src/views/BookingsView/components/DateRequestedAppointmentItem.vue
@@ -1,30 +1,39 @@
 <script setup lang="ts">
-import type { Dayjs } from 'dayjs';
 import { computed, inject } from 'vue';
 import { useI18n } from 'vue-i18n';
 import { dayjsKey } from '@/keys';
 import { BaseBadge, BaseBadgeTypes } from '@thunderbirdops/services-ui';
+import type { Appointment } from '@/models';
+import { isUnconfirmed } from '@/utils';
 
 const props = defineProps<{
-  name: string;
-  email: string;
-  requestDate: string;
-  meetingDate: string | Dayjs;
-  needsConfirmation: boolean;
+  appointment: Appointment;
 }>();
 
 const dj = inject(dayjsKey);
 const { t } = useI18n();
 
-const formattedRequestDate = computed(() => dj(props.requestDate).format('L, LT'));
-const formattedMeetingDate = computed(() => dj(props.meetingDate).format('L, LT'));
+const appointmentItem = computed(() => {
+  return {
+    name: props.appointment.slots[0].attendee.name,
+    email: props.appointment.slots[0].attendee.email,
+    requestDate: props.appointment.time_created,
+    meetingDate: props.appointment.slots[0].start,
+    needsConfirmation: isUnconfirmed(props.appointment),
+  }
+});
+
+const formattedRequestDate = computed(() => dj(appointmentItem.value.requestDate).format('L, LT'));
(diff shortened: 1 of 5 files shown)


frontend/src/stores/appointment-store.ts
```diff
@@ -26,13 +25,37 @@ export const useAppointmentStore = defineStore('appointments', () => {
...
+
+        if (start.isToday()) {
+          today.push(appointment);
+        } else if (start.isSame(now, 'week')) {
+          thisWeek.push(appointment);
+        } else if (start.isSame(now, 'month')) {
+          thisMonth.push(appointment);
+        } else {
+          earlier.push(appointment);
+        }
```
rwood-moz, line 53: What happens if the appointment slot start date is actually next month / one month in the future?  Looks like it would be added to the 'earlier' category. This could happen if the current date is close to the end of a month and the booking request window is open for the start of the next month.

rwood-moz:
Looks awesome!

When trying this out locally I notice that the `Sort by` setting is not sticky i.e.:

- Go to the Bookings tab and notice `Sort by` is set to `Date requested`
- Change the `Sort by` value to `Meeting date`
- Bookings page updates as expected
- Now click on a different tab like `Availability`
- Click back on the `Bookings` tab
- Notice the `Sort by` value has now changed back to `Date requested` and the page is displayed accordingly

Should we consider making this sticky? Just thinking a user might like to have the bookings always sorted by `Meeting date` but they will need to constantly select that option every time they view the bookings page.

frontend/src/stores/appointment-store.ts
davinotdavid: Good point! And yeah, it would currently put in the `earlier` category. I am not sure if I fully understand the design decision behind it but it seems to be the granularity that the designs show here: https://app.zeplin.io/project/66dd0f2b2fe8f1b63cda36f1/screen/68dd4daa64e9e012388d6bb4. Perhaps `earlier` is not a great name here in case it doesn't fit in other categories!

davinotdavid:
@rwood-moz I think that would be a great enhancement, for sure! It does introduce some oddities / potential bugs in terms of having the state of the select filters in the URL like we had before so perhaps we should include it in a next PR / separate ticket.

Also, there's a bit of functionality missing here (clicking the appointment should open that sliding panel with the confirm / cancel, for example) so I am wrapping it up soon for a re-review! 🙏

rwood-moz:
> @rwood-moz I think that would be a great enhancement, for sure! It does introduce some oddities / potential bugs in terms of having the state of the select filters in the URL like we had before so perhaps we should include it in a next PR / separate ticket.
> 
> Also, there's a bit of functionality missing here (clicking the appointment should open that sliding panel with the confirm / cancel, for example) so I am wrapping it up soon for a re-review! 🙏

Sounds good!

frontend/src/stores/appointment-store.ts
rwood-moz: Yes, maybe there should be a `Next month` category; although if you have `Sort by` set to `Meeting date` then you can scroll to the next month(s) to see the appointments there, so at least the booking can still be seen.

davinotdavid:
@rwood-moz cool! I've pushed the changes to make the list items to be buttons instead so you can click on them to open the sliding panel with the booking actions and did some CSS cleanup. Could you do a re-review? 🙏

PR #1289, opened 2025-10-03
Fully remove `normalize_secrets`; stop deploying changes to `appointment.day` domain

ryanjjung:
## Description of the Change

This removes the `normalize_secrets` function while also disabling deployments to all of the Terraform-backed infrastructure. This is with the expectation that these environments are short-lived and that we are now encouraging users to use the Pulumi-backerd infrastructure.

## Benefits

- Tidier code with more granular configuration control
- Stop supporting the legacy infrastructure

## Applicable Issues

#1110


.github/workflows/validate.yml
```diff
@@ -32,9 +32,6 @@ jobs:
       id: check
       with:
         filters: |
-          validate-iac:
-            - 'tofu/**'
```
davinotdavid, line 36: This goes beyond the scope of this PR / issue but does this mean that we could also remove the `tofu` folder altogether?

davinotdavid (approved):

.github/workflows/validate.yml
ryanjjung: I'm going to leave the `tofu` folder here for now because we will need it to run all the "destroy" commands to tear down this infrastructure later. Here, I'm indicating that I no longer care if anything in that folder changes, as it pertains to CI.

Issue #1276, opened 2025-09-29
Import users from appointment into keycloak


malini:
**Is your feature request related to a problem? Please describe.**
When we migrate to accounts.tb.pro, we will want each existing user of appointment to have an account setup. Determine how we can migrate the users to keycloak as verified users, and then wait until we deploy https://github.com/thunderbird/appointment/issues/1275 before hitting the 'reset password' url

davinotdavid:
From an initial research, I see a few strategies to accomplish this but there are some unknowns that I am still investigating.

Assuming that we are able to access a backend container in the various environments, we can write a new `command` like `backend/src/appointment/commands/export_users_to_keycloak.py` and:

- Manipulate the output format as desired (e.g. JSON, CSV, etc) and use a third-party plugin / Keycloak script to bulk import or;

- Make a POST request _per user_ to [Keycloak's Admin REST API](https://www.keycloak.org/docs-api/latest/rest-api/index.html#_users) through their `POST /admin/realms/{realm}/users` endpoint.
  - This is a built-in approach without any plugins but apparently users are created one by one and the only built-in bulk import that I could find in during an "import realm" which is probably not what we want.
  - Q: `Timezone` doesn't seem to be included in [Keycloak's `UserRepresentation` object](https://www.keycloak.org/docs-api/latest/rest-api/index.html#UserRepresentation) but it is required in the Keycloak's admin web interface. Unsure if this is a metadata of sorts?

--

Another option is to use https://github.com/daniel-frak/keycloak-user-migration plugin as @MelissaAutumn suggested which requires us to implement two endpoints (a GET for user data and a POST for password check). This seems to be a great approach as the import would happen "on-demand" / when the user actually logs in.

However, this requires us to keep the FXA infrastructure around so that when the user attempts to login through Keycloak, they would need to successfully authenticate through FXA to do the "handshake" of being granted access to retrieve the user data in the GET request and validate the password in the POST request for the plugin.

--

Even though the first option would be a one-off run, I wonder if it is viable given the amount of users to migrate? Or if there are some other strategies that I am missing? 🤔

MelissaAutumn:
I'd like to limit the admin api access of Keycloak to just Accounts if possible. 

We have user creation in the Accounts admin panel hooked up with Keycloak user creation, so if we can get a dump of email addresses we can more easily build an importer on that side. 

I have a handy-dandy ticket here for that issue: https://github.com/thunderbird/thunderbird-accounts/issues/260

[...]

PR #1290, opened 2025-10-06
Export users to keycloak via Admin API

davinotdavid:
## Description of the Change

Example on how to use the [Keycloak Admin REST API](https://www.keycloak.org/docs-api/latest/rest-api/index.html#_users) to create users from our backend.

This assumes that it is possible to run a command inside of the backend container in the various environments.

```
run-command main export-users-to-keycloak
```

I am still unsure on how to create / update custom fields like `timezone`. We have this information already but the Keycloak Admin REST API doesn't seem to have a field for that.

## Benefits

Built-in way of importing users into Keycloak, no third party plugins needed for that.

## Applicable Issues

https://github.com/thunderbird/appointment/issues/1276

### DIFF
--- a/backend/src/appointment/commands/export_users_to_keycloak.py
+++ b/backend/src/appointment/commands/export_users_to_keycloak.py
@@ -1,0 +1,80 @@
+import logging
+import os
+import requests
+from ..database import models
+from ..dependencies.database import get_engine_and_session
+
+
+def export_users() -> None:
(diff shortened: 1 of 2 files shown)


markstos:
Closing as linked issue is closed. If we re-open, we should re-consider security implications of giving our app Keycloak admin access, especially if it can't be narrowly scoped.

PR #1292, opened 2025-10-07
Remove Terraform references from release process

ryanjjung:
## Description of the Change

In a [previous PR](https://github.com/thunderbird/appointment/pull/1289), I removed a lot of Terraform references from this project's CI processes. Unfortunately, I missed a few things related to releases, and that [caused some CI failures](https://github.com/thunderbird/appointment/actions/runs/18318770647/job/52166532489). This PR removes further Terraform references from the release process.

## Benefits

CI should hopefully work again.

## Applicable Issues

#1110


.github/workflows/deploy-staging.yml
```diff
@@ -119,17 +118,8 @@ jobs:
         with:
           python-version: 3.13
 
-      - name: Set up virtual environment
```
ryanjjung, line 122: Pulumi manages its own virtual environment. There is no need for us to do this work here.

frontend/.env.prod.example
```diff
@@ -1,21 +1,21 @@
...
 VITE_SENTRY_DSN=https://66f57328baac4929be7d62b77f41b510@o4505428107853824.ingest.sentry.io/4505428280279040
 
 # -- Auth scheme --
-VITE_AUTH_SCHEME=fxa
+VITE_AUTH_SCHEME=oidc
 
-# For fxa/oidc
-VITE_AUTH_EDIT_PROFILE=https://accounts.firefox.com/settings
+# For fxa
+VITE_FXA_EDIT_PROFILE=https://accounts.tb.pro/self-serve
```
davinotdavid: This is being consumed [here](https://github.com/thunderbird/appointment/blob/main/frontend/src/App.vue#L53) as `VITE_AUTH_EDIT_PROFILE` and eventually gets linked in the "Edit Profile" button in the Profile page.

While the profile page is for a replacement, if we want to keep the button working for now I believe we should do this instead:

```
# For oidc
VITE_AUTH_EDIT_PROFILE=https://accounts.tb.pro/self-serve
```

Same for the `.env.stage.example`

ryanjjung: Hrm, ok. Lemme take a closer look. I just `git mv`'d those files over the ones the Terraform processes used. I'll see if I missed anything else.

ryanjjung: I applied the changes you suggested. I also did some Bash magic to get a list of env-vars present in the before and after versions of the files. It appears the two OIDC options are new, but were only being applied to the tb.pro environments. I don't know if that caused any problems in the TF envs.

davinotdavid: Yeah, I've added it recently as part of https://github.com/thunderbird/appointment/pull/1282. Unsure if that causes problems in the TF envs but possibly, sorry about that 😅

davinotdavid (approved):

PR #1306, opened 2025-10-24
Add some frontend vue.js component tests

### LINKED ISSUE #1202: Add a vue.js component test (and investigate running in CI)
There is a need to test some UI elements that are not covered in the E2E tests (E2E tests are for the most common user flows only because they take a lot of time to run and maintain). Investigate adding a first vue.js component test and running it in CI. Separate tickets can then be created to expand a vue.js components test suite.

An example would be a vue.js component test suite that covers the FTUE.

rwood-moz:
Fixes #1202. Just add some starting vue.js UI component tests. These should be built out further over time.

### DIFF
--- a/frontend/test/components/ExternalRedirect.test.js
+++ b/frontend/test/components/ExternalRedirect.test.js
@@ -1,0 +1,38 @@
+import { vi, describe, it, expect } from 'vitest';
+import { mount } from '@vue/test-utils'
+import ExternalRedirect from '@/components/ExternalRedirect.vue';
+import i18ninstance from '@/composables/i18n';
+
+
+describe('ExternalRedirect', () => {
+  it('renders correctly', () => {
+    var ourProps = {
+        redirectUrl: 'tb.pro',
+    }
+
+    // mock the window.location object for specific properties that the component will call
+    vi.stubGlobal('location', {
+      assign: vi.fn(),
+      replace: vi.fn(),
+      href: 'http://localhost/', // Set a default value if needed
+    });
+
+    const wrapper = mount(ExternalRedirect, {
+      propsData: ourProps,
+      global: {
+        plugins: [i18ninstance],
+      },
+    });
+
+    // verify the redirect link was created
+    expect(wrapper.props()).toEqual(ourProps);
+    const linkWrapper = wrapper.find('a');
+    expect(linkWrapper.exists()).toBe(true);
+    expect(linkWrapper.text()).toBe(ourProps.redirectUrl);
+    expect(linkWrapper.attributes('href')).toBe(ourProps.redirectUrl);
+
+    // clean up our mock and unmount the component
+    vi.unstubAllGlobals()
+    wrapper.unmount();
+  });
+});
--- a/frontend/test/components/NavBarMobile.test.js
+++ b/frontend/test/components/NavBarMobile.test.js
@@ -1,0 +1,143 @@
+import { vi, describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { mount } from '@vue/test-utils'
+import { createTestingPinia } from '@pinia/testing';
+import i18ninstance from '@/composables/i18n';
+import router from '@/router';
+import withSetup from '../utils/with-setup';
+import { useUserStore } from '@/stores/user-store';
+import { RouterLink } from 'vue-router';
+import NavBarMobile from '@/components/NavBarMobile.vue';
+
+
+describe('NavBarMobile', () => {
+  var app;
+  var wrapper;
+
+  beforeEach(() => {
+    app = withSetup();
+    app.use(createTestingPinia());
+  });
+
+  afterEach(() => {
+    wrapper.unmount();
+  });
+
+  it('renders correctly when not logged in', () => {
+    wrapper = mount(NavBarMobile, {
+      global: {
+        plugins: [i18ninstance, router],
+      },
+    });
+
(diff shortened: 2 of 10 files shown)


rwood-moz:
```
Roberts-MacBook-Pro test % npm test components 
...
 Test Files  7 passed (7)
      Tests  22 passed (22)
   Start at  10:54:03
   Duration  1.35s (transform 556ms, setup 61ms, collect 4.22s, tests 222ms, environment 1.34s, prepare 236ms)
```

frontend/test/components/ExternalRedirect.test.js
```diff
@@ -0,0 +1,36 @@
...
+    var ourProps = {
+        redirectUrl: 'tb.pro',
+    }
+
+    // mock the window.location object for specific properties that the component will call
+    vi.stubGlobal('location', {
+      assign: vi.fn(),
+      replace: vi.fn(),
+      href: 'http://localhost/', // Set a default value if needed
+    });
```
davinotdavid, line 18: This works well but I believe we need to "unstub" / cleanup the global stub here to not affect future tests that might need different values. According to [the docs](https://vitest.dev/guide/mocking/globals.html#mocking-globals), we can either manually call [vi.unstubAllGlobals()](https://vitest.dev/api/vi.html#vi-unstuballglobals) or have a config to auto unstub after each test.

frontend/test/components/NavBarMobile.test.js
```diff
@@ -0,0 +1,143 @@
...
+
+  it('able to click the copy link button', async () => {
+    // we need to mock the clipboard so the rendered component can copy to clipboard
+    Object.defineProperty(navigator, 'clipboard', {
+      value: {
+        writeText: vi.fn(() => Promise.resolve()),
+        readText: vi.fn(() => Promise.resolve('Mocked content')),
+      },
+      configurable: true,
+    });
```
davinotdavid, line 116: Just for curiosity, does this define also persist through other tests? If so I think we might need to do a cleanup as well just in case!

frontend/test/components/confirmation-modal.test.js
```diff
@@ -0,0 +1,95 @@
...
+
+  it('renders correctly when modal is open', () => {
+    const wrapper = mount(ConfirmationModal, {
+      propsData: ourProps,
+      global: {
+        plugins: [i18ninstance], // install the i18ninstance as a global plugin
+      },
+    });
+    expect(wrapper.props()).toEqual(ourProps);
+    wrapper.unmount();
```
davinotdavid: Nice usage of the `beforeEach` there! I believe we can also leverage [afterEach](https://vitest.dev/api/#aftereach) to do the unmount / cleanup of stubs so that we don't have to repeat `wrapper.unmount()`.

rwood-moz:
Thanks for the review and input!

> 
>     I see that we have a `package.json` script for `test` but I just wanted to double check that it is already automagically picked up by our CI to be ran or if there's anything else needed there?
 
Yes the tests that I am adding run with the other [pre-existing frontend tests](https://github.com/thunderbird/appointment/tree/main/frontend/test/stores) which already run in CI; the github validate frontend job [does an npm install ](https://github.com/thunderbird/appointment/blob/main/.github/workflows/validate.yml#L103)that uses the `frontend/package.json` which includes the test dependencies.

>     (Nit) The actual component file names are written as `TabBar` instead of `tab-bar`. I wonder if we should keep the file naming consistent here or is this a limitation of sorts for testing purposes?

Ah yes, I noticed that too but decided to keep the same naming format as the already existing tests in https://github.com/thunderbird/appointment/tree/main/frontend/test/stores, however I see the stores are actually named with `store-name.ts` so that's why. I'll change my new tests names to match the component `TabBar` format, makes sense.

> Oh also there are some conflicts on the `package-lock.json` file to be resolved before merge!

Ok thanks I didn't see this locally but I'll have a look.

frontend/test/components/ExternalRedirect.test.js
rwood-moz: Great point, I'll do that thank you.

frontend/test/components/NavBarMobile.test.js
rwood-moz: TBH I'm not sure (this is the first time I've written vue.js component tests haha) but that's a great point, I would think the same (I'll look into it further) thanks.

frontend/test/components/confirmation-modal.test.js
rwood-moz: Good catch, I'll add that thank you.

frontend/test/components/NavBarMobile.test.js
rwood-moz: I looked into this further, so the reason I had to add this in the first place was because the test environment (JSDOM) doesn't provide a navigator.clipboard, so in order to mount the component that uses the clipboard (and get the clipboard value in the test) I had to create it using the `defineProperty`. Also properties added this way are persisted between tests, so I will destroy the clipboard after the test is done. :)

rwood-moz:
Thanks for the review @davinotdavid, updates are complete and ready for another review whenever you have some time.

```
Roberts-MacBook-Pro test % npm test components 
...
 Test Files  7 passed (7)
      Tests  22 passed (22)
   Start at  14:45:42
   Duration  1.43s (transform 580ms, setup 59ms, c
```

Issue #1323, opened 2025-11-07
Can't login into Stage


davinotdavid:
**Describe the bug**
When trying to login into `appointment-stage.tb.pro`, the `/post-login` route errors out and doesn't show any information on the screen.

Sentry error:
https://thunderbird.sentry.io/issues/6977487987/?project=4505428124827648&query=is%3Aunresolved&referrer=issue-stream

This happens because there is a mismatch between the `oidc_id` coming from the access token on post-login and the `oidc_id` that we have on the DB. The Subscriber lookup fails then it attempts to create a brand new Subscriber. However, there are cases where the Subscriber already exists in our DB and attempting to create a new one will cause a DB error.

davinotdavid:
As an update to this issue, the code fix in https://github.com/thunderbird/appointment/pull/1324 actually only matches by email in case the user does not have an `ExternalConnection` of type `oidc` in the Appointment database yet so this fixes the connection.

In case there's a user that has an `ExternalConnection` entry of type `oidc` in the Appointment database and that **mismatches** with the incoming Keycloak `oidc_id` coming from the access token introspection, it will still fail the same way but that should be a super exception one-off case. This was my particular case.

When that happens, we have to manually delete the `ExternalConnection` entry in the Appointment database in Neon and the connection will fix itself in the next login.

PR #1324, opened 2025-11-07
Add email lookup for `oidc_id` mismatch (OIDC_FALLBACK_MATCH_BY_EMAIL)

davinotdavid:
## Description of the Change

- This PR adds `OIDC_FALLBACK_MATCH_BY_EMAIL` env var (default as `True`) so that for existing / returning users (meaning that there's already a `Subscriber` row in the DB for that email) that might be attempting to login and may have a different `oidc_id`, we can try to match them by `secondary_email` and allow login instead of attempting to create a brand new `Subscriber` and breaking the unique DB constraint.

This (somewhat) follows the existing idea used in `thunderbird-accounts`: https://github.com/thunderbird/thunderbird-accounts/blob/main/src/thunderbird_accounts/authentication/middleware.py#L77

## Benefits

- Existing users will be able to login as there won't be attempts to create duplicated Subscribers in our DB upon post-login

- There's a failing test happening on `main` as well but it is unrelated with this (it is related with timezones):

```
FAILED test/integration/test_schedule.py::TestSchedule::test_public_availability - AssertionError: assert '2024-03-04T08:00:00-08:00' == '2024-03-04T09:00:00-08:00'
```

## Applicable Issues

https://github.com/thunderbird/appointment/issues/1323

### DIFF
--- a/backend/src/appointment/database/repo/external_connection.py
+++ b/backend/src/appointment/database/repo/external_connection.py
@@ -136,13 +136,21 @@ def get_subscriber_by_oidc_id(db: Session, type_id: str):
     return None
 
 
-def get_subscriber_by_email(db: Session, email: str):
-    """Return a subscriber by the OIDC recovery email address"""
+def get_subscriber_without_oidc_by_email(db: Session, email: str):
+    """Return a subscriber by the OIDC recovery email address without an OIDC connection"""
+    # Subquery to check if a subscriber has an OIDC external connection
+    oidc_exists = (
+        db.query(models.ExternalConnections.id)
+        .filter(models.ExternalConnections.owner_id == models.Subscriber.id)
+        .filter(models.ExternalConnections.type == models.ExternalConnectionType.oidc)
+        .exists()
+    )
+    
+    # Find subscribers without an OIDC connection that match the email
     query = (
         db.query(models.Subscriber)
-        .filter(models.Subscriber.secondary_email == email)
+        .filter(~oidc_exists)
+        .filter(models.Subscriber.email == email)
     )
 
-    result = query.first()
-
-    return result
+    return query.first()
--- a/backend/src/appointment/routes/auth.py
+++ b/backend/src/appointment/routes/auth.py
@@ -661,8 +661,8 @@ def oidc_token(
 
     # If we can't find the subscriber by oidc_id, try to find them by email
     # as a fallback in case the OIDC_FALLBACK_MATCH_BY_EMAIL env var is set to True
-    if os.getenv('OIDC_FALLBACK_MATCH_BY_EMAIL') == 'True' and not subscriber:
-        subscriber = repo.external_connection.get_subscriber_by_email(db, email)
+    if os.getenv('OIDC_FALLBACK_MATCH_BY_EMAIL', 'false').lower() == 'true' and not subscriber:
+        subscriber = repo.external_connection.get_subscriber_without_oidc_by_email(db, email)
 
     if not subscriber:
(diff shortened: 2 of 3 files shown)


backend/src/appointment/database/repo/external_connection.py
```diff
@@ -134,3 +134,15 @@ def get_subscriber_by_oidc_id(db: Session, type_id: str):
         return result.owner
 
     return None
+
+
+def get_subscriber_by_email(db: Session, email: str):
+    """Return a subscriber by the OIDC recovery email address"""
+    query = (
+        db.query(models.Subscriber)
+        .filter(models.Subscriber.secondary_email == email)
```
MelissaAutumn: is primary email their thundermail address in our glorious thunderbird pro future?

backend/src/appointment/database/repo/external_connection.py
```diff
@@ -134,3 +134,15 @@ def get_subscriber_by_oidc_id(db: Session, type_id: str):
...
+    )
```
MelissaAutumn, line 154: can you also add .filter(models.Subscriber.oidc_id is None) so we don't match folks who should have passed in the first place.

backend/src/appointment/database/repo/external_connection.py
```diff
@@ -134,3 +134,15 @@ def get_subscriber_by_oidc_id(db: Session, type_id: str):
...
+
+    result = query.first()
+
+    return result
```
MelissaAutumn: You can just return query.first().

backend/src/appointment/routes/auth.py
```diff
@@ -658,6 +658,12 @@ def oidc_token(
     name = token_data.get('name')
 
     subscriber = repo.external_connection.get_subscriber_by_oidc_id(db, oidc_id)
+
+    # If we can't find the subscriber by oidc_id, try to find them by email
+    # as a fallback in case the OIDC_FALLBACK_MATCH_BY_EMAIL env var is set to True
+    if os.getenv('OIDC_FALLBACK_MATCH_BY_EMAIL') == 'True' and not subscriber:
```
MelissaAutumn:
```suggestion
    if os.getenv('OIDC_FALLBACK_MATCH_BY_EMAIL', 'false').lower() == 'true' and not subscriber:
```

MelissaAutumn: I think that's correct, double check that...we want to avoid caps errors 😄

MelissaAutumn (changes requested):
Awesome, thanks for getting this up so fast. Just some nits!

backend/src/appointment/database/repo/external_connection.py
davinotdavid: Apparently we are considering that `secondary_email` may not exist and relying on the `Subscriber.email` for everything so I've changed this to be `Subscriber.email` instead. Thanks!

backend/src/appointment/database/repo/external_connection.py
davinotdavid: Subscriber doesn't have direct access to `oidc_id` but I made a subquery to check against the `ExternalConnections` table as well. Thanks!

backend/src/appointment/routes/auth.py
davinotdavid: Ah yes, I vaguely remember seeing this error before 😅 thanks!

backend/src/appointment/database/repo/external_connection.py
MelissaAutumn: Oh right. Awesome, thanks!

PR #1332, opened 2025-11-12
Changes so uv will build appointment correctly.

### LINKED ISSUE #756: Switch to uv
Low priority!

We should switch over from pip to uv
https://docs.astral.sh/uv/

It should be a fairly easy switch, we'll just need to `uv sync` the directory and commit the lock file. 

Then whenever we want to add a project to requirements we'll just need to use `uv add` instead of `pip install`. 

We'll have to adjust our docker container to also use uv or convert the lockfile on the fly with `uv export --format requirements-txt`.

MelissaAutumn:
Some changes I had to make to make [uv](https://docs.astral.sh/uv/) actually work with appointment.

We use it in our other python projects so this might help resolve #756. 

It'll need some more testing and documentation though.

Our build system uses setuptools which is no longer included in python but gets pulled in via requires iirc, and we were using an outdated version of psycog because for some reason psycopg is newer than psycopg2 (https://pypi.org/project/psycopg/).

### DIFF
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -17,8 +17,6 @@ cli = [
     "ruff",
 ]
 db = [
-    "mysqlclient==2.2.5",
-    "psycopg==3.2.9",
 ]
 test = [
     "Faker==26.0.0",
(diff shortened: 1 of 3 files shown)


devmount:
Oh I only saw this now: Why is a timezone test failing now? This doesn't seem connected to this change, or is it?

MelissaAutumn:
> Oh I only saw this now: Why is a timezone test failing now? This doesn't seem connected to this change, or is it?

Nope, it's been failing for a bit. I'm kinda afraid to take a look at it, as it made me go crazy last time. :scream_cat: We should maybe look into that soon.

devmount:
> > Oh I only saw this now: Why is a timezone test failing now? This doesn't seem connected to this change, or is it?
> 
> Nope, it's been failing for a bit. I'm kinda afraid to take a look at it, as it made me go crazy last time. 🙀 We should maybe look into that soon.

Alright, I can take a look at that if you want. @davinotdavid and I were investigating timezone stuff for the bookers page anyway. And yes, it is hell 😅

Issue #1311, opened 2025-10-29
Availability toggle is out of sync


devmount:
**Describe the bug**
Availability toggle that shows "You're bookable" is sometimes out of sync. Perhaps we should double check if we are not holding references / caching on the frontend where we shouldn't.

**To Reproduce**
We need to investigate how to reproduce this. This might have to do with adding or changing calendars.

**Expected behavior**
Availability toggle keeping its state independent from other actions. Unless the user is informed about the availability being disabled for reasons (like e.g. the corresponding calendar was removed).

devmount:
@malini @davinotdavid I wasn't able to reproduce that directly (but I saw it with my own eyes in the Appointment meeting the other day). But, it may be fixed by fixing #1333

devmount:
Closing for now. Please reopen if this still occurs.

PR #1334, opened 2025-11-13
Disable deactivation of default calendar and sync availability

### LINKED ISSUE #1333: Default calendar toggle should be disabled
**Describe the bug**
The current UI allows the default calendar (the one used for the active schedule) to be deactivated. This shouldn't be possible according to designs.

**To Reproduce**

1. Go to Settings > Connected applications
2. Remove the tick on the default calendar
3. Save
4. Reload / revisit page to see that the changes were actually saved

**Expected behavior**
The default calendar is grayed out, so that it cannot be disabled, unless another calendar is set in the availability settings before.

**Actual behavior**
The default calendar can be disabled. This leads to an empty calendar select field in the availability settings. And even worse: When reloading the availability page, the next available calendar is displayed (but not actually used).

**Screenshots**
Design:

Image

Current implementation:

Image

- OS: Tuxedo OS (Ubuntu 24)
- Thunderbird Version 140.4.0esr (64-Bit)

**Additional context**
This might be related to #1311 as well.

devmount:
## Description of the Change

This change fixes a UI issue, where the default calendar can be disabled in the calendar settings. Also the default calendar changed in the settings is now properly synced to the availability page too (this was only the case the other way around until now).

## Benefits

Prevention of faulty availability/calnedar behaviour. Reduction of user confusion.

## Applicable Issues

Closes #1333 and hopefully #1311

### DIFF
--- a/frontend/src/views/SettingsView/components/ConnectedApplications.vue
+++ b/frontend/src/views/SettingsView/components/ConnectedApplications.vue
@@ -268,8 +268,8 @@ async function refreshData() {
             </div>
           </template>
         </drop-down>
-  
-        <br />
+        <span v-else></span>
+        <span></span>
       </template>
     </template>
 


frontend/src/views/SettingsView/components/ConnectedApplications.vue
```diff
@@ -57,6 +59,7 @@ const initialCalendars = computed(() => {
       type_id: externalConnection?.type_id,
       connection_name: externalConnection?.name,
       provider_name: connectionProvider,
+      is_default: currentState.value.defaultCalendarId === calendar.id,
```
devmount, line 62: This improves the markup a lot, because we now already have that information as property

frontend/src/views/SettingsView/components/ConnectedApplications.vue
```diff
@@ -129,7 +132,13 @@ function onCalendarChecked(event: HTMLInputElementEvent, calendarId: number) {
...
   // Only update local state, the actual schedule's calendar change
   // happens on save changes in SettingsView/index.vue
-  settingsStore.$patch({ currentState: { defaultCalendarId: calendarId }})
+  settingsStore.$patch({ currentState: { defaultCalendarId: calendarId }});
+  // Update the current availability too, otherwise it will be out
+  // of sync until the user refreshes the page
+  availabilityStore.$patch({
+    initialState: { calendar_id: calendarId },
+    currentState: { calendar_id: calendarId },
+  });
```
devmount, line 141: This fixes the out of sync availability config page, when changing calendars in the settings. We need to update both states, otherwise the availability page will find a mismatch and offer to save changes.

rwood-moz:
Ooops I added my review comment to the issue instead, sorry:

https://github.com/thunderbird/appointment/issues/1333#issuecomment-3528377784

devmount:
Thanks both of you,  @rwood-moz and @davinotdavid!! I fixed the row alignment now and the drop down now only shows up, if at least one active button exists:

image

rwood-moz:
Thanks @devmount! The menu now is not available for the default calendar (awesome) but now unfortunately there are some issues with the alignment of the other calendars that appear below the default one:

Screenshot 2025-11-13 at 4 47 26 PM

devmount:
> Thanks @devmount! The menu now is not available for the default calendar (awesome) but now unfortunately there are some issues with the alignment of the other calendars that appear below the default one:

Oh hahaha, I just love CSS grids 😂  Thanks so much for catching this, Rob! Should be fixed now

PR #1336, opened 2025-11-13
Fix custom availabilities when switching dates

### LINKED ISSUE #1284: Custom Availabilities error out when switching it ON / OFF and changing days
**Describe the bug**
As @rwood-moz [pointed out here](https://github.com/thunderbird/appointment/pull/1269#issuecomment-3362978053), custom availabilities stop working if you turn on a day that wasn't available before _after_ toggling the custom availabilities on and off.

**To Reproduce**

- Go to the Availability page
- Click the Set custom times for each day check box to turn it ON
- The custom days options appears below for each day, as expected
- Turn the Set custom times for each day check box OFF again
- The custom days options no longer appear below, as expected
- Now on the Available days and times section above, turn ON Sundays
- Then click the Set custom times for each day checkbox to turn it ON and notice the options don't appear
- Even if I save changes (after turning on Sundays) clicking the Set custom times for each day to turn it on still doesn't show the custom day options below

**Expected behavior**
The custom availabilities should still work as expected.

**Actual behavior**
The custom availability remains broken.

**Screenshots**

https://github.com/user-attachments/assets/abdcd9a5-e54b-4e5b-93dc-ce56bffe9b23

devmount:
## Description of the Change

This change is a fix for setting custom availability in combination with changing available days. I switched from a complex array generation to a simpler object notation for storing validation errors.

In addition, the current design changes were implemented. The availability fields were rearranged a bit:
image image

Finally, a bug was fixed were the custom availability validation was throwing false positives.

## Benefits

Changing day in and out of custom availabilities should be always in sync now, without errors.

## Applicable Issues

Closes #1284

### DIFF
--- a/backend/test/unit/test_schedule_availability.py
+++ b/backend/test/unit/test_schedule_availability.py
@@ -1,0 +1,110 @@
+from appointment.database.repo.schedule import all_availability_is_valid
+from appointment.database import schemas
+from datetime import date, time
+
+
+class TestScheduleAvailability:
+
+    def test_empty_availability_is_valid(self):
+        # Test empty availability is valid
+        schedule = schemas.ScheduleValidationIn(
+            name='test',
+            calendar_id=1,
+            slot_duration=30,
+            start_date=date.today(),
+            start_time=time(9, 0),
+            end_time=time(17, 0),
+        )
+        assert all_availability_is_valid(schedule)
+
+
+    def test_all_availability_is_valid(self):
+        # Test already sorted availabilities
+        schedule = schemas.ScheduleValidationIn(
+            name='test',
+            calendar_id=1,
+            slot_duration=30,
+            start_date=date.today(),
+            start_time=time(9, 0),
+            end_time=time(17, 0),
+            availabilities=[
+                schemas.AvailabilityValidationIn(
+                    schedule_id=1, day_of_week=1, start_time=time(9, 0), end_time=time(10, 0)
+                ),
+                schemas.AvailabilityValidationIn(
+                    schedule_id=1, day_of_week=1, start_time=time(10, 0), end_time=time(11, 0)
+                ),
+                schemas.AvailabilityValidationIn(
+                    schedule_id=1, day_of_week=2, start_time=time(9, 0), end_time=time(10, 0)
+                ),
+            ]
+        )
+        assert all_availability_is_valid(schedule)
+
+        # Test unordered availabilities
+        schedule.availabilities = [
+            schemas.AvailabilityValidationIn(
+                schedule_id=1, day_of_week=1, start_time=time(15, 0), end_time=time(16, 0)
+            ),
+            schemas.AvailabilityValidationIn(
+                schedule_id=1, day_of_week=1, start_time=time(10, 0), end_time=time(11, 0)
+            ),
+            schemas.AvailabilityValidationIn(
+                schedule_id=1, day_of_week=1, start_time=time(9, 0), end_time=time(10, 0)
+            ),
+            schemas.AvailabilityValidationIn(
+                schedule_id=1, day_of_week=1, start_time=time(17, 0), end_time=time(18, 0)
+            ),
+        ]
+        assert all_availability_is_valid(schedule)
+
+
+    def test_all_availability_is_invalid(self):
+        # Test overlapping end-start times
+        schedule = schemas.ScheduleValidationIn(
+            name='test',
+            calendar_id=1,
+            slot_duration=30,
+            start_date=date.today(),
+            start_time=time(9, 0),
+            end_time=time(17, 0),
+            availabilities=[
+                schemas.AvailabilityValidationIn(
+                    schedule_id=1, day_of_week=1, start_time=time(9, 0), end_time=time(11, 0)
+                ),
+                schemas.AvailabilityValidationIn(
+                    schedule_id=1, day_of_week=1, start_time=time(10, 0), end_time=time(12, 0)
+                ),
+            ]
+        )
+        assert not all_availability_is_valid(schedule)
+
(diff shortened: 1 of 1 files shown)


rwood-moz (changes requested):
Thanks @devmount

The change did fix the original issue (#1284) however I notice the following issue. I don't know if this has always been there or if this is from the current change.

- I have my availability set to M-F 9am-5pm
- Go into `Availability` and turn on the `Set custom times for each day` check box
- Now below that checkbox, in the rows (one for each day) click on `Wed` to turn off availability for Wednesdays
- There's an unexpected error and I can't save the change (see below)
- Note: If I instead don't turn on custom days, and instead just click above that and turn off Wednesdays on the general availability (under `Available days and times`), it does work and can be saved

Screenshot 2025-11-13 at 5 04 56 PM

devmount:
@rwood-moz Awesome catch, thanks so much for testing so thoroughly! I'll take a look at that

rwood-moz:
> @rwood-moz Awesome catch, thanks so much for testing so thoroughly! I'll take a look at that

Yw! I notice the M and T above it now says start 9PM and end 5PM so that makes sense it can't save that, but I don't know how the start for those two days changed from 9AM to 9PM...

devmount:
You're right, the error message is correct indeed. How did the start switch from am to pm 🤔 That's a strange one... I'll try to reproduce that

devmount:
@rwood-moz Hm, I did everything like you described, I even deleted all availabilities in the database to start from scratch, but I didn't get any error like this. Is this reproducible on your end?

rwood-moz:
> @rwood-moz Hm, I did everything like you described, I even deleted all availabilities in the database to start from scratch, but I didn't get any error like this. Is this reproducible on your end?

Hey @devmount I tried and couldn't reproduce that issue, very strange. Question though, when you set custom times for days, the time is not restricted by the general available days and times above is it? If I turn on `set custom times for each day`, and make the end time greater than the end time in the `available days and times` setting above, there is an error. Example below, I had set availability to M-F 9am-5pm, turned on custom times for each day, and tried to change M to end at 8PM and receive this error:

Screenshot 2025-11-14 at 9 44 33 AM

devmount:
Again an excellent find, Rob! But it shouldn't be restricted by the times above. Interestingly I can save the config from your screenshot without any problem on my end 🤔 

image

There's something going on here but I didn't find it yet. Thanks for your reports Rob, we'll find that one out, I'm sure 😇

davinotdavid:
@devmount @rwood-moz 

Sorry for the delay on replying to this and thanks for the great work so far!

So turns out that were some recent-ish design changes on the Custom Availability that give us some answers here:

https://app.zeplin.io/project/66dd0f2b2fe8f1b63cda36f1/screen/68e84a733b0d91c74e3dbe18

Availbility v02 - custom times per day ON

Seems like we shouldn't show / should ignore the top "start_time" / "end_time" if the checkbox for custom availability is set.

devmount:
Thanks for the design update! I will do the following two things, before you can test again:

1. Implement the design changes
2. Tweak the backend to only take the custom availability day limits into account if enabled

backend/src/appointment/database/repo/schedule.py
devmount, line 1: Oh wow, that was actually a nasty issue!! 🙈 We were sorting the availabilities beforehand, but then we calculated the predecessor and successor via index on the unsorted(!) list!

Didn't check git blame, but sorry if this was may bad. At least I found and fixed it 😂

davinotdavid: Works wonders, thanks for finding this! It is a great candidate for a unit / integration test though!

davinotdavid (commented):
Thank you so much for working on this! Indeed it was a surprisingly tricky one 🙏 

It works well during my manual testing here but it would be great if we could add some tests for `all_availability_is_valid` somewhere.

Oh also, if you could rebase the branch with `main` just to make sure that the CI is passing, it would be great! Thanks!

rwood-moz:
For me for some reason my local stack frontend fails to startup on this branch (localhost:8080  page shows a `sorry, this page could not be found` error.

 The backend says `healthy` but  then when I try to load the frontend URL it shows a `backend-1   | TRACE:    185.199.110.133:21051 - HTTP connection lost` error in the console.

When I switch back to the main branch (and restart the containers) my stack runs fine and I can sign in to the frontend etc.

backend/src/appointment/database/repo/schedule.py
devmount: True, I will add tests for this, great suggestions!

devmount:
Thanks all of you for the feedback! I will add some tests for this soon.

@rwood-moz The branch is already a bit older I think, I can rebase it and we could test if this still happens on your end.

rwood-moz:
> Thanks all of you for the feedback! I will add some tests for this soon.
> 
> @rwood-moz The branch is already a bit older I think, I can rebase it and we could test if this still happens on your end.

Thanks @devmount. Actually I just had the same error happen when I restarted my local Appointment stack on main, I think it's some weird intermittent issue on my machine/setup.

devmount:
@rwood-moz @davinotdavid I rebased the branch and added unit tests for schedule availability now, which cover the issue above (I tested with the implementation before the fix and they are failing correctly).

If you'd be so kind and review this agian, then we can finally merge it 🙏🏻

rwood-moz (approved):
I was able to get my local stack up and running again with this branch and tested it out, works good for me! Also I see your new unit test (and all tests) passed in CI on your branch. Nice, great fix @devmount 🏅

devmount:
My pleasure. Thanks for your great review feedback and also for reminding me again and again to write tests 😂 I always forget about them 😇

PR #1346, opened 2025-11-19
New FTUE for Appointment

### LINKED ISSUE #1345: Implement new designs for FTUE
We have a new experience / designs for the FTUE, so we should implement those.

Zeplin: https://app.zeplin.io/project/66dd0f2b2fe8f1b63cda36f1/screen/68e84c849eccc1aeb4415da1

davinotdavid:
## Description of the Change

- While using the same `/setup` route and localStorage logic as before, this route now goes to a brand new set of components that follow close resemblance to the Keycloak theming done in `thunderbird-accounts`.
- Added `VITE_TB_ACCOUNT_DASHBOARD_URL` env var so that the "Cancel" button redirects the user to `accounts.tb.pro/dashboard`.
- Removed `VITE_AUTH_EDIT_PROFILE` env var and linked the "Edit" button in the Profile page to be `VITE_TB_ACCOUNT_DASHBOARD_URL` instead.
- Removed all previous components from the old FTUE.
- On `post-login` route, removed the "app skeleton" (header / footer, etc).

Zeplin:
https://app.zeplin.io/project/66dd0f2b2fe8f1b63cda36f1/screen/68e84c8f8b0327368e72b1fd

## Video Walk-through

https://github.com/user-attachments/assets/e41a40ba-e902-4315-a6dc-caa49e1e3ad1

## Benefits

- A better / smoother first time user experience to setup a new user for success!

- ~~When Design is ready, we need to update the left side imagery as for now I am using the same one as `thunderbird-accounts`' Keycloak Theme.~~
- There's going to be another option for calendar connection which will be an auto-setup with TB Pro / Stalwart's Calendar. This requires a token introspection https://github.com/thunderbird/appointment/issues/1352 and Design is working on how to best display this option. There is a commented out button on where the final implementation should be but that's not required for this ticket as of right now.

## Applicable Issues

Resolves #1345

### DIFF
--- a/frontend/src/views/FTUEView/steps/SetupCompleteStep.vue
+++ b/frontend/src/views/FTUEView/steps/SetupCompleteStep.vue
@@ -1,7 +1,6 @@
 <script setup lang="ts">
 import { onMounted } from 'vue';
 import { useI18n } from 'vue-i18n';
-import { useRouter } from 'vue-router';
 import { PrimaryButton } from '@thunderbirdops/services-ui';
 import { useUserStore } from '@/stores/user-store';
 
@@ -9,11 +8,10 @@ import StepTitle from '../components/StepTitle.vue';
 
 const { t } = useI18n();
 
-const router = useRouter();
 const userStore = useUserStore();
 
 const onContinueButtonClick = async () => {
-  router.push('/dashboard');
+  window.location.href = '/dashboard';
 };
 
 onMounted(async () => {
@@ -23,8 +21,9 @@ onMounted(async () => {
   // Clear the FTUE flow
   window.localStorage?.removeItem('tba/ftue');
 
-  // Redirect to dashboard after a delay
-  setTimeout(() => router.push('/dashboard'), 5000);
+  // Hard reload to dashboard after a delay so that
+  // we initialize all the stores and data properly
+  setTimeout(() => window.location.href = '/dashboard', 5000);
 });
 </script>
 


frontend/src/locales/de.json
```diff
@@ -46,38 +46,51 @@
...
-    "finishAltText": "Ein Benutzer-Icon vor zwei Kalendern.",
-    "finishScreenText": "Bevor Du gehst: Kopiere den Link zu Deinem Terminplan, um Termin-Buchungen zu erhalten",
+    "createYourProfile": "Dein Profil erstellen",
     "fullName": "Vollständiger Name",
+    "urlUsername": "URL Benutzername",
+    "createAccount": "Konto erstellen",
+    "connectYourCalendar": "Kalender verbinden",
+    "connectYourCalendarInfo": "Thunderbird Appointment benötigt Zugriff auf einen Kalender, um deine Verfügbarkeit zu prüfen. Du kannst jederzeit weitere Kalender aus den Einstellungen verbinden.",
+    "connectCalendarCalDav": "Mit CalDAV verbinden",
+    "connectCalendarCalDavInfo": "Synchronisiere Thunderbird Appointment Kalender mit deinem Kalender mithilfe eines CalDAV Links.",
```
```suggestion
    "connectCalendarCalDavInfo": "Synchronisiere Thunderbird Appointment mit deinem Kalender über einen CalDAV Link.",
```

frontend/src/locales/de.json
```diff
@@ -46,38 +46,51 @@
...
+    "connectCalendarGoogle": "Mit Google Kalender verbinden",
+    "connectCalendarGoogleInfo": "Synchronisiere Thunderbird Appointment mit deinem Google Kalender. Du kannst auswählen, welche Kalender eingeschlossen werden sollen.",
+    "connectCalendarTBPro": "Mit Thunderbird Pro verbinden",
+    "connectCalendarTBProInfo": "Synchronisiere Thunderbird Appointment mit deinem Thunderbird Pro Kalender. Du kannst auswählen, welche Kalender eingeschlossen werden sollen.",
+    "connectWithCalDav": "Mit CalDAV verbinden",
+    "calendarUrl": "Kalender URL",
+    "calendarUrlPlaceholder": "Full URL eingeben",
```
```suggestion
    "calendarUrlPlaceholder": "Vollständige URL eingeben",
```

frontend/src/locales/de.json
davinotdavid: Awesome thanks! Applied!

frontend/src/locales/de.json
davinotdavid: Thank you 🙏

davinotdavid:
@devmount Thank you so much for the review, I really appreciate it!

> 1. Schedule times might (!) not be taking local timezone into account

Ah indeed, there was a missing conversion during the schedule creation (the `timeToFrontendTime` / `timeToBackendTime`). It should be fixed in [this commit](https://github.com/thunderbird/appointment/pull/1346/commits/c38bbbfc51f81391124f9ed26b17e1bcf7c83c9e).

> 2. Custom availability error message

Ah yes, it shouldn't be possible to do custom availabilities in the FTUE so that is most likely because of attempting to do a FTUE for an existing user with a custom availability set (during that process we are attempting to create / update the schedule and it relies on the user not having a previously set custom availability).

> 3. CalDAV UX

Great catch! I've added disabled buttons and a "Connecting..." message to the button after pressing so hopefully the feedback should be better now. Thanks!

davinotdavid:
@rwood-moz Thanks for the review! Great points added!

> On the Create your profile screen, for full name and URL username you can just enter a single blank space in each and that is accepted.

Oh that's bad, we shouldn't be able to accept a single blank space and I am amazed that it still worked until the booking link! I've now added some extra validation using `.trim()` to prevent that from happening. Thanks!

> It would be nice if on the Create your profile screen it could default to your local timezone instead of UTC

Hmm interesting, here it does default to the local timezone. I tried using the same logic as the previous appointment FTUE for guessing the user's local timezone if none are provided. Were you using a previously existing user that was set to UTC before? We store user information in local storage so I wonder if UTC ended up there somehow. When you re-test it, could you check what your local storage has?

image


> This calendar doesn't require sign-in credentials check box 

Great catch! When I was checking how to solve this, there are quite a few changes required in our backend to make it work so I went to check the current FTUE for Appointment and turns out that we also don't have this support yet. I've punt it into another ticket just so we can isolate the changes over there for closer review: https://github.com/thunderbird/appointment/issues/1363

> On the Create booking page the calendar for new appointments field has no default value and you are forced to click the pull down list and select one; maybe it would be better to have the default calendar (that was just connected on the previous screen) already selected

Agreed, changed here! Good catch, thanks!

> An inconsistency, the page headings only capitalize the first word (which makes sense to me too) however the Video Meeting Link page has capitalized each word

Awesome attention to detail! Changed here, thanks!

rwood-moz:
Thanks for the updates @davinotdavid, ran your latest branch locally.

> Oh that's bad, we shouldn't be able to accept a single blank space and I am amazed that it still worked until the booking link! I've now added some extra validation using `.trim()` to prevent that from happening. Thanks!

Yep looks good, now an error appears saying the fields cannot be empty, nice.

> > It would be nice if on the Create your profile screen it could default to your local timezone instead of UTC
> 
> Hmm interesting, here it does default to the local timezone. I tried using the same logic as the previous appointment FTUE for guessing the user's local timezone if none are provided. Were you using a previously existing user that was set to UTC before? We store user information in local storage so I wonder if UTC ended up there somehow. When you re-test it, could you check what your local storage has?

No I've always set the timezone to `America/Toronto`. Shows UTC in the local storage too. Happens in incognito or non-private tabs.
Screenshot 2025-12-01 at 3 05 05 PM

> Great catch! When I was checking how to solve this, there are quite a few changes required in our backend to make it work so I went to check the current FTUE for Appointment and turns out that we also don't have this support yet. I've punt it into another ticket just so we can isolate the changes over there for closer review: #1363

Ok awesome thank you.

> > On the Create booking page the calendar for new appointments field has no default value and you are forced to click the pull down list and select one; maybe it would be better to have the default calendar (that was just connected on the previous screen) already selected
> 
> Agreed, changed here! Good catch, thanks!

Hmm for me it still is empty even though I have connected my CalDAV thundermail calendar:

Screenshot 2025-12-01 at 3 09 00 PM

> > An inconsistency, the page headings only capitalize the first word (which makes sense to me too) however the Video Meeting Link page has capitalized each word
> 
> Awesome attention to detail! Changed here, thanks!
Screenshot 2025-12-01 at 3 09 15 PM

Thank you, yep looks great!

kewisch:
> > Oh that's bad, we shouldn't be able to accept a single blank space and I am amazed that it still worked until the booking link! I've now added some extra validation using `.trim()` to prevent that from happening. Thanks!
> 
> Yep looks good, now an error appears saying the fields cannot be empty, nice.

Are we validating this server side as well?

davinotdavid:
@rwood-moz Thanks again for re-checking! Really appreciate it 🙏 

> No I've always set the timezone to America/Toronto. Shows UTC in the local storage too. Happens in incognito or non-private tabs.

Ah, mystery solved! For OIDC / Keycloak, it seems that we are sourcing the timezone from the `zoneInfo` attribute in the Keycloak User record. [Mine works](https://auth-stage.tb.pro/admin/master/console/#/tbpro-dev/users/7bbb3ee0-cdf2-4420-a93f-6d4cc165e61c/settings) because it is set as `america/regina` and [yours default to UTC since it is unset](https://auth-stage.tb.pro/admin/master/console/#/tbpro-dev/users/e582b514-21d2-4555-b0dc-2a522d7a4b1f/settings).

In the `password` type of auth, there is an explicit line defaulting it to UTC [here](https://github.com/thunderbird/appointment/blob/main/backend/src/appointment/routes/auth.py#L544). So since this is expected / pre-existing behaviour, I will leave it as is. Good catch though!

> Hmm for me it still is empty even though I have connected my CalDAV thundermail calendar

I could reproduce this after a few attempts, it was a race condition on fetching the calendar data and setting the default values. I've added a `watch` there to solve for this and now it is reliable / consistent. Would you give it a try again?

Please delete your local storage data and do a `docker compose down -v` and `docker compose up -d --build` to clean everything before retrying as if you had old data somewhere, the code will consider that you've already had a pre-selection!

davinotdavid:
@kewisch we did not have guards in place for that yet no :( I've added the guards on the schema level + unit tests [in this commit](https://github.com/thunderbird/appointment/pull/1346/commits/3b4680fd55a5a921a18ab4355f194b92a5d5fad1).

Also, it seems that the previously existing code for surfacing errors from the backend was also a bit faulty so fixed that too. Thanks for surfacing this!

rwood-moz:
Awesome thank you @davinotdavid 

> I could reproduce this after a few attempts, it was a race condition on fetching the calendar data and setting the default values. I've added a `watch` there to solve for this and now it is reliable / consistent. Would you give it a try again?
> 
> Please delete your local storage data and do a `docker compose down -v` and `docker compose up -d --build` to clean everything before retrying as if you had old data somewhere, the code will consider that you've already had a pre-selection!

Yep looks great now, my caldav thundermail calendar shows up now by default as the selected one. **Nice fix!**

Now I just found one more issue (sorry) maybe you could see if you can reproduce it:

- Note for VITE_AUTH_SCHEME I am using 'password' not OIDC in my frontend .env
- docker compose down -v
- docker compose up --build -d
- Open a new Firefox incognito window (not sure that matters though) and browse to local Appointment (localhost:8080)
- Walk through the FTUE successfully, connect my thundermail stage calendar, and use the default availability options
- Then at the end of the FTUE get redirected to the dashboard as expected
- Now click on the `Availability` tab
- Notice there is a message saying I have unsaved changes, and the settings are all empty:
Screenshot 2025-12-02 at 1 52 34 PM
Screenshot 2025-12-02 at 1 52 50 PM

- Then if I click `revert changes` the settings (that I chose during the FTUE) now appear as they should (except for the `booking to` option:

Screenshot 2025-12-02 at 1 53 06 PM
Screenshot 2025-12-02 at 1 53 12 PM

davinotdavid:
@rwood-moz fantastic catch! I was using `vue-router` to navigate to the Dashboard but that means that I would need to re-initialize / reset all the stores and fetch the expected data to initialize de app.

Instead, I am not "hard redirecting" which already refreshes / initializes all the data as expected! Would you mind checking it again? Sorry to ask a lot of times 😅 🙇

rwood-moz (approved):
> Instead, I am not "hard redirecting" which already refreshes / initializes all the data as expected! Would you mind checking it again? Sorry to ask a lot of times 😅 🙇

@davinotdavid No worries at all! Tried your latest updates and yep that issue is fixed now. Awesome work, you're super fast at debugging! 🥇

davinotdavid:
@rwood-moz @devmount thank you so much for the reviews! Really appreciate you 🙇 🙏

Issue #1340, opened 2025-11-18
Public availability date time mistmatch test failure


davinotdavid:
**Describe the bug**
We started having a test failure related to Availability / Schedule recently:

```
FAILED test/integration/test_schedule.py::TestSchedule::test_public_availability - AssertionError: assert '2024-03-04T08:00:00-08:00' == '2024-03-04T09:00:00-08:00'
```

Which I believe is due to a daylight savings time mismatch. However, this should be further investigated to check if this is actually a bug on the application code or a test fixture error.

**To Reproduce**
This can be seen in all current PRs since it breaks CI but also one can run the tests locally.

**Expected behavior**
Tests should pass / code should be fixed if there's a bug

PR #1357, opened 2025-11-26
Fix schedule public availability test

davinotdavid:
## Description of the Change

- The `make_schedule` sets `time_updated` to `func.now()` which creates the time inconsistency if it is not under the `freeze_time`.

In the original test:
- `start_time` was set to 16:00 UTC
- `make_schedule` sets `time_updated` to `func.now()`
- We calculate the local start time by combining `time_updated` with `start_time` and the converting the timezone to `America/Vancouver`
- If the test is ran in PDT
  - 16:00 UTC is 09:00 PDT
  - `start_time_local` becomes 09:00, all good, tests pass, hooray!
- If the test is ran in PST
  - 16:00 UTC is 08:00 PST
  - `start_time_local` becomes 08:00, no good, test fail, boo!

## Benefits

- Backend tests should now pass and not flake depending on the time of the year it is being ran.

## Applicable Issues

Solves https://github.com/thunderbird/appointment/issues/1340

### DIFF
--- a/backend/test/integration/test_schedule.py
+++ b/backend/test/integration/test_schedule.py
@@ -533,28 +533,30 @@ class TestSchedule:
         monkeypatch.setattr(CalDavConnector, 'get_busy_time', MockCaldavConnector.get_busy_time)
 
         start_date = date(2024, 3, 1)
-        start_time = time(16, tzinfo=UTC)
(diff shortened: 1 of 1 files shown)


devmount:
Ohh this is such a good find! Thanks for fixing this 🙌🏻

Issue #1250, opened 2025-09-15
Add caldav connection dialog issues on android mobile browser


rwood-moz:
**Describe the bug**
While developing E2E tests on android mobile I noticed a couple of issues with the add caldav connection dialog that only happen on the android mobile browser:

- There is no way to close the caldav connection dialog once you open it (except by manually refreshing entire page)
- The `CalDAV Connection` header is missing from the dialog

**To Reproduce**
- Sign into Appointment on the android device browser
- Once on the dashboard, go into `Settings` and scroll down to `Connected Applications`
- Click on the `Add CalDav` button

**Expected behavior**
The `CalDAV` connection dialog appears, and the `Connect CalDAV` header appears, and you are able to close the dialog easily.

**Actual behavior**
The header is missing and there is no close button; the only way to close the add caldav connection dialog is to manually refresh the browser page (even selecting the `hamburger menu => settings` on the mobile browser only returns to the same caldav connection dialog.

**Screenshots**
Screenshot from android Google Pixel 10 chrome:

<img width="235" height="470" alt="Image" src="https://github.com/user-attachments/assets/fe9d991c-5f5a-4b2c-af6b-a903d04715c9" />

Screenshot from Desktop Firefox:

<img width="542" height="409" alt="Image" src="https://github.com/user-attachments/assets/53386117-f136-461b-b4bc-5a9dcc7b24a7" />

**System (please complete the following information):**
Chrome browser on Google Pixel 10 device.

**Additional context**
These issues aren't present in Appointment on a desktop browser.

davinotdavid:
@rwood-moz this should be now be fixed in Stage and will make its way into Prod in the next deployment!

rwood-moz:
> [@rwood-moz](https://github.com/rwood-moz) this should be now be fixed in Stage and will make its way into Prod in the next deployment!

I tried it out on stage using the playwright google pixel viewport and it looks good! On the caldav connection page the header is there now and I can chose to exit and go back to the dashboard via the 'Dashboard' hamburger menu item. Nice!

davinotdavid:
@rwood-moz Woohoo! Thanks for trying it out (and for filing the issue) 🚀

PR #1367, opened 2025-12-03
Improve responsiveness of GenericModal component

davinotdavid:
## Description of the Change

- The `GenericModal` component was largely hardcoded design-wise with various absolute positioned elements. This PR changes that so that it is a bit more flexible before we revamp the UI for this component (which I suspect will happen sooner than later).
- We were rendering a close button during the login modal when using the `password` auth method, which had no event handler. This PR hides it.
- The main problem to solve here was that the modal was being rendered _behind_ the top navigation on mobile, which covered the close button. This has been changed to 

## Screenshots

Desktop / Mobile (Login usage)

image

image

Desktop / Mobile (CalDAV connection modal on Settings > External Connections > Add CalDAV)

image

image

## Benefits

- Better responsiveness for mobile device users

- I assume that this `GenericModal` will probably be replaced by a new design soon-ish as it is not matching that much with the new design language of the rest of the app, so this is more of a mitigation for now.

## Applicable Issues

Solves https://github.com/thunderbird/appointment/issues/1250

### DIFF
--- a/frontend/src/components/GenericModal.vue
+++ b/frontend/src/components/GenericModal.vue
@@ -230,24 +230,37 @@ onUnmounted(() => {
 
 @media (--md) {
   .modal {
-    width: 50rem; /* 800px */
+    width: 48rem; /* 768px */
     padding: 2rem 2rem 0;
-    overflow: visible;
+    overflow: hidden;
     height: min-content;
+    max-height: 90vh;
+  }
+
+  .modal-header {
+    flex-shrink: 0;
+  }
+
+  .modal-body {
+    overflow-y: auto;
+    min-height: 0;
   }
 
   .modal-actions {
     justify-content: center;
     margin-block: 2rem;
+    flex-shrink: 0;
   }
 
   .divider {
     width: 50rem;
     margin: 0;
+    flex-shrink: 0;
   }
 
   .footer {
     padding-bottom: 0;
+    flex-shrink: 0;
   }
 }
 


frontend/src/components/GenericModal.vue
```diff
@@ -232,39 +229,24 @@ onUnmounted(() => {
 }
 
 @media (--md) {
-  .modal-header {
-    margin-bottom: 0;
-  }
-
   .modal {
     width: 50rem; /* 800px */
```
devmount: Can we align that value with the current breaking point? Otherwise we have an intermediate state, where the sides are cut from the modal:

Image


```suggestion
    width: 48rem; /* 768px */
```

davinotdavid:
@devmount thank you so much for the review! Great catches all around, sorry to miss those 😅 

Alright so I've applied some changes to hopefully address your points but please let me know if you have different ideas / suggestions! Mobile is unchanged as it seemed to work well but desktop now has a max-height to leave some space at the top / bottom for visual flair. Also, the modal-body is scrollable but the header is not so that it still gives the user the reference of which modal they are interacting with.

CalDAV case
https://github.com/user-attachments/assets/4a0887cc-53f0-48af-9f0a-8bd99716b880


Login case
https://github.com/user-attachments/assets/6da04fec-080c-4d25-9f52-3df1f0eaeccc

devmount (approved):
Don't worry, thanks for fixing that real fast! Looks great 🎉

PR #1370, opened 2025-12-04
Improve nightly E2E test suite authentication issues in BrowserStack

### LINKED ISSUE #1369: E2E Test Authentication Issues
The Appointment E2E tests are having auth issues when running the nightly test suite on production on BrowserStack The issue is that the setup auth step is performed for all browsers at the start of the entire suite; but the tests take a fair amount of time to run as BrowserStack is quite slow, so during the suite the auth session times out and so we are no longer signed into Appointment and the tests fail.

The solution is to restructure the E2E test jobs so that each browser that we test on is run in it's own BrowserStack session, this has advantages:

- Auth is done only for the single browser in the test session
- Shorter session jobs in BrowserStack, easier to cancel out if we need to
- Easier to see test results inside BrowserStack
- Easy to see if tests failed on a particular browser inside the github action nightly tests runner steps

rwood-moz:
Fixes #1369: Restructure the nightly E2E test suite into multiple BrowserStack sessions to fix authentication issues. Also adds a check at the start of each test in case the Appointment session has timed out, then it will log in again and save the auth state before the test starts. This is still not foolproof in that the Appointment sign in session might expire during a test and then the test would fail, however in that case the very next test would sign in again.

BrowserStack links here for [desktop](https://automate.browserstack.com/dashboard/v2/builds/e57dc6914e642908215e1d1a55d57b138d181a7c) tests and [mobile](https://automate.browserstack.com/dashboard/v2/builds/a85038adb83f01a39c0f944c863ddcb3d5898138) tests on stage.


test/e2e/pages/tb-accts-page.ts
```diff
@@ -14,7 +14,7 @@ export class TBAcctsPage {
 
   constructor(page: Page) {
     this.page = page;
-    this.signInHeaderText = this.page.getByText('Enter your password');
+    this.signInHeaderText = this.page.getByText('Sign in to your account');
```
MelissaAutumn, line 17: I wonder if we can just hook into the localization file for appointment and then call it by the l10n key. 

...sorry just thinking ahead, no action for this PR 😄

rwood-moz:
> Looks fine to me, thank you!
> 
> I wonder if we can eventually figure out a wait to remove all these waits, seems like a lot of unfortunate busy work 😅.

Yes that would be ideal, but unfortunately the timing is different on local vs stage vs prod vs mobile (especially slower in BrowserStack), so it's pretty tricky. Thanks for the review!

PR #1377, opened 2025-12-12
First set of E2E tests on iOS

### LINKED ISSUE #1261: Add the first set of E2E tests to run on iOS
Port over the first set of E2E tests (settings tests) to run on an iOS (iPhone) browser, and add the tests to the existing nightly mobile E2E tests BrowserStack job.

rwood-moz:
Enable running the first set of E2E tests (settings tests) on iOS in BrowserStack, as part of the nightly mobile test job. Fixes #1261.

### DIFF
--- a/test/e2e/package.json
+++ b/test/e2e/package.json
@@ -7,7 +7,7 @@
     "e2e-test-headed": "npx playwright test --grep e2e-suite --project=firefox --headed",
     "e2e-test-browserstack-firefox": "npx browserstack-node-sdk playwright test --grep e2e-suite --project=Firefox-OSX --max-failures=2 --browserstack.buildName 'Appointment E2E Tests Firefox Desktop' --browserstack.config 'browserstack-desktop.yml'",
     "e2e-tests-mobile-browserstack-android-chrome": "npx browserstack-node-sdk playwright test --grep e2e-mobile-suite --project=android-chrome --max-failures=2 --browserstack.buildName 'Appointment E2E Tests Android Chrome' --browserstack.config 'browserstack-mobile.yml'",
-    "e2e-tests-mobile-browserstack-ios-safari": "npx browserstack-node-sdk playwright test --grep e2e-mobile-suite --project=ios-safari --max-failures=2 --browserstack.buildName 'Appointment E2E Tests iOS Chrome' --browserstack.config 'browserstack-mobile.yml'",
+    "e2e-tests-mobile-browserstack-ios-safari": "npx browserstack-node-sdk playwright test --grep e2e-mobile-suite --project=ios-safari --max-failures=2 --browserstack.buildName 'Appointment E2E Tests iOS Safari' --browserstack.config 'browserstack-mobile.yml'",
     "prod-sanity-test": "npx playwright test --grep prod-sanity --project=firefox",
     "prod-sanity-test-headed": "npx playwright test --grep prod-sanity --project=firefox --headed",
     "prod-sanity-test-browserstack-firefox": "npx browserstack-node-sdk playwright test --grep prod-sanity --project=Firefox-OSX --max-failures=2 --browserstack.buildName 'Appointment Production Sanity Test Firefox Desktop' --browserstack.config 'browserstack-desktop.yml'",
--- a/test/e2e/pages/settings-page.ts
+++ b/test/e2e/pages/settings-page.ts
@@ -30,6 +30,7 @@ export class SettingsPage {
   readonly displayNameInput: Locator;
   readonly bookingPageURLInput: Locator;
   readonly copyLinkBtn: Locator;
+  readonly copyLinkToolTipText: Locator;
   readonly cancelServiceBtn: Locator;
   readonly cancelServiceConfirmCancelBtn: Locator;
   readonly bookingPageSettingsBtn: Locator;
@@ -68,6 +69,7 @@ export class SettingsPage {
     this.displayNameInput = this.page.locator('#booking-page-display-name');
     this.bookingPageURLInput = this.page.locator('#booking-page-url');
     this.copyLinkBtn = this.page.locator('#copy-booking-page-url-button');
+    this.copyLinkToolTipText = this.page.locator('#tooltip-body');
     this.cancelServiceBtn = this.page.getByRole('button', { name: 'Cancel Service' });
     this.cancelServiceConfirmCancelBtn = this.page.getByRole('button', { name: 'Cancel', exact: true });
     this.bookingPageSettingsBtn = this.page.getByRole('button', { name: 'Booking Page Settings' });
--- a/test/e2e/tests/desktop/settings-account.spec.ts
+++ b/test/e2e/tests/desktop/settings-account.spec.ts
@@ -12,7 +12,6 @@ import {
   APPT_MY_SHARE_LINK,
   TIMEOUT_1_SECOND,
   TIMEOUT_30_SECONDS,
-  TIMEOUT_3_SECONDS,
  } from '../../const/constants';
 
 let settingsPage: SettingsPage;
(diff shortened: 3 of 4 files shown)


rwood-moz:
Update: With these updates all of the mobile tests pass on iOS when each one is run one at a time in it's own BrowserStack session; but when they are run all together as a group I am seeing `BROWSERSTACK_IDLE_TIMEOUT` session errors for some of the tests. This is happening on iOS only; the same test set on android works fine as a group. No idea why... but will continue looking into it.

BrowserStack link running the mobile tests [on android chrome](https://automate.browserstack.com/dashboard/v2/builds/ea526ae44843880cc89697a8eb040b7d3882cbb3) (all pass).
BrowserStack link running the same mobile tests [on iOS safari](https://automate.browserstack.com/dashboard/v2/builds/1e4a391b82f3f37d4e1878302b4273b76c382da6) (BrowserStack idle errors).

rwood-moz:
Increasing the BrowserStack `idleTimeout` and also closing the browser (instead of just the page) after all tests **_seems_** to be fixing the issue, all tests [passing here on iPhone 17](https://automate.browserstack.com/dashboard/v2/builds/8dd4b5e06f538f70490f2075e41813c29d826e94) 🥳

rwood-moz:
Last issue I'm seeing on iOS: Intermittently failing to read from local storage; probably just a timing issue as it doesn't happen that often, and just on iOS not android:

```
  1) [ios-safari] › tests/mobile/settings-connected-apps.spec.ts:38:7 › settings - connected applications on mobile browser › verify connected applications settings on mobile browser @e2e-mobile-suite @prod-mobile-nightly 

    SyntaxError: "undefined" is not valid JSON

       at ../utils/utils.ts:77

      75 | export const setDefaultUserSettingsLocalStore = async (page: Page) => {
      76 |   console.log('setting user settings to default values in browser local store')
    > 77 |   var localUserStoreData = JSON.parse(await page.evaluate("localStorage.getItem('tba/user')"));
         |                                 ^
      78 |
      79 |   console.log(`original user settings from local browser store: ${JSON.stringify(localUserStoreData['settings'])}`);
      80 |
        at setDefaultUserSettingsLocalStore (/Users/rwood/dev/appointment/test/e2e/utils/utils.ts:77:33)
        at mobileSignInAndSetup (/Users/rwood/dev/appointment/test/e2e/utils/utils.ts:113:3)
        at /Users/rwood/dev/appointment/test/e2e/tests/mobile/settings-connected-apps.spec.ts:26:5
```

rwood-moz:
This PR is looking good now! Mobile tests passing on [iOS safari ](https://automate.browserstack.com/dashboard/v2/builds/f34b964cf7780354a79fb6debccee02a326225a7)and still passing fine on [android chrome](https://automate.browserstack.com/dashboard/v2/builds/d61c5ed0273262520f5185b5fa57d2ffb358647a).

test/e2e/pages/settings-page.ts
```diff
@@ -5,15 +5,16 @@ import {
   APPT_HTML_DARK_MODE_CLASS,
   TIMEOUT_1_SECOND,
   TIMEOUT_2_SECONDS,
-  TIMEOUT_3_SECONDS,
   TIMEOUT_5_SECONDS,
   TIMEOUT_30_SECONDS,
   APPT_LANGUAGE_SETTING_EN,
   } from '../const/constants';
+import test from 'node:test';
```
davinotdavid: Nit: this is unused at the moment apparently, so we can remove it!

test/e2e/pages/settings-page.ts
davinotdavid: Unrelated to this PR but I was checking the file itself and I was wondering if this is a valid call? At least it doesn't seem to be a Promise to be awaited.

test/e2e/utils/utils.ts
```diff
@@ -4,22 +4,22 @@ import { expect, type Page } from '@playwright/test';
...
+  APPT_BROWSER_STORE_LANGUAGE_EN,
+  APPT_BROWSER_STORE_THEME_LIGHT,
+  APPT_BROWSER_STORE_12HR_TIME,
+  APPT_BROWSER_STORE_START_WEEK_SUN,
+  TIMEOUT_1_SECOND,
+  TIMEOUT_2_SECONDS,
+  TIMEOUT_5_SECONDS,
+  TIMEOUT_60_SECONDS,
 } from "../const/constants";
+import { createDecipheriv } from "crypto";
```
davinotdavid: Nit: This also seems to be unused

test/e2e/browserstack-mobile-nightly.yml
```diff
@@ -68,6 +74,7 @@ debug: false # <boolean> # Set to true if you need screenshots for every seleniu
 networkLogs: false # <boolean> Set to true to enable HAR logs capturing; off as may contain sensitive info like login API requests
 consoleLogs: info # <string> Remote browser's console debug levels to be printed (`disable`, `errors`, `warnings`, `info`, or `verbose`)
 framework: playwright
+idleTimeout: 500 # seeing if this helps rid BROWSERSTACK_IDLE_TIMEOUTs on ios
```
davinotdavid, line 77: I am curious! Was this the change that solved the idle timeout or was this block? Or both? 

```ts
test.afterAll(async ({ browser }) => {
    // close the browser when we're done, BrowserStack needs this or may report BROWSERSTACK_IDLE_TIMEOUT
    await browser.close();
  });
```

test/e2e/browserstack-mobile.yml
```diff
@@ -68,6 +74,7 @@ debug: false # <boolean> # Set to true if you need screenshots for every seleniu
 networkLogs: false # <boolean> Set to true to enable HAR logs capturing; off as may contain sensitive info like login API requests
 consoleLogs: info # <string> Remote browser's console debug levels to be printed (`disable`, `errors`, `warnings`, `info`, or `verbose`)
 framework: playwright
+idleTimeout: 500 # seeing if this helps rid BROWSERSTACK_IDLE_TIMEOUTs on ios
```
davinotdavid, line 77: Same question here!

test/e2e/package.json
```diff
@@ -5,14 +5,17 @@
...
+    "e2e-tests-mobile-browserstack-ios-safari": "npx browserstack-node-sdk playwright test --grep e2e-mobile-suite --project=ios-safari --max-failures=2 --browserstack.buildName 'Appointment E2E Tests iOS Chrome' --browserstack.config 'browserstack-mobile.yml'",
     "prod-sanity-test": "npx playwright test --grep prod-sanity --project=firefox",
     "prod-sanity-test-headed": "npx playwright test --grep prod-sanity --project=firefox --headed",
-    "prod-sanity-test-browserstack-firefox": "npx browserstack-node-sdk playwright test --grep prod-sanity --project=Firefox-OSX --browserstack.buildName 'Appointment Production Sanity Test Firefox Desktop' --browserstack.config 'browserstack-desktop.yml'",
+    "prod-sanity-test-browserstack-firefox": "npx browserstack-node-sdk playwright test --grep prod-sanity --project=Firefox-OSX --max-failures=2 --browserstack.buildName 'Appointment Production Sanity Test Firefox Desktop' --browserstack.config 'browserstack-desktop.yml'",
```
davinotdavid, line 13: Why are we adding `--max-failures=2` for both prod and non-prod for the browserstack-firefox tests? Are there some tests that are always failing for now?

rwood-moz: When the E2E tests run after deploying to stage or prod I added this so that if 2 of the tests fail then the suite just stops; that way we don't have to wait for all the other tests to finish running when the test job has already failed. This is just to speed up waiting for the tests to run after deployments.

davinotdavid: Ah I see! Makes sense, thanks!

test/e2e/pages/settings-page.ts
rwood-moz: hahaha great catch, will update to what it was meant to be:

`await this.startOfWeekMondayBtn.scrollIntoViewIfNeeded();`

rwood-moz: Actually `scrolIntoViewIfNeeded` is not supported yet by Playwright for IOS:

```
locator.scrollIntoViewIfNeeded: Error: Unsupported Playwright command on iOS: "scrollIntoViewIfNeeded".
This command is not yet available on iOS; we're working to expand supported commands. 
```

So just removing that line, it's not really necessary - just is good when debugging/test failures get screenshots.

rwood-moz: Actually `scrollIntoView(locator)` is supported, but `locator.scrollIntoViewIfNeeded` is not, so will just use `scrollIntoView`.

`await this.scrollIntoView(this.startOfWeekMondayBtn);`

test/e2e/browserstack-mobile-nightly.yml
rwood-moz: Good question! I did some more testing and it looks like the `idleTimeout: 500` setting is enough to fix the BROWSERSTACK_IDLE_TIMEOUT errors; however I am going to leave the browser close in because it is good practice anyway to do this especially for mobile.

rwood-moz:
Looks like there was a change that is now on stage and the `Settings => Booking page URL' field is no longer editable (which makes sense). The account settings E2E test was using this field to paste in the copied URL and verify it is correct (after the test clicks the copy share link button) since the test cannot access the clipboard, so the test is failing now. I will make an update in this PR to update the test (maybe use the display name field instead to paste in the copied share link to verify it).

davinotdavid:
@rwood-moz oh sorry about that and good catch! Hopefully there's a workaround to check the correctness of the booking link 🙏

rwood-moz:
> @rwood-moz oh sorry about that and good catch! Hopefully there's a workaround to check the correctness of the booking link 🙏

@davinotdavid No worries, it makes sense for that field to be readonly, I'll see if I can make a workaround.

rwood-moz:
> @davinotdavid No worries, it makes sense for that field to be readonly, I'll see if I can make a workaround.

I wasn't able to make a workaround that works on android, chrome and desktop (BrowserStack and Playwright mobile support isn't the most consistent unfortunately). Just updating the test to ensure the copy link button is enabled and can be clicked. It's better to have the test less complex anyway as more complexity means more chance for intermittent failures and future maintenance.

rwood-moz:
Latest updates running on [Android Chrome](https://automate.browserstack.com/dashboard/v2/builds/f23b0593c54f9ca4a58b065537e932fb453438f7) and on [IOS Safari](https://automate.browserstack.com/dashboard/v2/builds/7455eaa22f5b15e4210cd8a55d9d67959cb0be40) on stage (all tests pass).

Issue #1388, opened 2025-12-13
Ensure that the avatar in Appointment links back to TB Pro accounts.


kellymcs01:
**Describe the bug**
Ensure that the avatar in Appointment links back to TB Pro accounts.

**Expected behavior**
When clicking on the Avatar in the upper right hand corner of the window expectation was that it would take you to the TB Pro Dashboard. 

**Actual behavior**
When clicking the Avatar it took us to the old Appointment Dashboard.

Issue #1396, opened 2026-01-05
Inputs have mismatched styles


davinotdavid:
**Describe the bug**
When the `AUTH_SCHEME=password` in the `.env` file, the password input has some mismatching styles inherited from Tailwind. Can also reproduce this in stage: https://appointment-stage.tb.pro 

**To Reproduce**

1. Change the `VITE_AUTH_SCHEME` to `password` in the `frontend/.env` file
2. Navigate to `localhost:8080/login`
3. See the mismatch

**Expected behavior**
The input should follow the same UI styles than the email input.

**Actual behavior**
The password input has some background transparency and a border.

**Screenshots**

Light:
<img width="1505" height="963" alt="Image" src="https://github.com/user-attachments/assets/f379de1e-c1e4-4d03-9ed2-a8f6698cd1d6" />

Dark:
<img width="1506" height="961" alt="Image" src="https://github.com/user-attachments/assets/2cf50c8e-dc1d-4a4f-9965-8b13c63fc5ad" />

davinotdavid:
Apparently not only for the password input but also for time. Also reproducible on https://appointment-stage.tb.pro

<img width="1501" height="960" alt="Image" src="https://github.com/user-attachments/assets/4d190604-194a-486a-8c03-28af39ee97a2" />

davinotdavid:
Might be worth to remove Tailwind altogether as proposed by https://github.com/thunderbird/appointment/issues/495

davinotdavid:
Closing as it is now fixed on Stage and will make its way into prod in the next deployment!

PR #1401, opened 2026-01-06
Fix UserAvatar should link back to TB Accounts

davinotdavid:
## Description of the Change

- Removed the drop-down from `UserAvatar` and made it a link to TB Accounts' Dashboard instead per the issue description
- Updated NavBarMobile's User Profile link to point to TB Accounts' Dashboard instead
- Overrides `UserAvatar` in the NavBar to match Zeplin
- Slightly increase the Appointment logo in the NavBar header to match Zeplin
- Updates `services-ui` to `1.4.3`
- Fixes the styles for `input[type="email]` in the Booker's page ([Tailwind upstream issues residue](https://github.com/thunderbird/appointment/issues/1396))
- Removed the `ProfileView` route + component as it is now unreachable (it was present in the drop-down only)
- Small refactor on NavBar and NavBarMobile tests
- Make Booking Page URL in Settings page `readonly`

> [!NOTE]
> Please update the `frontend/.env` file to include `VITE_TB_ACCOUNT_DASHBOARD_URL=https://accounts.tb.pro/dashboard` if you don't have it set to anything else.

## Screenshots

Before / NavBar
image

After / NavBar
image

## Benefits

- More UI/UX cohesiveness according to the design team

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/1388

### DIFF
--- a/frontend/test/components/NavBarMobile.test.js
+++ b/frontend/test/components/NavBarMobile.test.js
@@ -7,11 +7,22 @@ import withSetup from '../utils/with-setup';
 import { useUserStore } from '@/stores/user-store';
 import { RouterLink } from 'vue-router';
 import NavBarMobile from '@/components/NavBarMobile.vue';
+import { accountsTbProfileUrlKey } from '@/keys';
 
 
 describe('NavBarMobile', () => {
   var app;
   var wrapper;
+  const testAccountsTbProfileUrl = 'https://accounts.tb.pro/dashboard';
+
+  const getMountOptions = () => ({
+    global: {
+      plugins: [i18ninstance, router],
+      provide: {
+        [accountsTbProfileUrlKey]: testAccountsTbProfileUrl,
+      },
+    },
+  });
 
   beforeEach(() => {
     app = withSetup();
@@ -23,11 +34,7 @@ describe('NavBarMobile', () => {
   });
 
   it('renders correctly when not logged in', () => {
-    wrapper = mount(NavBarMobile, {
-      global: {
-        plugins: [i18ninstance, router],
-      },
-    });
+    wrapper = mount(NavBarMobile, getMountOptions());
 
     // verify all expected router-link child components were rendered (only one when not signed in)
     const allRouterLinks = wrapper.findAllComponents(RouterLink);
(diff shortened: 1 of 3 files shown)


frontend/src/views/SettingsView/components/AccountSettings.vue
```diff
@@ -105,7 +105,7 @@ const actuallyDownloadData = async () => {
       {{ t('label.bookingPageURL') }}
     </label>
     <div class="booking-page-url-input-container">
-      <text-input name="booking-page-url" class="booking-page-input" v-model="userStore.myLink" />
+      <text-input name="booking-page-url" class="booking-page-input" v-model="userStore.myLink" readonly />
```
devmount, line 108: For some reason, this doesn't work on my end. It adds a readonly attribute on the label, not on the input... we might need to check services-ui for this issue.

I can still edit that input and get some error messages doing so:

Image

frontend/test/components/NavBarMobile.test.js
```diff
@@ -67,24 +70,19 @@ describe('NavBarMobile', () => {
...
+    wrapper = mount(NavBarMobile, getMountOptions());
 
     // click to open/show the menu
     const menuButton = wrapper.find('button[aria-label="Open menu"]');
     await menuButton.trigger('click');
 
     // verify all expected router-link child components were rendered when signed in and menu is now shown
     const expectedLinks = [
-      'dashboard', // there are two dashboard rounter-link child components
+      'dashboard', // there are two dashboard router-link child components
```
devmount, line 81: whoops, nice catch!

devmount (commented):
Oh and the user avatar has no href on my end. But I guess I need a corresponding .env entry for TB accounts, right?

frontend/src/views/SettingsView/components/AccountSettings.vue
davinotdavid: Oh interesting! Here, the `readonly` attribute seems to be applied both to the label and the input itself for me 🤔 Even after you rebuild the containers with a `docker compose up -d --build` it still shows the error? I've deleted my containers just in case and it still works for me

davinotdavid:
> Oh and the user avatar has no href on my end. But I guess I need a corresponding .env entry for TB accounts, right?

@devmount ah yes, I forgot to add this to the instructions in the PR, good catch! It needs a `frontend/.env` for the `VITE_TB_ACCOUNT_DASHBOARD_URL`. Updated in the PR description now!

devmount (approved):
Ohh you're right! I did `npm install` but didn't restart the frontend container. Now everything works like a charm! Thanks!

Ah thanks for adding that env var, the link is now there too. Nothing to complain about anymore 😂 Thanks for the great work, approved!

Issue #1389, opened 2025-12-13
Increase the size/visibility of the "Availability saved successfully" confirmation message in Appointment.


kellymcs01:
**Is your feature request related to a problem? Please describe.**
The message that is displayed to indicate successful saving of availability uses a smaller font than most of the other text and is a little difficult to read. 

**Describe the solution you'd like**
Would like to have the font size increased so it's equal to the non-heading font size on the rest of that page.

kewisch:
@kellymcs01 Do you have a screenshot for this?

davinotdavid:
Judging from the recording, it seems that the desired sizing is the default one for the `NoticeBar` component.

I will follow the same sizing as the Storybook in `services-ui` for consistency:
https://thunderbird.github.io/services-ui/?path=/story/components-noticebar--success&globals=theme:lightApmt

PR #1402, opened 2026-01-06
Fix Availability Page and Settings Page NoticeBar styles

davinotdavid:
## Description of the Change

- We were using the `AlertBox` component (present only in Appointment, not in `services-ui`) for the success / failures of the Availability page and Settings page. However, the `AlertBox` is a wrapper for `NoticeBar` changing some of its styles, hence the visual difference between the warning state and the success state font-sizes.
- This PR normalizes the component used to be `NoticeBar`s instead, also making it sticky and making sure that only one NoticeBar is present at a time.

## Screenshots

Before / Availability

image

After / Availability

image

--

Before / Settings
image

After / Settings
image

## Benefits

- No more overlapping of NoticeBar / AlertBox
- Better hitbox for the close button
- Consistent font-size across NoticeBars

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/1389

### DIFF
--- a/frontend/src/views/AvailabilityView/index.vue
+++ b/frontend/src/views/AvailabilityView/index.vue
@@ -110,10 +110,13 @@ async function onSaveChanges() {
   })
 }
 
-function onRevertChanges() {
+function clearNotices() {
   validationError.value = null;
   saveSuccess.value = null;
+}
 
+function onRevertChanges() {
+  clearNotices();
   availabilityStore.revertChanges();
 }
 </script>
@@ -166,7 +169,7 @@ export default {
 
       <template #cta>
         <icon-button
-          @click="validationError = null"
+          @click="clearNotices"
           :title="t('label.close')"
           class="btn-close"
         >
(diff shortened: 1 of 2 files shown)


davinotdavid:
@devmount thanks for the review! So there are some of those instances where I think it makes sense to have a smaller version of it, for example in the `GenericModal` since there's less space all around.

IMHO we can keep the `AlertBox` for these instances for now but I'd rather have a variant in the `NoticeBar` component in `services-ui` that is "dense" or "sm" for these cases. However, IIRC we don't have a design for it planned yet :(

What do you think?

devmount:
> IMHO we can keep the `AlertBox` for these instances for now but I'd rather have a variant in the `NoticeBar` component in `services-ui` that is "dense" or "sm" for these cases. However, IIRC we don't have a design for it planned yet :(
> 
> What do you think?

Yes, agree! A variant for the NoticeBar would be best. I'll create an issue in services-ui for this and see, if we can implement that there.

davinotdavid:
@devmount sounds good, thanks! I will merge this then 🙇

Issue #1382, opened 2025-12-13
Update the "connecting Zoom accounts" banner text to reflect current app approval status.


kellymcs01:

kewisch:
@kellymcs01 which banner was this, the one on the main page, or a different one?

<img width="923" height="174" alt="Image" src="https://github.com/user-attachments/assets/35d09748-236b-4386-8075-07579260f53a" />

kewisch:
If it is this banner I would remove it completely, even for community release.

kewisch:
Yes confirmed, at 1:08:50 https://mozilla.hosted.panopto.com/Panopto/Pages/Viewer.aspx?id=0e5ed8cd-2908-4cb7-a5ac-b3af0176db84&start=4123.551251

Let's remove the banner.

PR #1404, opened 2026-01-07
Remove Zoom banner and related files

davinotdavid:
## Description of the Change

- Removed the "Zoom banner" / "Beta Warning". Note that this also removes quite a bit of extra code that _could_ be useful in the future but was unused except for this beta warning (`UserActivityStore` and related). In case we need to add dismisable warnings we can refer back to this PR and re-add it!

## Benefits

- Less user confusion around being beta / not beta as it does not reflect the limitation we had before
- Better mobile experience as the banner took quite a bit of space in the dashboard

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/1382


davinotdavid:
> Are you sure you want to remove the underlying code for dismisable warnings (even though it can be added back)? One example use I can maybe see is if we ever need to plan some downtime for maintenance.

@rwood-moz  That's a good point, I agree that this is very likely to make a come back. It is either a question of leaving unused / dead code in the code base for the potential use in the future or having only reachable code.

IMHO it is a small lift to re-add it if needed and I'd prefer to reduce code when we can if not used but I'm also ok to keep it if you all prefer since it is a small change anyways.

rwood-moz:
> @rwood-moz That's a good point, I agree that this is very likely to make a come back. It is either a question of leaving unused / dead code in the code base for the potential use in the future or having only reachable code.
> 
> IMHO it is a small lift to re-add it if needed and I'd prefer to reduce code when we can if not used but I'm also ok to keep it if you all prefer since it is a small change anyways.

I'm fine with whatever you think is best. Either way, I tried out your branch locally, went through the FTUE, and didn't see the banner anymore after navigating to the dashboard. Awesome!

devmount (approved):
Thanks Davi. I'm fine with deleting everything, it's actually quite a lot for one banner, if I think about it 😅

Issue #1286, opened 2025-10-02
Update Settings page with new hi-fi designs


davinotdavid:
**Describe the solution you'd like**
Match as close as possible with the [Zeplin designs](https://app.zeplin.io/project/66dd0f2b2fe8f1b63cda36f1/screen/68cd9885ebfdb53d68206d7e):

<img width="1280" height="2150" alt="Image" src="https://github.com/user-attachments/assets/6a3d0efb-51f2-4815-ad88-4a1b40a371ef" />

davinotdavid:
Moving this over to the community release since there's already a PR in review for it!

cc @kewisch @malini

malini:
nice!

Issue #1295, opened 2025-10-08
Start of week setting doesn't change dashboard or bookee page calendar view


rwood-moz:
**Describe the bug**
Changing the `Start of week` setting doesn't have any effect on the dashboard or bookee page calendars.

**To Reproduce**

- Sign into the Appointment dashboard
- Notice the dashboard calendar displayed starts with `Sun`
- Go into Settings => Preferences and set start of week to `M` for Monday and save
- Go back to the main dashboard

**Expected behavior**
Expected the calendar week to be displayed starting with `Mon`.

**Actual behavior**
- The calendar week displayed still starts at `Sun` (even after a page refresh)
- Same goes for the calendar on the bookee page

**Additional context**
The only effect that I can see that changing the `Start of week` setting has is on the Availability page, the day selection buttons listed under `Available days and times` change their order based on the start of week setting (after refreshing).

davinotdavid:
Related https://github.com/thunderbird/appointment/issues/1237

Issue #901, opened 2025-03-04
24hr format not applied to availability start and end time settings


rwood-moz:
**Describe the bug**
When in 24hr time setting mode I notice that 24hr format is not applied in the availability start and end time settings.

**To Reproduce**
- Go into `Settings` => `General`, and set `Date and Time Format` to `24hr`.
- Go back to the `Dashboard` and notice 24hr has taken effect in the calendar as expected (appointment slots are in 24hr)
- Still on the Dashboard expand the `Availability` panel on the right side

**Expected behavior**
Expect the `Start time` and `End time` field values to be in 24hr format.

**Actual behavior**
The `Start time` and `End time` fields remain in 12hr format.

rwood-moz:
Also after changing to 24 hour time format:

- The dashboard 'grid' / times listed on the left side are still in 12 hr format (although actual events on the dashboard are in 24 hr format)
- All appointments listed on the `Bookings` tab are listed in 12 hr format
- On the bookee page (share link page) all time slots and the entire calendar is in 12 hr format; and when you select a time slot the booking details are also in 12 hr format

So basically changing the time format setting to 24 hr format really doesn't do anything except change the time inside any displayed events themselves on the dashboard to 24 hr time. Is this what is intended for this setting? Do we even want to keep this setting if that is all it is changing? Or should the setting be renamed to indicate it only changes the time format in events? Stuff to think about.

lisajill:
I'm in favor of keeping this setting and having it impact all areas of the dashboard. Especially important for some regions as we grow an international audience, and a strong preference with some individuals (myself included).

kewisch:
I believe we should remove this setting and rely on `Intl.DateTimeFormat().resolvedOptions()`. There may be some cases where locale doesn't match OS settings, but given the conversion is much easier than lets say C to F, I think that is acceptable. I believe Chrome already provides the OS setting through `Intl.DateTimeFormat().resolvedOptions()` while Firefox has traditionally been more locale-based.

This will give us a good default without needing extra settings.

devmount:
@kewisch Alright, I can take care of the removal of that setting.

davinotdavid:
@devmount just double checking here since you're actively working in other issues too, is this something you're working on at the moment or would you prefer me to tackle it?

Asking since there has been yet another customer asking about it and I have some bandwidth to fix it if need be!

devmount:
[...]

PR #1425, opened 2026-01-13
Update Settings Page UI

davinotdavid:
## Description of the Change

- Updated the Settings Page UI with the latest UI designs
- Updated the `NoticeBar` sizing and sticky behaviour to match the Availability page
- Removed `RadioGroupPill` component and tests since they are not used anymore / changed in favour of `SegmentedControl` from `services-ui`

Zeplin mockup:
https://app.zeplin.io/project/66dd0f2b2fe8f1b63cda36f1/screen/68cd9885ebfdb53d68206d7e

## Benefits

- Better UI alignment with the rest of the application

## Screenshots

Light mode
image

Dark mode
image

- There are a few slight deviations from the components in terms of sizing / typography but those are quite minor and should be tackled directly in `services-ui` and will automagically be applied here once the work is done and package is updated.
- The `ConfirmationModal` component used in the `Delete Appointment Data` flow is largely unchanged except for a few color tweaks. This requires a bigger refactor that may impact different screens so [the modal shown in this Zeplin design](https://app.zeplin.io/project/66dd0f2b2fe8f1b63cda36f1/screen/690501b3597ce47550baec83) has not been implemented in this PR. Created an issue for it here: https://github.com/thunderbird/appointment/issues/1426

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/1286

### DIFF
--- a/frontend/src/locales/de.json
+++ b/frontend/src/locales/de.json
@@ -285,7 +285,7 @@
     "delete": "Löschen",
     "deleteBooking": "Buchung löschen",
     "deleteAllAppointmentData": "Alle Appointment-Daten löschen",
-    "deleteAllAppointmentDataInfo": "Deine TB Pro-Abonnement wird nicht beeinflusst.",
+    "deleteAllAppointmentDataInfo": "Dein TB Pro-Abonnement bleibt davon unberührt.",
     "deny": "Ablehnen",
     "denyBooking": "Buchung ablehnen",
     "disable": "Deaktivieren",


rwood-moz:
FYI this breaks a bunch of Appointment E2E tests as can be expected; I've created #1428 to update the tests and will take that on in a separate PR.

rwood-moz:
Just a separate note, changing the time format itself really doesn't do much - I know that is outside of the scope of this PR - but I just wanted to mention that again (I previously created #901).

davinotdavid:
@rwood-moz Thank you so much for the review and for opening the issue for the E2E 🙏 

I wonder if there's a cost effective way for us to run the E2E in the PR checks so that folks (like me 😅) remember / are aware of potential breaks there... or maybe I can add to the PR template as a reminder to run locally?

> In Zeplin the share link field header is Booking Page Link but the implemented header is actually Booking Page URL

Great catch! Missed that completely and added in the latest commit. Thanks!

> The header text for start of the week is Start of the week with only capitalizing the first word; this is inconsistent with other field headers that capitalize all words like Default Time Zone (I realize it is implemented as noted in zeplin)

Agreed! It looks a little off even though it matches Zeplin. I've changed it to the `Start of the Week` instead as a middle ground. Thanks!

> Just to verify we only support SUN and MON for start of week still correct? Just double checking since in zeplin it shows all days of week as options for start of the week

The limitation came from the calendar lib we were using before IIRC but since removing it, we could add all the days there. We still have an open bug about it that even the current two options don't change the calendar display just yet though https://github.com/thunderbird/appointment/issues/1295.

Same for the changing the time format :( good bugs to be tacked next!

davinotdavid:
> However now that we have Thundermail stage online we can connect a Thundermail stage CalDAV calendar instead. 

Oooh that's a great idea! Not sure how the approach would be but it seems like we are running a specific set of steps for testing the FTUE for Community Release during the live walkthroughs. It could be a cool project to port those into E2E!

Thanks for the review again!

frontend/src/locales/de.json
```diff
@@ -170,8 +168,8 @@
     "slotIsAvailableAgain": "Das Zeitfenster ist jetzt wieder buchbar."
   },
   "label": {
-    "12hAmPm": "12-hour AM/PM",
-    "24h": "24-hour",
+    "12hAmPm": "12:00 Uhr (12-Stunden-Format)",
```
devmount, line 171: Nice catch, thanks!

frontend/src/locales/de.json
```diff
@@ -288,6 +284,8 @@
     "defaultTimeZone": "Standard Zeitzone",
     "delete": "Löschen",
     "deleteBooking": "Buchung löschen",
+    "deleteAllAppointmentData": "Alle Appointment-Daten löschen",
+    "deleteAllAppointmentDataInfo": "Deine TB Pro-Abonnement wird nicht beeinflusst.",
```
```suggestion
    "deleteAllAppointmentDataInfo": "Dein TB Pro-Abonnement bleibt davon unberührt.",
```

frontend/src/locales/de.json
```diff
@@ -550,6 +548,7 @@
     "chooseDateAndTime": "Wähle Tag und Zeit für ein Treffen.",
     "connectZoom": "Du kannst dein Zoom-Konto verbinden, um Besprechungen direkt mit einer Zoom-Einladung zu erstellen.",
     "contactRequestForm": "Bitte verwende das nachstehende Kontaktformular, um Feedback, Fragen oder Anliegen an unser Support-Team zu senden. Wir werden ggf. versuchen, so schnell wie möglich zu antworten.",
+    "deleteAppointmentDataWarning": "Du wirst keinen Zugriff mehr auf Thunderbird Appointment haben. Alle Buchungs-Kalender werden auf \"Inaktiv\" gesetzt und alle zukünftigen Buchungen werden storniert.\nAndere Termin-Daten werden 30 Tage lang beibehalten, entsprechend der TB Pro Datenschutzrichtlinie.",
```
devmount, line 551: Nice, this sounds very natural 👍🏻

frontend/src/views/SettingsView/components/AccountSettings.vue
```diff
@@ -93,148 +92,141 @@ const actuallyDownloadData = async () => {
...
-  <!-- TODO: This should be account _cancellation_ instead -->
-  <confirmation-modal
-    :open="cancelAccountModalOpen"
-    :title="t('heading.cancelAppointmentService')"
-    :message="t('text.accountCancelWarning')"
-    :confirm-label="t('label.cancelService')"
-    :cancel-label="t('label.cancel')"
-    :use-caution-button="true"
-    @confirm="actuallyDeleteAccount"
-    @close="cancelAccountModalOpen = false">
```
devmount, line 162: Honestly, I found it better readable before. But that's very opionated 😇  Is that our formatting configuration?

frontend/src/locales/de.json
davinotdavid: Thank you so much! 🙏

frontend/src/views/SettingsView/components/AccountSettings.vue
davinotdavid: I agree :( My editor is configured to use the Vue extension as a formatter but I think we should settle on a Prettier config of sorts to standardize so we can all turn on "format on save" or something similar. Do you use something different / How are you auto-formatting the code on your end?

devmount: I don't 😇 😂 Haven't found the perfect auto-formatter for HTML yet 😅

davinotdavid: Ooooh I see! That's a good code standards bit to set and forget, I opened up a issue for discussing this and creating a .json / .rc config of sorts: https://github.com/thunderbird/appointment/issues/1430

Issue #1237, opened 2025-09-09
Dashboard start of week display is changed after changing language setting


rwood-moz:
**Describe the bug**
Changing the language setting is also causing start of week dashboard display to change.

**To Reproduce**

- I have have `Sunday` set for the `Start of Week` setting, and `EN` for the language
- Go to the dashboard, notice the first day displayed is Sunday as expected
- Now go into `Settings => Preferences` and change the language from EN to DE, and save
- Now click on the `Dashboard` tab to go back to the Dashboard
- When back on the Dashboard, click the browser refresh button to refresh the page
- Notice the start of week displayed is now `Monday` even though it was never changed in settings

**Expected behavior**
Changing the language should not effect the start of week displayed in the dashboard.

**Actual behavior**
After changing the language the start of week in the dashboard is changed after a refresh. Note that the actual start of week setting remains as Sunday even though it is now appearing as Monday in the actual dashboard.

**Additional context**
The start of week saved to local storage remains as `7` even though the dashboard now shows Monday as the start of week.

davinotdavid:
oh I believe this is on purpose as iirc `DE` week starts on a `Monday` localization-wise instead of a `Sunday`

rwood-moz:
> oh I believe this is on purpose as iirc `DE` week starts on a `Monday` localization-wise instead of a `Sunday`

Interesting. It looks like the start of week setting is being ignored then though, as it is still set to Sunday in settings even though the dashboard switches to Monday after changing the language.

PR #1433, opened 2026-01-15
Make `startOfWeek` setting work properly alongside different locales

davinotdavid:
## Description of the Change

- Fixes `startOfWeek` setting not being applied to the Calendar component. This was due to a mismatch between the `isoWeekdays` vs the `dayJS` work days.
- Fixes week options being scrambled in Settings page when select and saving a different start of the week.
- Fixes a locale problem when switching languages where the week day options in Settings and the calendar in the Dashboard wouldn't show the translated week days until a full page refresh.
- Added all 7 days as options to start of the week since this is not a limitation anymore.

> [!NOTE]
> The Bookee / Booking page still shows the calendar starting with the weekday based off of the locale of the user seeing said page, which I believe is the correct behaviour but in case we want to enforce a different setting, I can update this

## Screenshot / Video Walkthrough

https://github.com/user-attachments/assets/64889a3d-ef36-4723-a31e-01475c01baf6

## Benefits

- Start of the Week actually works / gets respected in the Dashboard
- We now have all 7 days as starting points for the calendar

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/1295
Fixes https://github.com/thunderbird/appointment/issues/1237

### DIFF
--- a/frontend/src/views/DashboardView/components/WeekPicker.vue
+++ b/frontend/src/views/DashboardView/components/WeekPicker.vue
@@ -4,8 +4,11 @@ import { useI18n } from 'vue-i18n';
 import { dayjsKey } from '@/keys';
 import { PhCaretLeft, PhCaretRight } from '@phosphor-icons/vue';
 import { TimeFormatted } from '@/models';
+import { useUserStore } from '@/stores/user-store';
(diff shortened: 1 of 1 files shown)


rwood-moz (approved):
LGTM and worked well when I gave it a try locally. Nice!

PR #1434, opened 2026-01-15
Update settings E2E tests for settings page UI overhaul

### LINKED ISSUE #1428: Update E2E tests for latest settings page UI overhaul
The settings page UI overhaul (#1425) breaks several E2E tests as can be expected. Update the E2E tests to work with the new settings UI.

rwood-moz:
Fixes #1428. Update the settings E2E tests so they work again after the settings page UI overahaul (#1425).

BrowserStack links running the updated tests on:

- Android Chrome is [here](https://automate.browserstack.com/dashboard/v2/builds/41c61f56d7c7aebb90b49a541cdaf41e51f7bdde)
- iOS Safari is [here](https://automate.browserstack.com/dashboard/v2/builds/a3a92aa7dd0f4eef1d73464c55d612cf7339eed2)
- Firefox Desktop is [here](https://automate.browserstack.com/dashboard/v2/builds/59726a9f500cbd20690cbe22fb491701dfbccaa5)

### DIFF
--- a/test/e2e/pages/settings-page.ts
+++ b/test/e2e/pages/settings-page.ts
@@ -8,6 +8,7 @@ import {
   TIMEOUT_5_SECONDS,
   TIMEOUT_30_SECONDS,
   APPT_LANGUAGE_SETTING_EN,
+  APPT_START_OF_WEEK_MON,
   } from '../const/constants';
 
 
@@ -31,9 +32,9 @@ export class SettingsPage {
   readonly bookingPageURLInput: Locator;
   readonly copyLinkBtn: Locator;
   readonly copyLinkToolTipText: Locator;
-  readonly cancelServiceBtn: Locator;
-  readonly cancelServiceConfirmCancelBtn: Locator;
-  readonly bookingPageSettingsBtn: Locator;
+  readonly deleteDataBtn: Locator;
+  readonly deleteDataConfirmCancelBtn: Locator;
+  readonly manageBookingLink: Locator;
   readonly downloadDataBtn: Locator;
   readonly connectedAppsHdr: Locator;
   readonly addCaldavBtn: Locator;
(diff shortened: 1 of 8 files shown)


test/e2e/pages/settings-page.ts
```diff
@@ -57,10 +58,10 @@ export class SettingsPage {
     // main settings view
     this.settingsHeaderEN = this.page.getByRole('main').getByText('Settings', { exact: true });
     this.settingsHeaderDE = this.page.getByRole('heading', { name: 'Einstellungen' }).first();
-    this.saveBtnEN = this.page.getByRole('button', { name: 'Save' });
+    this.saveBtnEN = this.page.getByRole('button', { name: 'Save' }).nth(1); // save button at bottom, not in notice bar
```
davinotdavid, line 61: Interesting! There's a save button in the notice bar though, is the bottom one preferable?

Image

test/e2e/utils/utils.ts
```diff
@@ -36,14 +36,14 @@ export const navigateToAppointmentAndSignIn = async (page: Page, testProjectName
...
   await page.waitForTimeout(TIMEOUT_5_SECONDS);
 
-  // if we are already signed in then we can skip this
+  // local dev can use 'password' only or oidc; check if using password only
+  if (APPT_TARGET_ENV == 'dev' && await tbAcctsSignInPage.localDevEmailInput.isVisible()) {
+    await tbAcctsSignInPage.localApptSignIn();
+  }
+
+  // using oidc/keyloak; if we are already signed in then we can skip this
   if (await tbAcctsSignInPage.signInHeaderText.isVisible() && await tbAcctsSignInPage.signInButton.isEnabled()) {
```
davinotdavid: I wonder if this should be in an `else` block since this will get evaluated regardless right? As in, even if the `APPT_TARGET_ENV` is `dev`, it will still try to await for the `signInHeaderText` I believe

test/e2e/pages/settings-page.ts
rwood-moz: Yes, on mobile clicking the save button in the notice bar at the top directly using playwright doesn't work because it is covered up by a div element (although it works on desktop) but clicking on the save button directly at the bottom via playwright works on mobile and desktop.

test/e2e/utils/utils.ts
rwood-moz: Yes good point, I'll update that, thanks for the review.

test/e2e/pages/settings-page.ts
davinotdavid: Oooh I see! Interesting and makes sense, thanks!

Issue #1435, opened 2026-01-16
Change typeface from OpenSans to Inter / Metropolis


davinotdavid:
Currently, we are defaulting the sans serif variant to OpenSans in our Tailwind config. We should change that to Inter and remove the font imports.

It would also be nice to audit a little bit of redundant CSS declarations that set the font-family across the application.

Also, let's guarantee that the base font size is 1rem -> 16px.

PR #1437, opened 2026-01-16
Cleanup typography + Remove maintenance page

davinotdavid:
## Description of the Change

- Removes old typography (Open Sans, Raleway and Roboto) leaving only Inter and Metropolis.
- Uses "Inter" as the standard sans serif in the Tailwind config (while we don't fully remove Tailwind).
- Removes the unused / unreachable maintenance page. To add one again, we need to update CloudFront configurations in order to actually serve it so it wouldn't be a simple style change).

## How to test

- Clean & build all containers locally and navigate through the pages.
- Inspect various text on screen and check the CSS for `font-family`. It should always be either `Inter` or `metropolis`.

## Benefits

- Less unused imports and better cohesion in the UI regarding typography.

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/1435


MelissaAutumn (approved):
lgtm, we use to have a font selector in settings so that's why we had so many fonts available.

davinotdavid:
@MelissaAutumn oooh interesting! I didn't know that, thanks for the background info!

Issue #1436, opened 2026-01-16
Favicon missing


kewisch:
The booking link doesn't appear to have a favicon.

It tries to load https://appointment.tb.pro/favicon-dark.ico which does not exist

davinotdavid:
Related, this seemed to be working around a month ago through this PR:

https://github.com/thunderbird/appointment/issues/1353

MelissaAutumn:
I'm no expert, but I think this might be related to the re-write function we use to smash frontend and api together and also fix s3's expectation of /index.html.
https://github.com/thunderbird/appointment/blob/main/pulumi/cloudfront-rewrite.js#L6

We might want to move favicon and sitemanifest to assets, or add an exception for them in the rewrite.

MelissaAutumn:
This is now fixed on stage. When we do a production push we'll need to manually run pulumi up targetting the re-write function as github deploys don't touch this function.

davinotdavid:
@MelissaAutumn good call! There's a walkthrough meeting set up for Monday morning so I will deploy and do this right after. Thanks!

davinotdavid:
Confirmed that it is now working in production too after the pulumi up targetting the re-write function!

https://app.pulumi.com/mzla-services/appointment/prod/updates/65

inattee:
The current favicon doesn't display well in browsers set to light mode. @laurelterlesky 

<img width="295" height="89" alt="Image" src="https://github.com/user-attachments/assets/0cf579b3-be37-433a-93d0-f95e2726406b" />

MelissaAutumn:
We removed no-preference as it's deprecated, but it might make sense to re-add it or remove the media query from the lightmode icon.

davinotdavid:
That's interesting, it shows correctly on Chrome but not on Firefox on the light theme:

Chrome
<img width="395" height="202" alt="Image" src="https://github.com/user-attachments/assets/17a4cc0f-5972-45dd-a607-5224abe492cb" />

Firefox
<img width="576" height="292" alt="Image" src="https://github.com/user-attachments/assets/ae94ce7d-148c-44f5-91d8-e70c400d9604" />

MelissaAutumn:
Ah spec issues. Might be a good bugzilla ticket too if its not already on there!

davinotdavid:
Indeed a long standing issue on Firefox:

https://bugzilla.mozilla.org/show_bug.cgi?id=1603885
[...]

PR #1438, opened 2026-01-16
Fix favicon media attribute

davinotdavid:
## Description of the Change

- Remove the `(prefers-color-scheme: no-preference)` since it is not supported by browsers anymore and causes the favicon to not load.

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@media/prefers-color-scheme

Screenshot 2026-01-16 at 11 51 28 AM

## How to test

- Load the application and check that the favicon now loads as expected in both light and dark modes.

## Benefits

- Favicon works again 🎉 

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/1436

### DIFF
--- a/pulumi/cloudfront-rewrite.js
+++ b/pulumi/cloudfront-rewrite.js
@@ -3,7 +3,19 @@
     const headers = request.headers;
     const host = headers.host.value;
     const apiPath = '/api/v1';
-    const ignorePaths = ['/fxa', '/assets', '/appointment_logo.svg', '/sitemap.txt'];
+    const ignorePaths = [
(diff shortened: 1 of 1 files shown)


MelissaAutumn:
I don't think this will fully fix #1436 as https://appointment.tb.pro/favicon.ico is not accessible either. It work on localhost, but we might be an issue with serving vue assets.

davinotdavid:
@MelissaAutumn alright, I think that should do it! Just in case I added all the files to the CloudFront rewrite!

Issue #1441, opened 2026-01-19
Cannot connect / disconnect calendar in the Settings page


davinotdavid:
**Describe the bug**
The checkbox from the connect / disconnect calendar in the Settings page doesn't work. Also, the state of the checkbox don't reflect the underlying data model so other features (e.g. Availability "Booking To" select) seem broken as well.

**To Reproduce**

1. During the FTUE, connect a Google Account or a CalDAV with multiple calendars
2. Go to the Settings page
3. Click on any of the checkbox of the calendars listed in the page

**Expected behavior**
The calendars should trigger the notice bar for saving changes, and saving changes should actually work for connecting and disconnecting calendars.

**Actual behavior**
Clicking on the checkboxes causes an error and the state of the checkbox doesn't reflect their actual state.

PR #1445, opened 2026-01-20
Fix calendar connection / disconnection in Settings page

davinotdavid:
## Description of the Change

- The `CheckboxInput` component from the calendars in the Settings page was still using the `@checked` event handler instead of the `v-model`. This caused the component not to behave properly so its state was not an accurate reflecting the underlying data of whether or not the calendar was actually connected and furthermore couldn't be toggled at all.

- We were preventing the default calendar to be disconnected by not showing the dropdown menu options for said calendar because it would cause problems if this calendar is the one set in the Availability. However, we were still allowing other calendars that share the same external connection to be disconnected which is virtually the same since disconnecting a calendar that share the same external connection as the default calendar would be the same as disconnecting the default calendar directly. In this PR, we are now preventing disconnecting calendars that share the same external connection as the default calendar.

## How to test

0. Go through FTUE / have a user setup with a valid external connection (either CalDAV or Google Calendar) and multiple calendars associated with it
1. Go to `Settings` page and scroll down to see the list of calendars
2. See that the non-default calendars in the list are unchecked by default
3. See that checking an unchecked checkbox (wow) triggers the "You have unsaved changes" notice bar and said notice behaves as expected
4. Connect a secondary CalDAV / Google Calendar account
5. Verify that the the secondary CalDAV / Google Calendar account's calendar(s) have the dropdown menu
6. Connect a calendar from the secondary CalDAV / Google Calendar account and mark it as the default calendar
7. See that the previously default CalDAV / Google Calendar account's calendar(s) now show the dropdown menu for disconnection
8. (Bonus) Check that every time you change the default calendar in Settings, it also changes in the Availability page and the Availability page's "Booking to" dropdown shows all the connected calendars in the dropdown

## Benefits

- Calendar connection / disconnection works again
- More E2E tests to test this behaviour

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/1441

### DIFF
--- a/backend/test/integration/test_auth.py
+++ b/backend/test/integration/test_auth.py
@@ -581,6 +581,79 @@ class TestCalDAV:
             calendar = repo.calendar.get(db, calendar.id)
             assert calendar is None
 
+    def test_disconnect_fails_when_connection_contains_default_calendar(
+        self, with_db, with_client, make_external_connections, make_caldav_calendar, make_schedule
+    ):
+        """Ensure we cannot disconnect an external connection that contains the default calendar (used by a schedule)"""
+        username = 'username'
+        type_id = json.dumps(['url', username])
+        ec = make_external_connections(TEST_USER_ID, type=models.ExternalConnectionType.caldav, type_id=type_id)
+        calendar = make_caldav_calendar(
+            subscriber_id=TEST_USER_ID, user=username, connected=True, external_connection_id=ec.id
+        )
+
+        # Create a schedule that uses this calendar (making it the "default" calendar)
+        make_schedule(calendar_id=calendar.id)
+
+        # Attempt to disconnect - should fail
+        # Note: type_id is a query parameter in the caldav disconnect route, not a body parameter
+        response = with_client.post('/caldav/disconnect', params={'type_id': ec.type_id}, headers=auth_headers)
+
+        assert response.status_code == 400, response.content
+        data = response.json()
+        assert data['detail']['id'] == 'CONNECTION_CONTAINS_DEFAULT_CALENDAR'
+
+        # Verify the connection and calendar still exist
+        with with_db() as db:
+            ecs = repo.external_connection.get_by_type(
+                db, TEST_USER_ID, models.ExternalConnectionType.caldav, type_id=type_id
+            )
+            assert len(ecs) == 1
+
+            calendar_check = repo.calendar.get(db, calendar.id)
+            assert calendar_check is not None
+
+    def test_disconnect_succeeds_when_different_connection_contains_default_calendar(
+        self, with_db, with_client, make_external_connections, make_caldav_calendar, make_schedule
+    ):
+        """Ensure we can disconnect an external connection when a different connection contains the default calendar"""
+        # Create first connection with calendar that will be the default
+        username1 = 'username1'
+        type_id1 = json.dumps(['url1', username1])
+        ec1 = make_external_connections(TEST_USER_ID, type=models.ExternalConnectionType.caldav, type_id=type_id1)
+        calendar1 = make_caldav_calendar(
+            subscriber_id=TEST_USER_ID, user=username1, connected=True, external_connection_id=ec1.id
+        )
+
+        # Create second connection with calendar that we'll try to disconnect
+        username2 = 'username2'
+        type_id2 = json.dumps(['url2', username2])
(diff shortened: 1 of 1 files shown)


frontend/src/views/SettingsView/components/ConnectedApplications.vue
```diff
@@ -40,12 +40,22 @@ const availabilityStore = useAvailabilityStore();
...
 const zoomAccount = computed(() => externalConnectionStore.zoom[0]);
 
 // This computed property refers to the calendars from the backend combined
 // with some data from external connections to facilitate connection / disconnection.
 // It should be immutable as it represents the current state of the backend data.
 const initialCalendars = computed(() => {
+  // Find the default calendar's external_connection_id to determine which
+  // calendars share the same connection. Calendars sharing the default's
+  // connection cannot be disconnected without affecting the default calendar.
+  const defaultCalendar = calendars.value?.find(
```
MelissaAutumn, line 54: I feel like this needs to be a backend check as well to avoid validation bypass.

devmount (commented):
Thanks Davi, looks great. One question though: What about a CalDAV connection with more than one calendar? E.g. I have a Nextcloud with 2 calendars, both showing up correctly in Appointment. But as soon as I set one of them as default, the other one loses it's 3-dots-menu. Is this intended behaviour or just wrong behaviour in my weird local dev application state? 😅 

Image

Other than that, everything in your how-to-test section works perfectly fine on my end. Btw. thanks for always providing perfect PR descriptions!!

davinotdavid:
@devmount thanks for the review and going through so many steps!

Yes, that's intentional. In your screenshot, you have two calendars sharing the same External Connection and one of them has been marked as `default`. If we allowed the disconnection of the non-default calendar that shares the same External Connection, both of them would be disconnected which would break the Availability

frontend/src/views/SettingsView/components/ConnectedApplications.vue
davinotdavid: Good call!

Added in this same PR here for completeness:

This commit adds the guards in Google and CalDAV `/disconnect` routes: https://github.com/thunderbird/appointment/pull/1445/commits/1c36cb782cbc35f4414bcedd1dd0a0c8b870add2

This commit adds the tests: https://github.com/thunderbird/appointment/pull/1445/commits/c0b29fb06ab267d8cc81dc015424b04f1dc929d6

(Note to self: a little weird that CalDAV gets the type_id as query param and Google gets it as body 🤔 perhaps a change for another time heh)

rwood-moz:
Hey @davinotdavid I'm looking at your E2E tests, I ran them on desktop and mobile. They work on desktop but the mobile tests you added fail on android and iOS. Perhaps you didn't try running your new E2E tests on mobile before merging? The test/e2e/README has instructions on how to run on BrowserStack but feel free to ping me for any help :)

Also if you want to add more calendar connections (i.e. so the `verify calendar checkbox change shows notice bar and revert restores previous state` test isn't skipped) you'll find the E2E test credentials for CI in 1Password just FYI.

The mobile connected accounts test updates fail on iOS because playwright doesn't support the `locator.count` method used when you're getting a list of all connected calendars; playwright does support that method on android. There is a different issue with the new tests on android though and the test's aren't finishing there, I haven't looked into that further.

BrowserStack link for your updated connected apps test running on [iOS here](https://automate.browserstack.com/dashboard/v2/builds/795d0ce3b52560f65592419c546cb549d1a9d5f6) and [android here](https://automate.browserstack.com/dashboard/v2/builds/84718bf2ca002bee8dde284144312b4cdfec5e7d).

davinotdavid:
@rwood-moz whoops my bad 😅 I will send a PR shortly removing the E2E tests I've added on mobile for now and create a new ticket to update that later on so that we can benefit for the fix now and save some time to focus on the mobile testing after

rwood-moz:
> @rwood-moz whoops my bad 😅 I will send a PR shortly removing the E2E tests I've added on mobile for now and create a new ticket to update that later on so that we can benefit for the fix now and save some time to focus on the mobile testing after

Thanks @davinotdavid Yes unfortunately E2E tests on mobile can be quite challenging vs on desktop due to the Playwright framework and BrowserStack not supporting the same methods/features/options on mobile as they do on desktop, and can take a fair amount of time to debug/tailor for mobile. It's really great that you're adding E2E tests though!

Issue #1442, opened 2026-01-19
Events created directly on a Calendar sometimes don't show in the Dashboard


davinotdavid:
**Describe the bug**
During the FTUE walk through on Jan 19th 2026, we found out that events created in a connected calendar were not being displayed in the `WeekCalendar` component in the Dashboard.

The user connected a Google Account with a valid calendar and went on Google Calendar's app and created a busy event manually for testing. Then, they returned to the Appointment's Dashboard, refreshed the page and even clicked on the manual refresh button and still the calendar event didn't show up.

One interesting aspect of the FTUE process was that the slot duration for the calendar was 90 min. This causes the `WeekCalendar` component to only render chunks of 90 min from the Availability days set and the remote events created directly in the Google Calendar were not in exact chunks of 90 min.

Maybe the reason why this is happening is because the remote events "don't fit" in the slot duration blocks in the `WeekCalendar`.

**To Reproduce**

1. Go through the FTUE with a Google Account that has a valid calendar
2. In the Availability step of the FTUE, choose the maximum slot duration
3. After the FTUE, in the Dashboard, check that you have a much narrower `WeekCalendar`
4. Create a remote event directly in Google Calendar on a period time that is not exactly a chunk of 90 min
5. Refresh the Dashboard page or click on the manual calendar refresh button and check that the events don't show up

**Expected behavior**
The events should show up in the calendar if they are within the bounds of time of the availability. The positioning of the render of the calendar event should reasonably fall within the grid of the calendar.   

**Actual behavior**
Events are not rendered on the `WeekCalendar`.

PR #1459, opened 2026-01-22
Refactor WeekCalendar component + misc fixes

davinotdavid:
## Description of the Change

[TL;DR]:
- WeekCalendar had some big rewrites, now shows 24hrs in 60-min blocks always
- Added loading states while fetching remote events
- Updated `LoadingSpinner` and `EventPopup` styles and converted them from Tailwind to vanilla scoped CSS
- Updated `UserCalendarSync` to calculate the "time ago" properly, share a loading state with the calendar and bust cache to get the most up-to-date remote events
- Fixed a bug where availability slots in BookersPage would sometimes bleed over the availability time. We shouldn't show available slots if the slot duration go over the schedule's end time
- Updated styles for the slots with fixed colours. This was using the calendar colours before but Design team decided to do a fixed colour for now, for consistency.

> [!NOTE]
> This is quite a big change so I'd appreciate some time to test the calendar out both in the Dashboard and in the BookersView with various Availability and calendar setups 

--

[Explanation of why the issue happened and this PR's proposal on fixing it]:
- The core of the issue from #1442 was that we were forcibly tossing out events / slots in the calendar that would fall outside of the calendar grid.
  - IIRC this was done because we were only rendering a "minimum calendar size" meaning the span of hours that we would show was dependent on a combination of the availability hours set on the schedule, slot duration and the earliest events in it.
  - This clearly isn't ideal as we may end up with a combination of availability hours and slot duration that would potentially not render enough calendar grid cells to hold an event block.
  - So then the idea here is to always show the full 24hrs span of grid cells regardless of schedule and use 15 min grid cells then align every event with the closest 15 min grid cell. The 15 min choice is because that's the smallest duration option and the increments of duration can be divided by 15 min as well.
  - The calendar is now much taller than before which requires scroll. To reduce some of the friction here, there's a scroll function that gets the id of the closest time block to the users current time and scrolls it into view. This will also be useful in case we want to show a horizontal line for the current "now" time.

## Benefits

- Actually be able to see all events in a day
- A more accurate representation of time when manually syncing events
- Better feedback to the user on whether or not events are still being loaded

## Screenshots

Light mode (Dashboard)
image

Light mode (Booker page)
image

Dark mode (Dashboard)
image

Dark mode (Booker page)
image

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/1442

### DIFF
--- a/frontend/src/views/BookerView/index.vue
+++ b/frontend/src/views/BookerView/index.vue
@@ -193,7 +193,6 @@ export default {
 
 .booking-slot-selection-container {
   width: 100%;
-  max-width: 1440px;
   margin: 0 auto;
   user-select: none;
   padding: 0 1rem;
--- a/frontend/src/views/DashboardView/index.vue
+++ b/frontend/src/views/DashboardView/index.vue
@@ -133,8 +133,6 @@ export default {
     flex-direction: row;
     gap: 2rem;
     overflow-y: auto;
-    margin: 0 auto;
-    max-width: 1440px;    
   }
 
   .calendar-header-container {


frontend/src/views/DashboardView/components/WeekCalendar.vue
```diff
@@ -24,26 +24,70 @@ interface Props {
...
+// Scroll sync state
+let pendingScrollSync: number | null = null;
+let lastScrollSource: 'header' | 'body' | null = null;
+let isProgrammaticScroll = false;
+
+/**
+ * Syncs horizontal scroll between header and body for mobile
+ * Uses requestAnimationFrame to batch updates and prevent layout thrashing
+ */
+function syncScroll(source: 'header' | 'body') {
```
MelissaAutumn, line 61: When does this happen, because in mobiler simulator it doesn't seem to scroll to anything.

MelissaAutumn (commented):
Looks nice in action, I can't trigger the animation / scroll to stuff though.

For small mobile viewports I think we need to remove all left and right padding on the component.

frontend/src/views/DashboardView/components/WeekCalendar.vue
davinotdavid: oh that was an interesting problem to solve for 😅 this is not an automatic scroll, it is actually just to be able to horizontally scroll the weekday header and the body since they are separate components (the calendar is a grid but the weekday header is in a separate div altogether).

MelissaAutumn: Well you fixed it so well I thought nothing was actually fixed haha.

rwood-moz:
TBH, and just my opinion of course, but from a user point of view I liked it better when the calendar grid only shows the hours that you have marked as your availability. I find it seems like a lot of extra white space and a lot of scrolling required to see your availability hours (and I am concerned it will especially be too much on mobile).

davinotdavid:
> TBH, and just my opinion of course, but from a user point of view I liked it better when the calendar grid only shows the hours that you have marked as your availability. I find it seems like a lot of extra white space and a lot of scrolling required to see your availability hours (and I am concerned it will especially be too much on mobile).

@rwood-moz  Good point! I tried to mitigate this by scrolling within the calendar to the current time so hopefully the requirements to keep scrolling around would be reduced. The problem with the previous approach though is the reason the ticket was created in the first place (remote events that couldn't fit in the calendar were not displayed at all)... it is a trade off of erroring on side of completeness vs keeping the calendar as small as possible.

I've submitted a request with the design team for a review next week to go deeper into this and see what's the best approach. If it is the previous implementation, happy to revert and document the shortcomings somewhere for support!

frontend/src/views/DashboardView/components/WeekCalendar.vue
```diff
@@ -357,161 +327,249 @@ const filteredSelectableSlotsForGrid = computed(() => {
     }
   }).filter(Boolean)
 });
+
+/**
+ * Scrolls the calendar to the current time position
+ */
+function scrollToCurrentTime() {
```
devmount, line 341: Question: Is this meant to be scrolling to the exact position? Because on my end, it shows some previous events (it's currently 10.05 am in that screenshot):

Image

If this is intended, all good!

frontend/src/elements/LoadingSpinner.vue
devmount, line 1: Looks so much better now!! thanks

davinotdavid:
@rwood-moz @devmount 

I had a Design Review session with the Design Team and got feedback around 24hrs vs only availability hours. The decision was to always show 24hrs but instead of separating each hour in 4x 15-min blocks, do 1 block of 60-min instead and adjust the height of the events based on their duration.

In terms of having a "now bar" to showcase the current time in the calendar, it was decided not to be done for now as least for the community release. However, there will be some revisions to the calendar coming up as some edge cases were brought up (e.g. styles of events should follow a specific rotation of accessible colour palettes and overlapping calendar events handling, etc).

Since we have now 4 times less blocks, it might also ease a bit the scrolling to find a booker time!

I've since them applied (some) of the changes that were discussed, updated the PR description with some screenshots as well. Thank you so much for the review so far, I really appreciate it! May I ask for another review? 🙇

davinotdavid:
@devmount thanks for the review! sounds good to me, there's an extra open issue for design adjustments open that can host changes if need be here: https://github.com/thunderbird/appointment/issues/1465

Issue #1458, opened 2026-01-22
Allow multiple accounts to connect to the same calendar


davinotdavid:
**Describe the bug**
Currently there's an architectural restriction on having a calendar be connected to one account only. This shouldn't be a forced limitation anymore.

**To Reproduce**

- Go through the FTUE process and connect a google account A of your choice
- Logout
- Go through the FTUE process again with a different account and, in the connect calendar step, connect the same google account A
- See the error

PR #1463, opened 2026-01-26
Check for subscriber calendar ownership when creating calendar records

davinotdavid:
## Description of the Change

Before, we were only creating calendar records if it hasn't been created yet, otherwise we were just updating. The lookup for whether or not it had been created was only based on the `calendar_url` itself.

Bug flow:
- Subscriber A connects Google Account A (`google_url_a`) with X, Y and Z calendars. This succeeds since it is the first time that `google_url_a` is used.
- Subscriber B also connects Google Account A (`google_url_a`), except that when we try to create the calendar records, we see that `google_url_a` is already there, therefore we attempt to update the calendars instead.

Now we are also checking for calendar ownership so that calendar records are created if they don't exist for a subscriber (even if the `calendar_url` is the same).

## How to test

Prerequisites:
- If you are using `AUTH_SCHEME=oidc` you will need access to whichever Keycloak instance you are connecting to and have two users set up there!
- Easier to do if you remove all containers and volumes with a `docker compose down -v` and `docker compose up -d --build`
- Have a Google Account with calendars available

1. Go through the FTUE with the first user (A) and connect a google account
2. Logout
3. Go through the FTUE with the second user (B) and **connect the same google account used in step 1**
4. Check that you can still connect and there are calendars available

## Benefits

- Calendar records will be created for each subscriber based on their ownership (same Google Calendar in different subscribers will each have their own calendar since they have different 

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/1458

### DIFF
--- a/backend/src/appointment/database/repo/calendar.py
+++ b/backend/src/appointment/database/repo/calendar.py
@@ -134,7 +134,7 @@ def update_or_create(
     """update or create a subscriber calendar"""
     subscriber_calendar = get_by_url(db, calendar_url)
 
-    if subscriber_calendar is None:
+    if subscriber_calendar is None or subscriber_calendar.owner_id != subscriber_id:
         return create(db, calendar, subscriber_id, external_connection_id)
 
     return update_by_calendar(db, calendar, subscriber_calendar)
--- a/backend/test/integration/test_calendar.py
+++ b/backend/test/integration/test_calendar.py
@@ -4,7 +4,7 @@ import pytest
 
 from appointment.database.models import CalendarProvider
 from appointment.controller.calendar import CalDavConnector, GoogleConnector
-from appointment.database import schemas, models
+from appointment.database import schemas, models, repo
 from appointment.defines import GOOGLE_CALDAV_DOMAINS
 
 from sqlalchemy import select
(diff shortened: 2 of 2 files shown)


backend/src/appointment/database/repo/calendar.py
```diff
@@ -134,7 +134,7 @@ def update_or_create(
     """update or create a subscriber calendar"""
     subscriber_calendar = get_by_url(db, calendar_url)
 
-    if subscriber_calendar is None:
+    if subscriber_calendar is None or subscriber_calendar.owner_id != subscriber_id:
```
devmount, line 137: Just to be sure: We're still checking valid credentials beforehand right? So this wouldn't open a security hole?

davinotdavid: Ah yes, this code is on the repository level and the endpoints that make use of this method have the `get_subscriber` DI that checks for auth

devmount: That's what I thought, just wanted to double check. Thank you!

davinotdavid:
@devmount thanks for the review! You can also run with `AUTH_SCHEME=password` as well in that case as it shouldn't change the fix

devmount:
@davinotdavid Oh you're right, haha! I can test this tomorrow if you want, although I think this PR is ready to be merged in. It's not that big of a change.

davinotdavid:
@devmount All good! It is a small change indeed so I will merge this for now. Thanks!

Issue #1457, opened 2026-01-22
Better error messages for calendar connections in FTUE


davinotdavid:
Even though we can catch some exceptions in the backend around calendar connection errors, we are not surfacing good error messages to the user which makes them confused on how to fix connection issues.

For this task, it will be important to review which errors are we catching at the moment and which messages are being presented to the end user and make adjustments.

kewisch:
Adding a use case related to this: If you sign in to a caldav calendar that has no calendars exposed, it will happily show an empty dropdown. We should show the no calendars warning in this case as well.

Issue #1419, opened 2026-01-13
Broken CalDAV calendar causes infinite loading of Bookers page


devmount:
**Describe the bug**
I just encountered an issue with a bookers page showing a never ending loading spinner. This was caused by the backend not being able to process a link to CalDAV calendar which was returning invalid XML.

**To Reproduce**

1. Add a working CalDAV connection in the Appointment settings
2. Now modify the external CalDAV server to return invalid responses
3. Go to the bookers page
4. See never ending loading spinner

**Expected behavior**
The bookers page should show feedback to the user like "The calendar connection is currently not available". The backend should be able to handle corrupt external connections.

**Actual behavior**
The bookers page shows a never ending loading spinner.

**Additional context**
I encountered this error, since I had a Nextcloud CalDAV calender (for testing) connected to my Appointment account. Some months later, PHP the Nextcloud was running on got upgraded to a version, that wasn't compatible with Nextcloud anymore, hence the calendar requests returned plain text (saying something like Nextcloud isn't compatible).

But since the Appointment backend expected valid XML, it failed with error 500. See this example error log:

<details><summary>CRITICAL Expected some valid XML from the server</summary>
<code><pre>

2026-01-13 16:41:35.945 | 2026-01-13 15:41:35 CRITICAL Expected some valid XML from the server, but got this: 
2026-01-13 16:41:35.945 | b'This version of Nextcloud is not compatible with PHP>=8.4.<br/>You are currently running 8.4.16.'
2026-01-13 16:41:35.945 | Traceback (most recent call last):
2026-01-13 16:41:35.945 |   File "/usr/local/lib/python3.12/site-packages/caldav/davclient.py", line 120, in __init__
2026-01-13 16:41:35.945 |     self.tree = etree.XML(
2026-01-13 16:41:35.945 |                 ^^^^^^^^^^
2026-01-13 16:41:35.945 |   File "src/lxml/etree.pyx", line 3286, in lxml.etree.XML
2026-01-13 16:41:35.945 |   File "src/lxml/parser.pxi", line 1995, in lxml.etree._parseMemoryDocument
2026-01-13 16:41:35.945 |   File "src/lxml/parser.pxi", line 1882, in lxml.etree._parseDoc
2026-01-13 16:41:35.945 |   File "src/lxml/parser.pxi", line 1164, in lxml.etree._BaseParser._parseDoc
[...]

PR #1464, opened 2026-01-26
Better error messages during calendar connection

davinotdavid:
## Description of the Change

We are currently capturing some errors during calendar connection:

```
CalDAV

- GoogleCaldavNotSupported
  - "Google CalDAV is not supported. Please use the 'Connect Google Calendar' option"
- RemoteCalendarConnectionError (no auth support)
  - "We couldn't connect to your calendar. The server does not support any of the available authentication methods."
- RemoteCalendarConnectionError (no CalDAV)
  - "We couldn't connect to your calendar. The remote calendar does not support CalDAV."
- RemoteCalendarConnectionError (unauthorized)
  - "We couldn't connect to your calendar. Please check your username and password, then try again."
- TestConnectionFailed
  - Custom reason passed through, depends on reason.

Google

- access_denied / google-connect-to-continue
  - "You must connect at least one Google account to continue."
- GoogleScopeChanged / google-scope-changed
  - "You must enable Calendar and Event access to use Thunderbird Appointment."
- GoogleInvalidCredentials / google-invalid-creds
  - "Google authentication credentials are not valid."
- State mismatch/ google-auth-fail
  - "Google authentication failed."
- Sync failure / google-sync-fail
  - "An error occurred while syncing calendars. Please try again later."
```

However, we were only showing the title "We couldn't connect to your calendar" and not the description. Now we are showing that in the notice bar!

Also, we are now separating a little bit the `NotFoundError` and an invalid XML returned as their own error messages.

Screenshots of some examples:

Screenshot 2026-01-26 at 5 19 06 PM

Screenshot 2026-01-26 at 5 21 47 PM

- Our [`NoticeBar` component](https://thunderbird.github.io/services-ui/?path=/docs/components-noticebar--docs) could use some style changes for title / description as there's not a lot of hierarchy in typography.

## Benefits

- Better / more actionable error messages during calendar connection

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/1457

### DIFF
--- a/backend/src/appointment/l10n/de/main.ftl
+++ b/backend/src/appointment/l10n/de/main.ftl
@@ -54,7 +54,7 @@ connection-contains-default-calendar = Dieses Konto kann nicht getrennt werden, 
 remote-calendar-reason-doesnt-support-caldav = Der Kalender bietet keine CalDAV-Unterstützung.
 remote-calendar-reason-doesnt-support-auth = Der Kalender unterstützt keine der verfügbaren Authentifizierungsmethoden
 remote-calendar-reason-unauthorized = Bitte überprüfe Benutzername und Passwort und versuche es nochmal.
-remote-calendar-reason-not-found = Die Kalender-URL konnte nicht gefunden werden. Bitte überprüfe die URL und versuche es erneut.
+remote-calendar-reason-not-found = Der Kalender konnte nicht gefunden werden. Bitte überprüfe die URL und versuche es erneut.
 remote-calendar-reason-invalid-response = Der Server hat eine ungültige Antwort zurückgegeben. Bitte überprüfe die URL und versuche es erneut.
 
 event-could-not-be-accepted = Es ist ein Fehler bei der Annahme der Buchungsdaten aufgetreten. Bitte später noch einmal versuchen.


backend/src/appointment/l10n/en/main.ftl
```diff
@@ -54,6 +54,8 @@ connection-contains-default-calendar = Cannot disconnect this account because it
 remote-calendar-reason-doesnt-support-caldav = The remote calendar does not support CalDAV.
 remote-calendar-reason-doesnt-support-auth = The server does not support any of the available authentication methods.
 remote-calendar-reason-unauthorized = Please check your username and password, then try again.
+remote-calendar-reason-not-found = The calendar URL could not be found. Please check the URL and try again.
```
devmount: Technically, the URL was there but the resource it linked to could not be found. How about this?


```suggestion
remote-calendar-reason-not-found = The calendar could not be found. Please check the URL and try again.
```

backend/src/appointment/l10n/de/main.ftl
```diff
@@ -54,6 +54,8 @@ connection-contains-default-calendar = Dieses Konto kann nicht getrennt werden,
 remote-calendar-reason-doesnt-support-caldav = Der Kalender bietet keine CalDAV-Unterstützung.
 remote-calendar-reason-doesnt-support-auth = Der Kalender unterstützt keine der verfügbaren Authentifizierungsmethoden
 remote-calendar-reason-unauthorized = Bitte überprüfe Benutzername und Passwort und versuche es nochmal.
+remote-calendar-reason-not-found = Die Kalender-URL konnte nicht gefunden werden. Bitte überprüfe die URL und versuche es erneut.
```
devmount:
```suggestion
remote-calendar-reason-not-found = Der Kalender konnte nicht gefunden werden. Bitte überprüfe die URL und versuche es erneut.
```

backend/src/appointment/l10n/en/main.ftl
davinotdavid: Hmmm that makes sense to me, thanks!

davinotdavid:
@devmount yes! Related with https://github.com/thunderbird/appointment/issues/1419 but not fixing that just yet as this PR is only for the FTUE and the one in the issue is about the BookersPage (although it might be easier now to catch the exception on our end and show an error).

Issue #1467, opened 2026-01-27
FTUE create booking page error message missing field name


rwood-moz:
**Describe the bug**
FTUE `Booking Page Title` field validation error message is missing the field name.

**To Reproduce**

- Blow away your Appointment containers on your local stack to start clean
- Sign into Appointment and work through the FTUE until you get to the `Create booking page`
- In the `Booking Page Title` field enter > 128 characters
- Try to proceed in the FTUE

**Expected behavior**
Expect error message to appear indicating the `Booking Page Title` is invalid.

**Actual behavior**
Error appears but the field name is missing:

<img width="369" height="208" alt="Image" src="https://github.com/user-attachments/assets/fa8ed71a-1b1b-4707-b383-493cc31c8170" />

**System (please complete the following information):**
Firefox on Mac M3 Tahoe 26.2.

PR #1472, opened 2026-01-29
Fix field specific errors in schedule endpoints

davinotdavid:
## Description of the Change

- For Schedule related endpoints, we have a field specific error parser that extracts which field had an error and what the error was. These two parts are localized independently and we were missing a translation for "name" (which is the DB field name for "Booking Page Title").
- I've separated these keys into a sub-key "schedule" so that if we want to use the same pattern for other endpoints, we can have two "name" fields with different translations based on their context for example.

## Screenshots

Screenshot 2026-01-29 at 10 40 29 AM

## Benefits

- No more `fields.name` error!

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/1467

### DIFF
--- a/frontend/src/locales/de.json
+++ b/frontend/src/locales/de.json
@@ -41,9 +41,12 @@
     "weCouldntConnect": "Es konnte keine Verbindung hergestellt werden."
   },
   "fields": {
-    "details": "Seitenbeschreibung",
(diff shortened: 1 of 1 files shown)


devmount (approved):
Thanks Davi!

PR #1475, opened 2026-02-02
Fix error handling on booker page while fetching schedule

davinotdavid:
## Description of the Change

- There might be a case where the external calendar connection may be misconfigured or invalid outside of the realm of Appointment (e.g. App Passwords no longer valid or revoked in a CalDAV for instance). In those instances, we were just crashing out.
- This PR adds a try / except block to handle this gracefully and shows a more meaningful error message in such scenarios.

## How to test

Disclaimer: I've tested this with a Fastmail CalDAV as it was easier for me to enable / disable an App Password instead of creating a new one. Probably the shortest path to test would be:

- Delete all containers & volumes and re-create them to start fresh
- During FTUE, setup a CalDAV calendar of your choice (make sure that you have an App Password set up)
- After FTUE go to your CalDAV provider and either change the App Password to something else or disable / revoke it
- Access your Booking Page URL

## Screenshot

image

## Benefits

- No more infinite loading spinner in case there are remote calendar connection problems

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/1419

### DIFF
--- a/frontend/src/views/BookerView/components/BookingViewError.vue
+++ b/frontend/src/views/BookerView/components/BookingViewError.vue
@@ -1,5 +1,4 @@
 <script setup lang="ts">
-import ArtInvalidLink from '@/elements/arts/ArtInvalidLink.vue';
 import { useI18n } from 'vue-i18n';
 
 const { t } = useI18n();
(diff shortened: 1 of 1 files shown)


frontend/src/locales/de.json
```diff
@@ -33,6 +33,8 @@
     "noConnectedCalendarsLink": "Zu den Einstellungen wechseln.",
     "noSpecialCharacters": "Bitte die Sonderzeichen entfernen und nochmal versuchen.",
     "routeNotFound": "Entschuldigung, diese Seite wurde nicht gefunden.",
+    "scheduleRemoteCalendarConnectionError": "Es konnte keine Verbindung zum Kalender hergestellt werden.",
+    "scheduleRemoteCalendarConnectionDetails": "Bitte kontaktiere den Eigentümer der Buchungsseite direkt, um dieses Problem zu beheben.",
```
devmount, line 37: This is perfect 👌🏻

davinotdavid:
@rwood-moz @devmount Thank you so much for the review! Really appreciate it 🙏 🙇

 > The only thing worth mentioning is, that the backend still logs a critical error (the same I already put into the issue description at https://github.com/thunderbird/appointment/issues/1419).

I think this could potentially be useful, still? Not sure yet but it is an easy lift / cleanup if need be in the future so let's leave it for now.

Merging!

Issue #1483, opened 2026-02-05
Last Hour in a custom availability schedule does not allow scheduling


kellymcs01:
**Describe the bug**
 Investigate and fix the bug where custom availability settings cause issues with the last hour of time slots being bookable when using multiple separate blocks of time per day in custom availability settings.

<img width="1743" height="1109" alt="Image" src="https://github.com/user-attachments/assets/3565289d-f18c-4765-97a7-fe19b5c931c3" />

Recording: https://mozilla.hosted.panopto.com/Panopto/Pages/Viewer.aspx?id=db99105c-3502-44a8-b842-b3e7017f7119&start=5169.289686

PR #1488, opened 2026-02-05
Consider custom availabilities during booking time validation check

davinotdavid:
## Description of the Change

Even though we were showing the correct slots in the BookersPage, surprisingly we were not considering custom availabilities during the final check if the slot if available during booking (more specifically on `is_this_a_valid_booking_time`).

Before this PR, we had the impression that it was working probably because we were testing the custom availabilities slots in the BookersPage and attempting to book a time slot that was within the boundaries of the original schedule (which works fine).

Now, if the `use_custom_availabilities` is set to `True`, we are not comparing the select booking slot against the custom configured blocks of time.

## How to test

- Have a user ready / Go through FTUE
- Navigate to Availability and check "Set custom times for each day"
- Have at least one day of the week that has a block of time that goes beyond the original schedule from the FTUE
  - For example, if in the FTUE your schedule was created as Mon-Fri 9-5, one of the custom availability days should go from 5pm ~ 6pm (originally it went until 5pm and now one day goes until 6pm)
- Access the Booking Page URL
- Check that the slots displayed are correct in relation to what has been set in the Availability
- Attempt to book a slot that match the criteria mentioned in the example above (e.g. 5:30pm to 6:00pm in that scenario).
- Check that the booking works successfully

Bonus points: repeat the test in the `main` branch and check that the booking fails.

## Benefits

- While booking a slot, the backend will now check for the custom availability setting and validate against the blocks set.

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/1483

### DIFF
--- a/backend/src/appointment/database/repo/schedule.py
+++ b/backend/src/appointment/database/repo/schedule.py
@@ -4,7 +4,9 @@ Repository providing CRUD functions for schedule database models.
 """
 
 import uuid
+import zoneinfo
 
+from datetime import datetime, time, timezone
 from sqlalchemy.orm import Session
 from .. import models, schemas, repo
 from ... import utils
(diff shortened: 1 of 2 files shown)


backend/test/integration/test_schedule.py
```diff
@@ -1225,9 +1225,9 @@ def test_success_with_custom_availability(
     ):
         """Test that a user can request a booking from a schedule with custom availability slots"""
         start_date = date(2024, 4, 1)
-        start_time = time(9, tzinfo=UTC)
+        start_time = time(9)
```
davinotdavid, line 1228: Heads up, seems like we are theoretically storing the `start_time` and `end_time` as UTC by convention and not by adding the `tzinfo` itself in it. So I removed it since it was breaking tests.

davinotdavid:
@rwood-moz all good! Thanks for mentioning it! There were two problems here at play that have been fixed in the previous commits:

> Error notice bar not showing

We are only ever showing one notice bar at a time and the "unsaved changes" had the first priority. This is a problem because in the scenario that you pointed out, even though we had a network error / request error, the form is not reset to the previous values so it continues to be dirty and the "unsaved changes" would still show up.

Now I've changed to have the error notice bar be the first priority. However, in case there's an error, the user would need to dismiss / close the error in order to see the "unsaved changes" notice bar. IMHO that's ok since there's also another save changes button at the bottom regardless.

> Valid custom availabilities returns a backend error

This was a tricky one to spot. We are storing the times in UTC as a convention and then we pair with the timezone in the availability to do the necessary time conversions. Turns out that when we are saving a schedule / availability, we are sending the UTC times to the backend (which is ok by convention) except that the backend was trying to validate the schedule (e.g. start_time > end_time, etc) as if they were UTC values without considering timezone.

Now, right before we validate the availability, we are converting it back to local times temporarily (just for the validation) using the schedule's timezone. Added some tests for it too!

Would you review / try again? Also, aside from the problem you've found that was also happening in production, was the original ticket problem solved?

rwood-moz:
> Would you review / try again? Also, aside from the problem you've found that was also happening in production, was the original ticket problem solved?

@davinotdavid. Wow time zones are so complicated eh! Yes the original issue was solved, and the bug I found is also solved now too with your latest updates. Thanks for that!

PR #1494, opened 2026-02-06
Fix set-availability E2E test failures in BrowserStack

### LINKED ISSUE #1493: Set availability E2E test failures in BrowserStack
The set availability E2E test started failing on BrowserStack in production recently (however it seems to work fine when the same test is run against prod on my local machine; probably BrowserStack timing issues).

Update: For some reason BrowserStack started complaining about some `expect locator.toBeChecked` and `expect locator.toBeVisible` calls that they used to support. Changing to `locator.isChecked()` and `locator.isVisible()` and checking those values instead fix the issue.

rwood-moz:
Fix the set availability E2E test failures that are happening in BrowserStack. Fixes #1493.

When I tested this PR I:
- Ran the set availability E2E test in BrowserStack, links here for [Firefox](https://automate.browserstack.com/dashboard/v2/builds/e68e7700b88bbc1be8c5c46bb47704459ce02b83) and [Chromium](https://automate.browserstack.com/dashboard/v2/builds/20adbbf16c69833e823da4594384403557432ef8)
- Ran the set availability E2E test on local machine on prod, passed
- Ran the set availability E2E test on local machine against local stack, passed

### DIFF
--- a/test/e2e/tests/desktop/availability-set.spec.ts
+++ b/test/e2e/tests/desktop/availability-set.spec.ts
@@ -95,7 +95,6 @@ test.describe('set availability on desktop browser', {
       await availabilityPage.turnOffCustomizePerDayAndVerify();
     } else {
       // is on already, turn it off, verify daily time slot options aren't visible
-      const custom1vis = await availabilityPage.customStartTime1Input.isVisible();
       await availabilityPage.turnOffCustomizePerDayAndVerify();
       await availabilityPage.turnOnCustomizePerDayAndVerify();
     }


test/e2e/tests/desktop/availability-set.spec.ts
```diff
@@ -95,6 +95,7 @@ test.describe('set availability on desktop browser', {
       await availabilityPage.turnOffCustomizePerDayAndVerify();
     } else {
       // is on already, turn it off, verify daily time slot options aren't visible
+      const custom1vis = await availabilityPage.customStartTime1Input.isVisible();
```
davinotdavid: Nit: the `custom1vis` variable is declared but never used

rwood-moz: Ah yes I was dumping that out locally in logs before but missed removing that, thank you :)

rwood-moz:
> Approving since the tests pass but I wonder what would be the best practice here.
> 
> In which cases should be use `expect(await ..` vs `await expect(...` so we can adopt the pattern to the future and don't run into the timeout problems again?

Thanks for the review! Yeah it's tricky how they are implemented slightly different, playwright isn't consistent with their implementations.

The `toBeTruthy()` and `toBeFalsey()` methods don't return a promise, so that's why the expect (with no await); and the await inside the call because `isChecked()` and `isVisible()` do return promises. Whereas most other locator methods do return promises so that's why we have awaits before those expects.

PR #1501, opened 2026-02-10
Bump version to 1.0.0

davinotdavid:
In preparation for the Community Release and for better version control, let's make the current state v1.0.0 for both backend and frontend.

Last version update was 19 months ago 😱


devmount (commented):
Yes, thank you!

PR #1505, opened 2026-02-11
Update book appt E2E test to have auto-confirm bookings off

### LINKED ISSUE #1504: Book appointment E2E test fails when checking pending appointments link
The book appointment test is failing when checking the pending appointments link/notification (that appears on the main Appointment dashboard). This is because auto confirm booking was turned on, so after a new appointment was booked it wasn't a pending appointment; so the pending appointments link doesn't exist because all appointments are confirmed. Update the tests so they expect auto confirm appointments turned off (which is the default anyway).

rwood-moz:
Fixes #1504. Have the E2E tests run with the `auto confirm bookings` option turned off, so that after a booking request is sent, new pending bookings will exist and therefore the pending bookings notification/link will appear on the dashboard. Also turning on 1 test retry when tests run in CI (this may help in case a test fails because of slow running in BrowserStack).

BrowserStack link running all E2E tests on [Firefox desktop](https://automate.browserstack.com/dashboard/v2/builds/056afcf75227ec868754425424c9ef8f015c8c68) with this update (all tests pass).

### DIFF
--- a/test/e2e/pages/booking-page.ts
+++ b/test/e2e/pages/booking-page.ts
@@ -47,7 +47,7 @@ export class BookingPage {
     this.availableBookingSlot = this.page.locator('.selectable-slot', { hasNotText: 'Busy'});
     this.bookSelectionNameInput = this.page.getByPlaceholder('First and last name');
     this.bookSelectionEmailInput = this.page.getByPlaceholder('john.doe@example.com');
-    this.bookingConfirmedTitleText = this.page.getByText('Booking confirmed');
+    this.bookingConfirmedTitleText = this.page.getByText('Booking Request Sent');
     this.requestSentAvailabilityText = this.page.getByText("'s Availability");
     this.requestSentCloseBtn = this.page.getByRole('button', { name: 'Close' });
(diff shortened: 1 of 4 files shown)


test/e2e/pages/booking-page.ts
```diff
@@ -47,7 +47,7 @@ export class BookingPage {
     this.availableBookingSlot = this.page.locator('.selectable-slot', { hasNotText: 'Busy'});
     this.bookSelectionNameInput = this.page.getByPlaceholder('First and last name');
     this.bookSelectionEmailInput = this.page.getByPlaceholder('john.doe@example.com');
-    this.bookingConfirmedTitleText = this.page.getByText('Booking confirmed');
+    this.bookingConfirmedTitleText = this.page.getByText('Booking Request Sent');
```
davinotdavid, line 50: TIL that `getByText` is case insensitive by default unless we add `{ exact: true }`

rwood-moz: True, I'll just leave it though, we don't really need to verify the case.

PR #1507, opened 2026-02-11
WIP: E2E tests stability improvements

### LINKED ISSUE #1506: E2E test timezone improvements
The E2E tests need to run successfully in any geographical location because we run on local machines as well as in CI and BrowserStack. This makes the book an appointment test tricky.

The Appointment timezone setting may be set to `America/Toronto` however the bookee page might be using a different local timezone (BrowserStack workers usually are in `UTC`). The book appointment test needs to ensure an appointment was created for the selected timeslot using the correct time.

The E2E tests for desktop use `test.use` to set the browser timezone via playwright, so that the bookee page timezone is the same as the Appointment timezone setting. However this is not supported on mobile device browsers, and also doesn't always work consistently in BrowserStack, so I need a better solution.

Instead I will try this solution, the tests will:
- Always set the Appointment timezone setting to `America/Toronto` regardless of local browser timezone
- For the book appointment test:
    - The test will grab the timezone that is displayed on the bottom of the bookee page (this is the local browser timezone, might not match the Appointment timezone setting);
    - After a timeslot is selected, the test will ensure that the appointment requested confirmation dialog shows the correct timezone;
    - Then when the test signs into the Appointment dashboard to verify a new appointment was created, it will convert the selected timeslot date/time from the bookee page timezone into `America/Toronto` so that when the test checks the pending bookings list it can find the new pending appointment that matches the selected timeslot.

With this change it will also be easier to get the book an appointment test running on mobile browsers.

rwood-moz:
Changes to the E2E test to improve stability when run in CI/BrowserStack. Fixes #1506, see the issue for details.


rwood-moz:
Closing and moving to a new branch with less changes (PR was getting too large).

Issue #1195, opened 2025-08-25
In Settings, add external connections error status / error handling


davinotdavid:
**Is your feature request related to a problem? Please describe.**
Currently, we only have a "binary state" of the external connection that comes from the DB (whether or not it is connected). However, for troubleshooting purposes, it would be very helpful to know if there are any connection errors related to the external connection.

**Describe the solution you'd like**
Add a "health check" flow for an external connection and, in case of errors, show them as the wireframe below (note the Zoom error and the error badges throughout)

<img width="1280" height="1709" alt="Image" src="https://github.com/user-attachments/assets/275997f2-1222-43c1-8402-797f87e7eb5a" />

**Additional context**
Originally part of the https://github.com/thunderbird/appointment/issues/1096

davinotdavid:
Related with https://github.com/thunderbird/appointment/issues/1157 since it would be a good use case to showcase an error

malini:
oh this would be great! I'll add it in Early Bird so we can have some feedback for users. I'm happy to move forward with your suggestion for Early Bird, but please loop in Jesse so he can develop a hi-fi MVP version if it's not to design's preferences!

laurelterlesky:
Here's the update for error handling in settings -> navigation, sidebar nav and connected applications. https://zpl.io/RmoZAoE

Note: we'll need to update the navigation button components to have options for a leading or trailing icon (design will update the components next week)

davinotdavid:
Moving this to planning as we are still detailing the scope of the changes needed through design <-> dev conversations!

PR #1509, opened 2026-02-12
Surface external connection issues in Settings page

davinotdavid:
## Description of the Change

- New DB migration to add `status` and `status_checked_at` in the external connections table.
- Added new route `POST /external-connections/check-status` and controller + methods that checks Google, CalDAV and Zoom external connections individually.
- In the Frontend side we make the request above in two moments (more on that in the "things to improve"):
  - After login, as part of the `post-login` route
  - On the `onMounted` of the `Settings` page
- Added a warning icon to the Settings nav item and the aside in the Settings page in case there's any external connection problem.
- Implemented styles for the Settings page following the [recent Zeplin design changes](https://app.zeplin.io/project/69a863b267476ceecd9280de/screen/68cd9885ebfdb53d68206d7e). **Note that the error messages and the items in the dropdown menu change based on whether or not there's a default calendar in the external connection.**

## How to test

0. Since there's a new DB migration, the most straight forward would be to delete all containers and volumes and start from scratch again (`docker compose down -v` and `docker compose up -d --build`)
1. In the FTUE, connect a CalDAV server of your choice
2. After the FTUE is finished, go to your CalDAV server and change the App Password to something else
3. Either logout / login again or navigate to Settings
4. See the error being displayed

... or you can also go to `_check_google` / `_check_caldav` / `_check_zoom` in `backend/src/appointment/controller/external_connection.py` and return `ExternalConnectionStatus.error` forcibly to see the errors individually!

Also, please attempt to break the external connections (google, caldav, zoom) in other ways to test the logic 🙏 

## Screenshots

CalDAV connection error (not having a default calendar within it)

image

Google Account connection error while having a default calendar within it

image

Zoom connection error

image

## Benefits

- Subscribers are more aware of potential external connection issues and a clearer path to solve them.

- This PR is only regarding the Settings page. There's still a need to consider other parts of the application such as the Dashboard and Availability pages.
- Even though we are introducing the external connections health checks, this is only being called in two instances whereas it could be called in multiple different places. Also, the `status` and `status_checked_at` columns are only being updated during health checks but there are opportunities in the near future to manipulate those status in various `try / except` cases from the backend (e.g. when requesting for someone else's availability and we can't reach the calendar).

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/1195

### DIFF
--- a/backend/src/appointment/migrations/versions/2026_02_17_1534-4f6693fcc9d7_add_language_to_attendee.py
+++ b/backend/src/appointment/migrations/versions/2026_02_17_1534-4f6693fcc9d7_add_language_to_attendee.py
@@ -11,7 +11,7 @@ import sqlalchemy as sa
 
 # revision identifiers, used by Alembic.
 revision = '4f6693fcc9d7'
-down_revision = '89da5bf99c88'
+down_revision = 'd44b0832307c'
 branch_labels = None
 depends_on = None
 


backend/src/appointment/controller/external_connection.py
```diff
@@ -0,0 +1,129 @@
...
+        )
+        con.test_connection()
+
+        return ExternalConnectionStatus.ok
+    except (TestConnectionFailed, Exception) as e:
+        logging.warning(f'[health_check] CalDAV connection {ec.id} is unhealthy: {e}')
+        return ExternalConnectionStatus.error
+
+
+def check_status(db: Session, subscriber: Subscriber) -> list[ExternalConnections]:
```
MelissaAutumn, line 101: I haven't done a lot of asyncio work, but you could potentially shift this to a async function and potentially run the checks async together. 

There's obviously a limit of how much an event loop can run in one..loop, but it might be worth looking into.

MelissaAutumn: If you do make this async you'll have to make every other function that calls it async. If it's just a view function that calls it, then that'll be okay since fastapi says it automatically handles async view functions.

MelissaAutumn: Ah asyncio.gather should be similar to Promises.all afaik 
https://docs.python.org/3/library/asyncio-task.html#asyncio.gather

You'll basically just build a loop of the async tasks, and `results = await asyncio.gather(tasks, return_exceptions=True)` at the end. 

In order to support the update, you'll need to punt essentially the loop's contents into a new function. And that new function will be the task you gather up to run async.

MelissaAutumn: ...this can be punt to a new ticket btw. This is just a suggestion, lol.

backend/src/appointment/database/schemas.py
```diff
@@ -465,6 +465,8 @@ class ExternalConnectionOut(BaseModel):
     name: str
     type: str
     type_id: str
+    status: str = 'ok'
```
MelissaAutumn:
```suggestion
    status: str = ExternalConnectionStatus.ok.value
```

I'd optionally suggestion a INVALID or NONE status to default to. But that would work too.

frontend/src/views/SettingsView/components/ConnectedApplications.vue
```diff
@@ -131,19 +167,22 @@ function calendarConnected(calendarId: number) {
...
-              [calendarId]: value
-            }
-          }
+        const calendar = groupedCalendars.value.flatMap(g => g.calendars).find(c => c.id === calendarId);
+        const originalValue = calendar?.connected ?? false;
+        const { [calendarId]: _, ...rest } = currentState.value.changedCalendars ?? {};
+
+        // Using the function form of $patch which fully replaces the object instead of merging into it
+        // So that we can actually remove keys when the value is false
+        settingsStore.$patch((state) => {
```
MelissaAutumn: Could we punt this logic into a settingsStore function, like `replaceCalendars` or something along those lines?

frontend/src/views/SettingsView/components/ConnectedApplications.vue
```diff
@@ -177,6 +216,20 @@ function onCalendarColorChanged(event: HTMLInputElementEvent, calendarId: number
...
 }
 
+function reconnectExternalConnection(provider: ExternalConnectionProviders | CalendarProviders, isCalendar: boolean = false) {
+  if (isCalendar) {
+    if (provider === CalendarProviders.Google) {
+      connectGoogleCalendar();
+    } else if (provider === CalendarProviders.Caldav) {
+      connectCalDavModalOpen.value = true;
+    }
+  } else {
```
MelissaAutumn: If you `return;` here you can remove the else and a level of nesting.

frontend/src/stores/external-connections-store.ts
```diff
@@ -78,6 +86,23 @@ export const useExternalConnectionsStore = defineStore('externalConnections', ()
     // CalDAV is handled in the modal
   };
 
+  /**
+   * Run health checks on external connections (zoom, google, caldav).
+   * Updates the store with the latest connection statuses from the server.
+   */
+  const checkStatus = async () => {
```
MelissaAutumn: Can we cache this for a few minutes, with a force parameter to bypass cache? 

Right now we call this during sfc onMounted, which if someone is switching back and forth might be quite a few slow requests.

frontend/src/views/SettingsView/components/ConnectedApplications.vue
```diff
@@ -44,13 +45,15 @@ const calendarConnectedMap = new Map<number, ReturnType<typeof computed<boolean>
...
-// with some data from external connections to facilitate connection / disconnection.
-// It should be immutable as it represents the current state of the backend data.
-const initialCalendars = computed(() => {
-  // Find the default calendar's external_connection_id to determine which
-  // calendars share the same connection. Calendars sharing the default's
-  // connection cannot be disconnected without affecting the default calendar.
+const providerDisplayNames: Record<string, string> = {
+  caldav: 'CalDAV',
+  google: 'Google',
+};
```
MelissaAutumn: Do we have this somewhere else 🤔

frontend/src/views/SettingsView/components/ConnectedApplications.vue
```diff
@@ -62,23 +65,56 @@ const initialCalendars = computed(() => {
...
       is_default: currentState.value.defaultCalendarId === calendar.id,
       shares_default_connection: calendar.external_connection_id === defaultExternalConnectionId,
     }
   }) || [];
 
-  const sortedCalendars = formattedCalendars.sort((a, b) => a.title.localeCompare(b.title));
+  formattedCalendars.sort((a, b) => a.title.localeCompare(b.title));
+
+  const groups = new Map<number, {
+    connectionId: number;
```
MelissaAutumn: Sorry bit confused, how does this differ from formattedCalendars besides the key values? Could we just adjust formattedCalendars?

davinotdavid: Ah I had to come up with a "groupBy" for the displaying groups of calendars within a connection in the UI.

Since we didn't have that in the previous design, having a flat list of calendars (the `formattedCalendars`) was enough. This is current prod:

Screenshot 2026-03-05 at 12 53 55 PM

Now I have to show an "external connection header / row" first and then all the calendars from that  external connection.

frontend/src/views/SettingsView/components/ConnectedApplications.vue
davinotdavid: Seems not but I've moved this to `definitions.ts` as it might be useful in other places. Thanks!

frontend/src/views/SettingsView/components/ConnectedApplications.vue
MelissaAutumn: Ah okay, hmm let me think.

frontend/src/views/SettingsView/components/ConnectedApplications.vue
MelissaAutumn: I think if you swap this be a externalConnection map (so we'll get an item per externalConnection, instead of per calendar.) and move the group logic here instead that'll hit two Thunderbirds with one Roc as they say.

frontend/src/views/SettingsView/components/ConnectedApplications.vue
```diff
@@ -62,23 +65,56 @@ const initialCalendars = computed(() => {
...
-      provider_name: connectionProvider,
+      connection_status: externalConnection?.status,
+      provider_name: providerDisplayNames[String(connectionProvider)] ?? String(connectionProvider),
       is_default: currentState.value.defaultCalendarId === calendar.id,
       shares_default_connection: calendar.external_connection_id === defaultExternalConnectionId,
     }
   }) || [];
 
-  const sortedCalendars = formattedCalendars.sort((a, b) => a.title.localeCompare(b.title));
+  formattedCalendars.sort((a, b) => a.title.localeCompare(b.title));
```
MelissaAutumn: You'd need to move this in the new loop I mentioned above.

rwood-moz (commented):
Worked great when I tried it out locally. I had my google calendar connected and Appointment working great. Then went into my google account and disconnected Appointment from that side. When I went back to Appointment and navigated around / caused a refresh then the red action needed banner appeared at the top - and when I clicked that it took me to the google login to reconnect my account and it worked great.

Image

frontend/src/views/SettingsView/components/ConnectedApplications.vue
davinotdavid: Omg yes, going from external connection -> calendar makes a lot more sense, thanks for the idea! (In case other readers are wondering, no birds were harmed while proceeding with this idea)

frontend/src/stores/external-connections-store.ts
davinotdavid: Good point! Yes, added to the frontend! Thanks!

frontend/src/views/SettingsView/components/ConnectedApplications.vue
davinotdavid: Ah yes, good call. Much better and no need to reach for $patch, nice!

davinotdavid:
@rwood-moz thank you so much for the review! To be super honest, I was a bit shocked that I have never seen that "Action Needed" banner before 😅 I was only testing breaking the connection by returning an error forcibly on the code but I believe that banner was in place from before my time here. Nice!

About the Settings on the NavBar, that's a good discussion to have. Currently I am checking for the external connection status health in two occasions only:

- upon login
- when navigating to the Settings tab (now added a 5 min cache in between checks)

I think it would be a good improvement in another PR to mark the connection as broken whenever we are attempting to fetch remote event data (for example in the Dashboard or someone tries to access the booking page url).

backend/src/appointment/controller/external_connection.py
davinotdavid: That's a great suggestion, thanks! TIL about asyncio! I put this into another ticket just so I can have some time to learn / explore it (https://github.com/thunderbird/appointment/issues/1561)

rwood-moz:
> @rwood-moz thank you so much for the review! To be super honest, I was a bit shocked that I have never seen that "Action Needed" banner before 😅 I was only testing breaking the connection by returning an error forcibly on the code but I believe that banner was in place from before my time here. Nice!

haha Yes the red action banner looks good, but it only appears when the google connection is broken, not CalDAV. Or maybe it appears when the primary / default calendar is disconnected (which was my google one) which would make sense.

> I think it would be a good improvement in another PR to mark the connection as broken whenever we are attempting to fetch remote event data (for example in the Dashboard or someone tries to access the booking page url).

+1 :)

rwood-moz:
> haha Yes the red action banner looks good, but it only appears when the google connection is broken, not CalDAV. Or maybe it appears when the primary / default calendar is disconnected (which was my google one) which would make sense.

Update: The red action banner does only appear for google calendars, even if I only have a CalDAV calendar (so it is the default) and the connection becomes broken. Any chance the same red action banner can be used for CalDAV too?

backend/src/appointment/migrations/versions/2026_02_12_1503-d44b0832307c_add_status_to_external_.py
```diff
@@ -0,0 +1,37 @@
...
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+from appointment.database.models import ExternalConnectionStatus
+
+
+# revision identifiers, used by Alembic.
+revision = 'd44b0832307c'
```
devmount, line 15: This currently breaks the migration flow (since this PR is already a bit older and we got additional migration files already). We should make sure, that the down_revision of the next migration file `2026_02_17_1534-4f6693fcc9d7_add_language_to_attendee.py` is set to `d44b0832307c`.

Otherwise, we'd have no linear migration sequence when running `run-command main update-db`:

<details><summary>Multiple head revisions are present for given argument 'head'</summary>
<p>
<pre>
Database already initialized, running migrations
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/alembic/script/base.py", line 213, in _catch_revision_errors
    yield
  File "/usr/local/lib/python3.12/site-packages/alembic/script/base.py", line 421, in _upgrade_revs
    for script in reversed(list(revs))
                           ^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/alembic/script/revision.py", line 814, in iterate_revisions
    revisions, heads = fn(
                       ^^^
  File "/usr/local/lib/python3.12/site-packages/alembic/script/revision.py", line 1447, in _collect_upgrade_revisions
    for rev in self._parse_upgrade_target(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/alembic/script/revision.py", line 1234, in _parse_upgrade_target
    return self.get_revisions(target)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/alembic/script/revision.py", line 544, in get_revisions
    resolved_id, branch_label = self._resolve_revision_number(id_)
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/alembic/script/revision.py", line 778, in _resolve_revision_number
    current_head = self.get_current_head(branch_label)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/alembic/script/revision.py", line 508, in get_current_head
    raise MultipleHeads(
alembic.script.revision.MultipleHeads: Multiple heads are present for given argument 'head'; d44b0832307c, 17792ef315c1

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/bin/run-command", line 6, in <module>
    sys.exit(cli())
             ^^^^^
  File "/app/appointment/main.py", line 278, in cli
    app()
  File "/usr/local/lib/python3.12/site-packages/typer/main.py", line 1152, in __call__
    raise e
  File "/usr/local/lib/python3.12/site-packages/typer/main.py", line 1135, in __call__
    return get_command(self)(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/click/core.py", line 1485, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/typer/core.py", line 795, in main
    return _main(
           ^^^^^^
  File "/usr/local/lib/python3.12/site-packages/typer/core.py", line 188, in _main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/click/core.py", line 1873, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/click/core.py", line 1873, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/click/core.py", line 1269, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/click/core.py", line 824, in invoke
    return callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/typer/main.py", line 1514, in wrapper
    return callback(**use_params)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/app/appointment/routes/commands.py", line 31, in update_database
    update_db.run()
  File "/app/appointment/commands/update_db.py", line 50, in run
    command.upgrade(alembic_cfg, 'head')
  File "/usr/local/lib/python3.12/site-packages/alembic/command.py", line 483, in upgrade
    script.run_env()
  File "/usr/local/lib/python3.12/site-packages/alembic/script/base.py", line 545, in run_env
    util.load_python_file(self.dir, "env.py")
  File "/usr/local/lib/python3.12/site-packages/alembic/util/pyfiles.py", line 116, in load_python_file
    module = load_module_py(module_id, path)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/alembic/util/pyfiles.py", line 136, in load_module_py
    spec.loader.exec_module(module)  # type: ignore
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap_external>", line 999, in exec_module
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
  File "/app/appointment/migrations/env.py", line 86, in <module>
    run_migrations_online()
  File "/app/appointment/migrations/env.py", line 80, in run_migrations_online
    context.run_migrations()
  File "<string>", line 8, in run_migrations
  File "/usr/local/lib/python3.12/site-packages/alembic/runtime/environment.py", line 969, in run_migrations
    self.get_context().run_migrations(**kw)
  File "/usr/local/lib/python3.12/site-packages/alembic/runtime/migration.py", line 614, in run_migrations
    for step in self._migrations_fn(heads, self):
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/alembic/command.py", line 472, in upgrade
    return script._upgrade_revs(revision, rev)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/alembic/script/base.py", line 409, in _upgrade_revs
    with self._catch_revision_errors(
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/contextlib.py", line 158, in __exit__
    self.gen.throw(value)
  File "/usr/local/lib/python3.12/site-packages/alembic/script/base.py", line 239, in _catch_revision_errors
    raise util.CommandError(multiple_heads) from mh
alembic.util.exc.CommandError: Multiple head revisions are present for given argument 'head'; please specify a specific target revision, '<branchname>@head' to narrow to a specific head, or 'heads' for all heads
</pre>
</p>
</details>

frontend/src/locales/de.json
```diff
@@ -562,6 +563,9 @@
     },
     "settings": {
       "connectedApplications": {
+        "connectionUnreachableDefault": "Die Verbindung zu dem {providerName} Server konnte nicht hergestellt werden. Bitte verbinde es erneut oder setz einen anderen Kalender als Standard, um das Konto zu entfernen.",
+        "connectionUnreachableNonDefault": "Die Verbindung zu dem {providerName} Server konnte nicht hergestellt werden. Bitte verbinde es erneut oder entferne das Konto.",
+        "connectionUnreachableZoom": "Die Authentifizierung für die Verbindung mit der E-Mail-Adresse {email} ist fehlgeschlagen.",
```
devmount: Only minor adjustments here:


```suggestion
        "connectionUnreachableDefault": "Die Verbindung zum {providerName}-Server konnte nicht hergestellt werden. Bitte stell die Verbindung erneut her oder setz einen anderen Standardkalender, um das Konto zu entfernen.",
        "connectionUnreachableNonDefault": "Die Verbindung zum {providerName}-Server konnte nicht hergestellt werden. Bitte stell die Verbindung erneut her oder entferne das Konto.",
        "connectionUnreachableZoom": "Die Authentifizierung für die Verbindung mit der E-Mail-Adresse {email} ist fehlgeschlagen.",
```

backend/src/appointment/migrations/versions/2026_02_12_1503-d44b0832307c_add_status_to_external_.py
davinotdavid: Ah great point! First time running into this so far... perks of keeping old PRs around heh. Thanks for the explanation! Changed here!

davinotdavid:
@rwood-moz 

> Update: The red action banner does only appear for google calendars, even if I only have a CalDAV calendar (so it is the default) and the connection becomes broken. Any chance the same red action banner can be used for CalDAV too?

It is possible yes! Although I suspect that the Design team is not aware about this behaviour. I will raise this with them to see if it is better to copy the behaviour for CalDAV or to even remove the banner for Google.

Since extending this to potentially show multiple error banners for the CalDAV, Zoom, etc would require some implementation though, I've opened a new issue and mentioned it in the Design team channel: https://github.com/thunderbird/appointment/issues/1562

davinotdavid:
@devmount 

> One thing I noticed though (which might be an easy fix): When I have Appointment already open and navigate to the settings page, it still shows all external connections intact although they're not. I have to reload the page/app to see the error messages. I'm wondering, if we could add the check to the mounted or another suitable hook to update the settings page when accessing it without page reload?

Ah theoretically yes but that's a tricky balance though. Checking for the external connections health is quite expensive and I've added a 5 min cache per @MelissaAutumn review here: https://github.com/thunderbird/appointment/pull/1509#discussion_r2891375584 which is likely why navigating to other pages and back to Settings by clicking didn't trigger it. 

There's many places in the code that we are raising `RemoteCalendarConnectionError` / `REMOTE_CALENDAR_CONNECTION_ERROR` that could also flip the `status` of a specific connection to `error` instead of running all of the health checks but that would probably be added as we go since next steps here would be to adjust the UI from Dashboard and also send emails about it to the Subscriber.

Happy to take ideas on places to do the full health check and add it here though as I might have missed other good opportunities!

devmount:
> Ah theoretically yes but that's a tricky balance though. Checking for the external connections health is quite expensive and I've added a 5 min cache per @MelissaAutumn review here: [#1509 (comment)](https://github.com/thunderbird/appointment/pull/1509#discussion_r2891375584) which is likely why navigating to other pages and back to Settings by clicking didn't trigger it.
> 
> There's many places in the code that we are raising `RemoteCalendarConnectionError` / `REMOTE_CALENDAR_CONNECTION_ERROR` that could also flip the `status` of a specific connection to `error` instead of running all of the health checks but that would probably be added as we go since next steps here would be to adjust the UI from Dashboard and also send emails about it to the Subscriber.
> 
> Happy to take ideas on places to do the full health check and add it here though as I might have missed other good opportunities!

Ah you're right, doing all the checks is indeed expensive. In that case I agree, let's just keep this in mind when we expand on the connection errors. Thanks!

Issue #1495, opened 2026-02-09
Booking details should be in the language of the subscriber always


davinotdavid:
**Describe the bug**
The Subscriber had its language set to English within Appointment but someone booked an appointment with them and the appointment details is in German.

**To Reproduce**

1. As a Subscriber, have your Availability's calendar set to be a Google Calendar and also set your language to English in Settings if not already set
2. Copy your Booking Link
3. As a Bookee, open a different browser with the language in the browser set to German first
4. Navigate to the Booking Link and book a slot
5. As a Subscriber again, check the calendar invite details in Google Calendar / .ics file and see that they are in German

**Expected behavior**
If the Subscriber has its language set in Settings as English, the bookee's language shouldn't change the Google Calendar event / .ics event details.

davinotdavid:
Originally reported by @hellsworth (thanks!)

Note: steps to reproduce may differ as we are not entirely sure what happened during the time of the booking from the bookee's perspective. Regardless, we should ensure that the Subscriber's language is respected in the calendar event creation.

hellsworth:
When I talk with the booker this week, I'll try to gather more information to see if his browser was in German at the time of booking and maybe ask him to book another meeting as a test to see how the calendar event shows up (if it is in German again or English this time).

PR #1510, opened 2026-02-13
Respect Subscriber's language setting in Subscriber facing events & emails

davinotdavid:
## Description of the Change

- When we were internationalizing various strings in our email template using the `l10n` method in the backend, we were relying on the `accept-language` header of the request to determine which language to use. However, for subscriber facing events and emails this can be quite confusing as the bookee might use a different language than the subscriber.
- This PR passes a `lang` argument to the relevant email templates and enforce that the subscriber's language prevails in the event creation for subject / body including for Zoom-enabled appointments.
- Added a `language` column to the `attendee` table.
- Updated Bookee facing emails to respect the `attendee`'s language set on the time of booking.
- Removes the Support mailer and related files as they are not being used anymore.
- Adds time & timezone information for Pending, Rejected and Cancelled emails.
- Fix a problem with localization of date in emails (we were using Python's `strftime` without setting the locale but it's actually better to use Babel's `format_date` instead).

## How to test

This is a little convoluted to test through the UI but here's the best way I've found to reproduce the issue in `main` vs this branch:

0. Login / Go through FTUE in Appointment and set your language to `en` (make sure that the `tba/user` key in the local storage has a key called `settings.language = 'en'`.
1. Go to the Availability page and check `Automatically confirm bookings if time is available` as `true`.
2. Copy your Booking Page URL.
3. Open up a private window and change your browser's preferred language to be `German (de)`.
4. Access the Booking Page URL in the private window and make sure that it shows the UI in German.
5. Open up the Network tab and click on the various network requests and make sure that you see in the `Request Headers` that you have `accept-language: de` only.
6. Select any available slot and book it.
7. Open up `Mailpit` and check the emails that have been sent. The email sent to the subscriber should be in English and the email sent to the bookee should be in German.

Repeat the process with the `Automatically confirm bookings if time is available` as `false` and with a Zoom account configured and check that the `HOLD:` subject title and the `Join online at:` with the zoom link are in the subscriber's language (English in this case).

## Benefits

- Subscriber's language gets respected / less confusion

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/1495

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ cp appointment/frontend/.env.example appointment/frontend/.env
 cd appointment
 ```
 
-Next we need to pull Thunderbird Accounts. If you have a dev copy you can just symlink the folder to accounts. 
+Next we need to pull Thunderbird Accounts. If you have a dev copy you can just symlink the folder to accounts.
 
 Other-wise run:
 
@@ -48,14 +48,13 @@ docker-compose exec accounts uv run manage.py create_client 'Appointment' 'dev c
 
 You should see your Client ID and Client Secret within the output like:
 
-```
-
+```log
 Your client was successfully created with the uuid of f71cf674-228c-4558-b8b2-7780d6a36925
 Your Client Details:
 * Client ID: f71cf674-228c-4558-b8b2-7780d6a36925
 * Client Secret: a5303c654c839d4c8ae8aae7d3b866f581e75280d7f477ee43dcf2200939c6a12ea97fbceda916c50e1136e1615f6e4e523e7a23e2282092b0f88d91c3898b91
-
 ```
+
 (The above are just example values)
 
 Copy the Client ID and Client Secret values to your backend's .env file as `TB_ACCOUNTS_CLIENT_ID` and `TB_ACCOUNTS_SECRET` respectively.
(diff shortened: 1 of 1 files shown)


davinotdavid:
@devmount Great catch! You're correct that we were not able to determine the language for the bookee's confirmation event if it's not auto-confirm without storing it in the DB. Also, I agree that it would be best to show the slot's time and timezone in the bookee facing emails as well!

Here's the change log since the last time:

- Added a `language` column to the `attendee` table.
- Updated Bookee facing emails to respect the `attendee`'s language set on the time of booking.
- Removed the Support mailer and related files as they are not being used anymore.
- Adds time & timezone information for Pending, Rejected and Cancelled emails.
- Fixed a problem with localization of date in emails (we were using Python's `strftime` without setting the locale but it's actually better to use Babel's `format_date` instead).

Thank you SO much for reviewing and testing this thoroughly as I know this can be a bit tedious to try all combinations and emails 😅 

Would you mind giving it a full test again with all the combinations? 🙏 🙇

devmount:
@davinotdavid Wow, you're so fast! I gladly take a look at it again tomorrow 🙏🏻

backend/src/appointment/l10n/de/email.ftl
devmount, line 1: Thanks, this sounds good! It's nice to have the time slot data in the email now 🎉

devmount:
(🤫 since I had to nuke my local appointment docker and rebuild everything from scratch, I found a small issue in the README and sneaked the fix in here)

davinotdavid:
@devmount awesome! Thank you so much for the review + testing and the update on the docs! 🎉

Issue #1371, opened 2025-12-04
Availability settings are very slow to load on production


rwood-moz:
**Describe the bug**
The `Availability` panel the availability settings are very slow to load the first time or after a browser page refresh.

**To Reproduce**
- Sign into Appointment production
- After the dashboard appears, click the `Availability` tab at the top
- Watch the `You're bookable` toggle and other page settings

**Expected behavior**
Expect the availability settings to load immediately and the controls/fields to be set appropriately.

**Actual behavior**
It takes at least a few seconds for the availability settings to be loaded and the controls/fields updated on the page. For example, the `You're bookable` toggle is very noticeably slow to be set to the on position.

**System (please complete the following information):**
MacBook M3 Pro Tahoe 26.1. Happens on Firefox, Chrome, and Safari.

**Additional context**
I see this on Firefox, Chrome, and Safari so it's not browser-specific. Also happens if you are on the Availability page and you just do a browser page refresh, you'll see the slow loading. Brought to my attention by the E2E tests fail intermittently in production.

devmount:
Interesting, not happening on my local env. Can you quickly check the dev tools > network if there is a request happening that takes very long?

rwood-moz:
> Interesting, not happening on my local env. Can you quickly check the dev tools > network if there is a request happening that takes very long?

I am seeing the delay on production, maybe it's because of I am further away from where the app is hosted.

rwood-moz:
On production after navigating to `Availability` I am even able to edit the page name in the booking page details field and save the changes BEFORE the  page and the 'You're bookable' toggle setting has loaded; and if I do that then I get this error haha

<img width="1032" height="182" alt="Image" src="https://github.com/user-attachments/assets/a6a73c43-ad3d-4186-9c6f-547aa75607db" />

davinotdavid:
Tagged as hard since it would probably require not only investigations / profiling but also potential risky refactors.

devmount:
I'll investigate, let's see if I can find something.

rwood-moz:
Also I wonder if the eventual move to the K8s infra might help (maybe running on pods with more CPU or memory might help here)? cc @aatchison Although I don't know if it's an Appointment back-end performance issue itself.

devmount:
I reloaded the availability page in production and looked at the network request times. I can reproduce it in prod, not locally. Here are two screencasts showing how long it takes (sometimes, not always!) for the active toggle to get into it's activated state after a CTRL+F5:

[...]

PR #1513, opened 2026-02-13
E2E test stability improvements

### LINKED ISSUE #1506: E2E test timezone improvements

rwood-moz:
Making some changes to the E2E tests to make them more stable. Fixes #1506 (see the issue for more details).

All tests passed when run against my local stack on Firefox desktop. BrowserStack links running on [Android Chrome](https://automate.browserstack.com/dashboard/v2/builds/61a64572878b72e41024ade4a93a7c4695d1d7de) and [Firefox desktop](https://automate.browserstack.com/dashboard/v2/builds/db0999d86f686aaa787be3c2da53454dddc5c4a1).


test/e2e/pages/availability-page.ts
```diff
@@ -95,6 +94,12 @@ export class AvailabilityPage {
     // go to availability page, sometimes takes a bit to load all element values!
     await this.page.goto(APPT_AVAILABILITY_PAGE, { timeout: TIMEOUT_30_SECONDS });
     await this.page.waitForTimeout(TIMEOUT_10_SECONDS);
+
+    // the bookable toggle especially seems to take a long time to load in BrowserStack
+    if (! await this.bookableToggle.isEnabled()) {
+      await this.page.waitForTimeout(TIMEOUT_10_SECONDS);
+    }
```
devmount, line 101: Thanks for the workaround. This is interesting though, should we investigate why this happens? Are there other control elements that take longer a usual to load? Or this this a BrowserStack issue?

rwood-moz: > Thanks for the workaround. This is interesting though, should we investigate why this happens? Are there other control elements that take longer a usual to load? Or this this a BrowserStack issue?

Thanks for the review @devmount. It is a production issue, that is seen on production but worse in BrowserStack since it is even slower. Actually I previously filed #1371 concerning the slow loading of the 'you're bookable' toggle / availability page.

devmount: Ah I knew this issue sounded familiar 😇  Thanks for the ticket, at some point we should take a look at that.

Issue #1519, opened 2026-02-18
Calendar events do not show in the Appointment dashboard calendar view


lisajill:
**Describe the bug**

My Google Calendar events, created outside of Appointment, do not show in the Appointment dashboard.

**To Reproduce**

Connect Google Calendar (Google Workspace)
View the Appointment dashboard calendar view: https://appointment.tb.pro/dashboard

**Expected behavior**

See my Google Calendar event titles in their correct date & time slot.

**Actual behavior**

Calendar is empty.

<img width="2150" height="1402" alt="Image" src="https://github.com/user-attachments/assets/f8335a60-3428-4b08-bbe9-91c06bf806c7" />

**Screenshots**

Settings:

<img width="1114" height="370" alt="Image" src="https://github.com/user-attachments/assets/66ae2b26-f73f-43ac-b079-0f67cbff17ab" />

**System (please complete the following information):**

- OS: MacOS Tahoe 26.2
- Browser Version: Firefox 147.0.3 (aarch64)

davinotdavid:
Sentry error:

https://thunderbird.sentry.io/issues/7195606736/events/recommended/?project=4505428124827648&query=user.id%3A%222%22&referrer=issue-stream

PR #1520, opened 2026-02-18
Manually paginate Google remote events for accurate query params

davinotdavid:
## Description of the Change

We were using the `list_next()` method from the `GoogleApiClient` to ease the pagination process when reading remote events which is nice except that it apparently attempts to re-encode the URI with the previously passed query params and fails its own validation when trying for duplicated query params during the process.

There are duplicated query params in our case since we are querying for different `eventTypes` and [event types are passed as repeated params per their own API](https://developers.google.com/workspace/calendar/api/v3/reference/events/list).

This problem only happens when the user has enough events to trigger a pagination (one more iteration of the existing loop). Then, in our case, we unfortunately need to handle the pagination manually.

Sentry error:
https://thunderbird.sentry.io/issues/7195606736/events/recommended/?project=4505428124827648&query=user.id%3A%222%22&referrer=issue-stream

Exact line that triggers this in the GoogleApiClient:
https://github.com/googleapis/google-api-python-client/blob/main/googleapiclient/_helpers.py#L157

## How to test

0. In the `main` branch (not this PR's branch), update [this method call](https://github.com/thunderbird/appointment/blob/main/backend/src/appointment/controller/apis/google_client.py#L205) with an extra parameter `maxResults=1`.
1. Go through FTUE / Have an account ready with a Google Account attached to it with a calendar that has at least 2 events in it.
2. **See that no events show up (which is the original bug).**
3. Change branches to this PR's branch.
4. Login again with the same account.
5. **See that it now loads any remote events you might have in the Google Account.**

## Benefits

Folks with lots of events that require pagination would be able to see their remote events in the calendar!

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/1519

### DIFF
--- a/backend/src/appointment/controller/apis/google_client.py
+++ b/backend/src/appointment/controller/apis/google_client.py
@@ -204,15 +204,34 @@ class GoogleClient:
                 eventTypes=event_types,
                 fields=fields,
             )
+
             while request is not None:
                 try:
                     response = request.execute()
-
                     items += response.get('items', [])
                 except HttpError as e:
                     logging.warning(f'[google_client.list_events] Request Error: {e.status_code}/{e.error_details}')
+                    break
 
-                request = service.events().list_next(request, response)
+                # We need to manually paginate because the Google API's list_next() method
+                # corrupts repeated query params like eventTypes on re-encoding.
+                # See https://github.com/googleapis/google-api-python-client/blob/main/googleapiclient/discovery.py#L1382-L1384
+                # and https://github.com/googleapis/google-api-python-client/blob/main/googleapiclient/_helpers.py#L141-L163
+                page_token = response.get('nextPageToken')
+
+                if page_token:
+                    request = service.events().list(
+                        calendarId=calendar_id,
+                        timeMin=time_min,
+                        timeMax=time_max,
+                        singleEvents=True,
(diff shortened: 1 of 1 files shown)


devmount:
@davinotdavid Question for testing: How can I use the `maxResults=1` in the main branch? Is this even possible or do I have to create lots of calendar events to reproduce this?

davinotdavid:
@devmount oh my bad, my test instructions are a bit confusing this time. The bug is because of the `list_next` which auto-paginates, so instead of creating lots of events in the calendar to reproduce this issue, you can add the `maxResults=1` to the code in the `main` branch as an additional parameter in this method call:

https://github.com/thunderbird/appointment/blob/main/backend/src/appointment/controller/apis/google_client.py#L205

Then, it will force the pagination (as long as you have at least 2 calendar events, that is) and you can see the error in the `main` branch!

Updated the PR description in case you want to give it a try!

devmount:
@davinotdavid Oh haha yes, I was just looking at the wrong line 😂  I can confirm the bug as well as the fix. Thanks so much! 🎉

davinotdavid:
@devmount awesome! Thank you so much for reviewing & testing! 🎉

PR #1525, opened 2026-02-20
Fix intermittent E2E test failures

### LINKED ISSUE #1524: Fix intermittent E2E test failures
A couple more intermittent E2E test failures to be fixed:

The Appointment production sanity test is failing because the `access-booking-page` test is failing (has been for awhile). BrowserStack link [here](https://automate.browserstack.com/dashboard/v2/builds/8d728fa451a001313eb2cafa328a58895098c1a4).

Also the nightly E2E tests on Win11 Chromium are failing intermittently on authentication because the timezone field is taking an extremely long time to populate; BrowserStack link [here](https://automate.browserstack.com/dashboard/v2/builds/fc12625e39765b6fb5f21895d3bf47d7f8799672).

rwood-moz:
A couple more intermittent E2E test failure fixes. Fixes #1524. More specifically:

- Fix the production sanity test: The access-booking-page test has been failing for quite some time; it was checking that an element on the bookee page was hidden, however the element is actually not visible but just not using the actual hidden element value. Just removing that check, not really necessary anyway
- On Win11 Chrome the tests were failing intermittently because of an issue with the timezone locator (during the auth setup) that only seems to happen on Win11
- Removing the connected applications `verify calendar checkbox change shows notice bar and revert restores previous state` test because it has been flaky and it's not a flow that really needs to be repeated
- Remove a couple of unused imports

Tested this branch on BrowserStack:
- Prod sanity test [now passes](https://automate.browserstack.com/dashboard/v2/builds/f1708c6d737bd08ba75d0a13bade0447744d600d)
- Nightly E2E tests pass on [Win11 Chromium](https://automate.browserstack.com/dashboard/v2/builds/0c14bc17acf1470604ba8897e3771972277ea90d), [Firefox on MacOS](https://automate.browserstack.com/dashboard/v2/builds/2dccb5a3f45701134ad881ab62a784dcfeebe681), and [Safari on MacOS](https://automate.browserstack.com/dashboard/v2/builds/a93c21c99dcf079dcc24ae6632ca6849763a3a9d)

### DIFF
--- a/test/e2e/tests/desktop/settings-connected-apps.spec.ts
+++ b/test/e2e/tests/desktop/settings-connected-apps.spec.ts
@@ -9,6 +9,7 @@ import {
   TIMEOUT_1_SECOND,
   TIMEOUT_30_SECONDS,
  } from '../../const/constants';
+import { setSeconds } from 'date-fns/setSeconds';
 
 let settingsPage: SettingsPage;
 let dashboardPage: DashboardPage;
@@ -34,6 +35,7 @@ test.describe('connected applications settings on desktop browser', {
     await settingsPage.defaultCalendarBadge.scrollIntoViewIfNeeded();
 
     // verify that clicking the 'add caldav' button brings up the caldav connection dialog; just close it
+    await settingsPage.addCaldavBtn.waitFor({ state: 'visible', timeout: TIMEOUT_30_SECONDS });
     await settingsPage.addCaldavBtn.scrollIntoViewIfNeeded();
     await settingsPage.addCaldavBtn.click();
     await page.waitForTimeout(TIMEOUT_1_SECOND);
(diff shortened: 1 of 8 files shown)


test/e2e/tests/desktop/settings-connected-apps.spec.ts
```diff
@@ -81,52 +83,4 @@ test.describe('connected applications settings on desktop browser', {
       await page.waitForTimeout(TIMEOUT_1_SECOND);
     }
   });
-
-  test('verify calendar checkbox change shows notice bar and revert restores previous state', async ({ page }) => {
```
davinotdavid, line 85: ~~Should we skip this test instead of fully removing? Is there a plan to bring it back?~~

Ops, never mind, just saw the description item there that we don't need to repeat the test.

davinotdavid (approved):
lgtm! thanks! one question though!

test/e2e/tests/desktop/settings-connected-apps.spec.ts
rwood-moz: Thanks Davi. Yes IMO this doesn't really need to be made into an E2E test as it's not really a common flow that will be exercised multiple times (and it just adds complexity to the test suite). The other tests also already verify saving and reverting changes.

Issue #82, opened 2023-05-26
Implement worker queue system


MelissaAutumn:
**Is your feature request related to a problem? Please describe.**
We're starting to implement tasks that should not block requests. e.g. Sending emails, download data requests, etc..

Doing tasks in the background is a great way to avoid waiting!

We currently use fastapi's background tasks but that does hold up the worker until the task is complete, which can potentially block future connections if all workers are busy doing things. 

**Describe the solution you'd like**
https://docs.celeryq.dev/en/stable/ via redis

**Additional context**
For any UI interactions we might need to shoot off an email once the job is done, or some kind of notification system...maybe just email though.

davinotdavid:
Bumping this to Earlybird for better visibility as it would be nice to work on this before more folks come in.

Not sure if the Milestone itself is relevant here though.

malini:
From discussions with @davinotdavid, this is good to have for renewing the google sync token, and for sending emails from fastAPI

PR #1529, opened 2026-02-25
Use Google Calendar API calendar invites / Implement Push Notifications

### LINKED ISSUE #80: Implement Google's push notifications API
This is a more involved project, but to reduce our google calendar api we can request to be notified of calendar updates (for CalendarList and Event endpoints.) We can then combine it with a slightly modified https://github.com/thundernest/appointment/issues/25 to accurately store events on our server. 

https://developers.google.com/calendar/api/guides/push

### LINKED ISSUE #1393: Confirmed Events not showing up on other person's calendar
**Describe the bug**
Twice now, someone has scheduled time with me and the event shows they are invited and attending - but they don't show. They say they never received an invite.

In both cases they are on Google Workspace accounts.

**To Reproduce**

(I have my account set to require confirmation)

1. Invite someone
2. Event doesn't show up for them (or they don't receive an invite)

**Expected behavior**
An invite should be sent to the other party

**Actual behavior**
Event doesn't show up for them (or they don't receive an invite)

davinotdavid:
## Description of the Change

TL;DR: If the default calendar (the one selected in the Availability) is a calendar attached to a Google Account, we are now using Google Calendar's email flow instead of our own. We retrofit the event decisions into Appointment through [Push Notifications](https://developers.google.com/workspace/calendar/api/guides/push) into a webhook.

Shouldn't be super dangerous as there's an env var (`GOOGLE_INVITE_ENABLED`) that functions as a feature flag that is `False` in prod and `True` in non-prod environments.

PS: Sorry for the super long PR, lots of cases to cover

--

Whenever a Subscriber has a google account / google calendar as their default calendar, we are passing the control of the invitation emails to Google through their Events API. This is done through using `insert` instead of `import` while adding the event to the remote calendar.

Not having our own invitation emails means that we lose the visibility on the Confirm / Deny / Cancel options so in order to solve for it, this PR implements their Push Notifications workflow that work as follows:

```mermaid
sequenceDiagram
    participant Bookee
    participant App
    participant GoogleCal as Google Calendar
    participant Subscriber

    Bookee->>App: Request time slot
    App->>GoogleCal: "insert() with bookee as attendee, status=tentative"
    GoogleCal-->>Bookee: Google Calendar invite (tentative)
    App-->>Subscriber: Branded confirmation email

    alt Subscriber confirms (email/UI)
        Subscriber->>App: Accept
        App->>GoogleCal: "patch() status=confirmed, sendUpdates=all"
        GoogleCal-->>Bookee: Update notification (confirmed)
        App->>App: Book slot, update DB status
    else Subscriber denies (email/UI)
        Subscriber->>App: Deny
        App->>GoogleCal: "delete() sendUpdates=all"
        GoogleCal-->>Bookee: Cancellation email
        App->>App: Mark slot declined
    else Bookee declines (Google Calendar)
        Bookee->>GoogleCal: Decline invite
        GoogleCal-->>Subscriber: Decline notification email
        GoogleCal->>App: Watch webhook
        App->>App: Mark slot declined, delete event
    else Bookee accepts (Google Calendar)
        Bookee->>GoogleCal: Accept invite
        GoogleCal->>App: Watch webhook
        Note over App: No action - subscriber must still confirm
    end
```

The above is a partial graph but consider that the Subscriber's actions through the Google Calendar Event would also backtrack to our application! **Also, there's no HOLD event for this flow anymore**.

Here's some specific use-cases to consider:

- FTUE use-case

For users that opt to connect a google account / google calendar we now create an entry in the `google_calendar_channels` table so that we have a sync token to compare with when the webhook is triggered with updates.

- Existing users that have a Google Account / Calendar as the default calendar (post-FTUE)

Added a one-off command that can be ran to back-fill the watch channels required for the push notifications to work if the user has the default calendar as a Google Calendar and it is currently connected.

- Existing users that have a Google Account / Calendar but it is currently disconnected in Settings and not default (post-FTUE)

When a calendar is connected through the Settings, if the external connection is a google connection, we are still not creating the watch channel. We only do so if the calendar is set as default. Likewise, if the calendar is not the default for a schedule anymore, we teardown the watch channel. 

- Expired sync tokens

There is a command to renew the sync tokens for the channels that are in the `google_calendar_channels` table. However, we don't have a task queue setup yet for this so this has to be ran manually for now.

## How to test

> [!NOTE]
> Unfortunately, the Push Notifications API / Google Webhook require a publicly available URL with HTTPS. The most seamlessly way that I could find was to use [Cloudflare Tunnels](https://developers.cloudflare.com/cloudflare-one/networks/connectors/cloudflare-tunnel/) pointing to a personal domain that I already own and routing that to localhost:5000.

0. Assuming that you already have a publicly exposed HTTPS URL, take down all your containers and bring them up again `docker compose down -v && docker compose up -d --build` and before you do anything else, update the `get_webhook_url`'s backend_url to be the publicly exposed HTTPS URL (backend/src/appointment/controller/google_watch.py).
1. Go through the FTUE and connect a Google Account / Google Calendar
2. Copy & open your Booking Page URL and book any events
3. Check that the Subscriber receives our own branded email
4. Check that the Bookee receives the Google Calendar invite email
5. As a Subscriber, go to Google Calendar and accept / decline the invite
6. As a Bookee, check that you've received the Google Calendar email notification about the change of status of the event
7. Go to the Bookings page and check that the status of the event matches what has been answered

Some other cases to test:
- "Automatically confirm bookings if time is available" On / Off should behave as expected
- Declining the Google Calendar invite from the Bookee perspective through the Google Calendar UI / email RSVP  should also decline / cancel the event for the Subscriber

## Benefits

- For Google Account users, this will reduce the amount of no shows since we would be able to reliably insert events into users' Google Calendars. The original problem was that since we were silently adding events through the `import` method (which doesn't send Google emails) instead of `insert` (which sends Google emails) and using our own emails for notification, unless the recipient had already interacted with no-reply@appointment.tb.pro or manually clicked in the "Add to calendar" button in their email, they wouldn't have the event added to their calendar automatically.

- There are more actions that can be done through the Google Calendar API that are currently not mapped to features we have in Appointment like the 'Maybe' option or the 'Change time' / 'Propose new time'.
- ~~We need to renew the watch channels periodically but we don't have yet a task queue setup. We had a brief discussion on using Celery but instead of having a new infrastructure for it inside Appointment, we can explore re-using the existing on in Accounts (TBD though). To mitigate this I can manually add to my calendar to run the commands :/~~ Now we have Celery! 🎉 

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/1393
Fixes https://github.com/thunderbird/appointment/issues/80


backend/src/appointment/defines.py
```diff
@@ -35,6 +35,16 @@
...
 
 
+def google_invite_enabled() -> bool:
+    """Check if Google Calendar native invites are enabled.
+    Explicit env var takes precedence; otherwise enabled for all envs except prod."""
+    explicit = os.getenv('GOOGLE_INVITE_ENABLED')
+    if explicit is not None:
+        return explicit.lower() in ('true', '1', 'yes')
+    app_env = os.getenv('APP_ENV', APP_ENV_DEV)
+    return app_env != APP_ENV_PROD
```
MelissaAutumn: Why do we need the prod check? We should just check the GOOGLE_INVITE_ENABLED.

backend/src/appointment/routes/webhooks.py
```diff
@@ -110,3 +119,317 @@ def zoom_deauthorization(
...
+    Google sends a POST here whenever events change on a watched calendar."""
+    channel_id = request.headers.get('X-Goog-Channel-Id')
+    resource_state = request.headers.get('X-Goog-Resource-State')
+
+    if not channel_id:
+        return Response(status_code=200)
+
+    # Google sends a 'sync' notification when the channel is first created; just acknowledge it
+    if resource_state == 'sync':
+        return Response(status_code=200)
```
MelissaAutumn: You can combine these two returns. Also mind forming the 200 response into a variable and returning that?

backend/src/appointment/routes/webhooks.py
```diff
@@ -110,3 +119,317 @@ def zoom_deauthorization(
...
+
+    token = Credentials.from_authorized_user_info(
+        json.loads(external_connection.token), google_client.SCOPES
+    )
+
+    if not channel.sync_token:
+        sync_token = google_client.get_initial_sync_token(calendar.user, token)
+        if sync_token:
+            repo.google_calendar_channel.update_sync_token(db, channel, sync_token)
```
MelissaAutumn: Can we roll with the newly updated sync token or should this return 200? (because we have to wait for a new event?)

backend/src/appointment/routes/webhooks.py
```diff
@@ -110,3 +119,317 @@ def zoom_deauthorization(
...
+        repo.google_calendar_channel.update_sync_token(db, channel, new_sync_token)
+
+    background_tasks.add_task(
+        _process_google_event_changes,
+        calendar_id=calendar.id,
+        changed_events=changed_events,
+        google_client=google_client,
+        google_token=token,
+        remote_calendar_id=calendar.user,
+    )
```
MelissaAutumn: We might want to implement celery support before we dump this on stage/prod. I have a feeling this will hang up request workers.

backend/src/appointment/routes/webhooks.py
```diff
@@ -110,3 +119,317 @@ def zoom_deauthorization(
...
+    slot_update = schemas.SlotUpdate(booking_status=models.BookingStatus.cancelled)
+    repo.slot.update(db, slot.id, slot_update)
+
+    logging.info(
+        f'[webhooks.google_calendar] Event cancelled for appointment {appointment.id}, '
+        f'slot {slot.id} marked as cancelled'
+    )
+
+
+def _create_zoom_meeting_link(
```
MelissaAutumn: Can we this code elsewhere, perhaps replacing the inlined caldav code for creating a zoom link (iirc)

MelissaAutumn (changes requested):
WIP review, but firefox keeps crashing on me. 🤷 

Sorry for the delay on this!

backend/src/appointment/routes/webhooks.py
davinotdavid: We could but it would probably return no new events since the token would represent everything up to this moment. The call afterwards to `list_events_sync` would probably return nothing but there's a small change that something may happen in between so I will remove the early return and roll with it!

Also, this is more of a rescue / fallback scenario since the `sync_token` should have been created during the channel creation so this is more like a "if we received a webhook call but somehow failed to create a sync token to ask for events" kind of thing.

backend/src/appointment/defines.py
davinotdavid: Hmm yeah, we don't need that. Checking GOOGLE_INVITE_ENABLED directly, thanks!

backend/src/appointment/routes/webhooks.py
davinotdavid: Combined! Thanks!

backend/src/appointment/routes/webhooks.py
davinotdavid: Yes, great catch / memory! Moved this to the zoom controller and using in both places where we create zoom links now. Thanks!

backend/src/appointment/routes/webhooks.py
davinotdavid: Yep! This is scoped for Early Bird here https://github.com/thunderbird/appointment/issues/82 and the flag is only `True` in stage so we can check to see if this will hang up request workers

backend/src/appointment/commands/backfill_google_channels.py
```diff
@@ -0,0 +1,99 @@
...
+        print('BACKEND_URL not set, aborting.')
+        db.close()
+        return
+
+    # Find connected Google calendars that are the default in a schedule
+    # and don't yet have a watch channel
+    schedule_calendar_ids = {
+        s.calendar_id
+        for s in db.query(models.Schedule).all()
+        if s.calendar_id is not None
```
MelissaAutumn: Might want to reshape this to

```
schedule_calendar_ids = db.query(models.Schedule).filter(models.Schedule.calendar_id == None).all()
```

backend/src/appointment/controller/apis/google_client.py
```diff
@@ -246,17 +262,139 @@ def save_event(self, calendar_id, body, token):
...
+        with build('calendar', 'v3', credentials=token, cache_discovery=False) as service:
+            try:
+                response = service.events().watch(
+                    calendarId=calendar_id,
+                    body={
+                        'id': channel_id,
+                        'type': 'web_hook',
+                        'address': webhook_url,
+                    },
+                ).execute()
```
MelissaAutumn, line 353: Looking at the docs, we should probably generate a unique `token` (another uuid), store it alongside the channel (encrypted) and check it on every notification. We can call it `state` to avoid confusing it with the google token.

We can punt that to another PR though.

MelissaAutumn: We should also set the ttl as a value from our env. That way if they ever change the default we hopefully won't have to change anything.

backend/src/appointment/controller/google_watch.py
```diff
@@ -0,0 +1,126 @@
...
+    """Build the Google Calendar webhook URL from the backend URL, requires https."""
+    backend_url = os.getenv('BACKEND_URL')
+    if not backend_url:
+        return None
+    return f'{backend_url}/webhooks/google-calendar'
+
+
+def get_google_token(google_client: GoogleClient, external_connection: models.ExternalConnections):
+    """Build Google Credentials from an external connection's stored token."""
+    return Credentials.from_authorized_user_info(
```
MelissaAutumn, line 30: Can you check for external_connection.token truthyness?

backend/src/appointment/controller/google_watch.py
```diff
@@ -0,0 +1,126 @@
...
+
+    external_connection = calendar.external_connection
+    if not external_connection or not external_connection.token:
+        return
+
+    try:
+        token = get_google_token(google_client, external_connection)
+    except (json.JSONDecodeError, Exception) as e:
+        logging.warning(f'[google_watch] Could not parse token for calendar {calendar.id}: {e}')
+        return
```
MelissaAutumn: This should probably be flagged as an error and we should return None or False

backend/src/appointment/controller/google_watch.py
```diff
@@ -0,0 +1,126 @@
...
+            repo.google_calendar_channel.create(
+                db,
+                calendar_id=calendar.id,
+                channel_id=response['id'],
+                resource_id=response['resourceId'],
+                expiration=expiration_dt,
+                sync_token=sync_token,
+            )
+    except Exception as e:
+        logging.warning(f'[google_watch] Failed to set up watch channel for calendar {calendar.id}: {e}')
```
MelissaAutumn, line 83:
```
\treturn False
return True
```

I'm lazy pretend that inserts a tab.

backend/src/appointment/controller/google_watch.py
```diff
@@ -0,0 +1,126 @@
...
+    channel = repo.google_calendar_channel.get_by_calendar_id(db, calendar.id)
+    if not channel:
+        return
+
+    if google_client and calendar.external_connection and calendar.external_connection.token:
+        try:
+            token = get_google_token(google_client, calendar.external_connection)
+            google_client.stop_channel(channel.channel_id, channel.resource_id, token)
+        except Exception as e:
+            logging.warning(f'[google_watch] Failed to stop channel {channel.channel_id}: {e}')
```
MelissaAutumn: How are we going to handle clean up of this channel when we delete the db entry? Wait for ttl?

backend/src/appointment/routes/webhooks.py
```diff
@@ -110,3 +119,289 @@ def zoom_deauthorization(
...
+
+def _find_appointment_by_external_id(
+    db: Session, calendar_id: int, external_id: str
+) -> models.Appointment | None:
+    """Find an appointment on the given calendar matching the Google event ID."""
+    appointments = (
+        db.query(models.Appointment)
+        .filter(models.Appointment.calendar_id == calendar_id)
+        .all()
+    )
```
MelissaAutumn: Why not add another filter for models.Appointment.external_id == external_id?

MelissaAutumn (changes requested):
Since this PR is pretty big this might take another once over of changes after this, but getting there.

Could you additionally cut a ticket to remove pulling the google events from Google's end if they have a watch channel open? Because then we can just pull from our db and assume the db is up-to-date. It's a bit out of scope here but would be a huge perf boost.

backend/src/appointment/controller/google_watch.py
davinotdavid: For now, wait for ttl but IMHO it shouldn't be a problem because if the webhook gets called we are just logging and returning success back (so it doesn't keep retrying).

backend/src/appointment/controller/apis/google_client.py
davinotdavid: Good point, it is optional in the docs but it does add another layer of safety so I've added these in this PR already:

- Migration to add `state` column in the `google_calendar_channels` table
- Similarly, add `state` to the model as `encrypted_type(String)`
- Generate the state as `str(uuid.uuid4())` where applicable
- When receiving the webhook call, check if the token matches the `X-Goog-Channel-Token` header

This is the commit:
https://github.com/thunderbird/appointment/pull/1529/commits/a9af4b1e826aa22dc1f2dc40572d426cde9ac14d

Also added the ttl as `GOOGLE_CHANNEL_TTL_IN_SECONDS` in the .env

backend/src/appointment/routes/webhooks.py
davinotdavid: Ah thanks, indeed we should leave that to the query itself, my bad! Changed!

davinotdavid:
@MelissaAutumn ah I've also created https://github.com/thunderbird/appointment/issues/1607 to track the perf boost idea you've mentioned now that we would have webhooks setup

(Also thank you the all the through reviews, really appreciate you 🙏)

backend/src/appointment/controller/apis/google_client.py
MelissaAutumn: Perfect, it'll be good to have state so folks who knew a channel id can't just send random events to our webhooks.

backend/src/appointment/controller/calendar.py
```diff
@@ -309,9 +337,54 @@ def save_event(
 
         return event
 
-    def delete_event(self, uid: str):
+    def confirm_event(
+        self, event_id: str, event: schemas.Event = None, organizer_language: str = None,
```
MelissaAutumn: I think Noneable types need to either be wrapped in `Optional[<type>]` or `<type>|None`.

backend/src/appointment/controller/calendar.py
```diff
@@ -309,9 +337,54 @@ def save_event(
...
+
+        self.google_client.patch_event(
+            calendar_id=self.remote_calendar_id,
+            event_id=event_id,
+            body=body,
+            token=self.google_token,
+        )
+        self.bust_cached_events()
+
+    def delete_event(self, uid: str, send_updates: str = 'none'):
```
MelissaAutumn: we might want to enum that send_updates. The google library we pull in might already have it done for us.

backend/src/appointment/controller/google_watch.py
MelissaAutumn: In the future this should be turned into a celery task, and this exception might be a good play to run a retry_task call.

backend/src/appointment/database/repo/__init__.py
```diff
@@ -1 +1,13 @@
-from . import appointment, attendee, availability, calendar, external_connection, schedule, slot, subscriber  # noqa: F401
+# ruff: noqa
```
MelissaAutumn, line 1: I think you can remove this.

backend/src/appointment/routes/webhooks.py
```diff
@@ -110,3 +119,293 @@ def zoom_deauthorization(
...
+        repo.google_calendar_channel.update_sync_token(db, channel, new_sync_token)
+
+    background_tasks.add_task(
+        _process_google_event_changes,
+        calendar_id=calendar.id,
+        changed_events=changed_events,
+        google_client=google_client,
+        google_token=token,
+        remote_calendar_id=calendar.user,
+    )
```
MelissaAutumn: This looks like it could be a celery call now 😄

backend/src/appointment/routes/webhooks.py
```diff
@@ -110,3 +119,293 @@ def zoom_deauthorization(
...
+            description = [appointment.details or '']
+            if location_url:
+                description.append(l10n('join-online', {'url': location_url}, lang=owner_lang))
+
+            body['description'] = '\n'.join(description)
+
+            try:
+                remote_event = google_client.get_event(remote_calendar_id, appointment.external_id, google_token)
+                if remote_event and remote_event.get('attendees'):
+                    for att in remote_event['attendees']:
```
MelissaAutumn:
```suggestion
                    for att in remote_event.get('attendees', []):
```
Eh just to be safe.

backend/src/appointment/routes/webhooks.py
```diff
@@ -110,3 +119,293 @@ def zoom_deauthorization(
...
+                )
+            except Exception:
+                logging.warning('[webhooks.google_calendar] Failed to confirm event in Google')
+
+        logging.info(
+            f'[webhooks.google_calendar] Subscriber confirmed appointment {appointment.id} '
+            f'via Google Calendar, slot {slot.id} booked'
+        )
+
+    elif response_status == 'declined':
```
MelissaAutumn: We should enum these (google's library might already have 'em.)

backend/src/appointment/tasks/google.py
```diff
@@ -0,0 +1,14 @@
...
+
+log = logging.getLogger(__name__)
+
+
+@celery.task
+def renew_google_channels():
+    from appointment.commands.renew_google_channels import run
+
+    log.info('Starting Google Calendar channel renewal')
+    run()
```
MelissaAutumn: We might want to wrap this in an exception catch, for anything else we're not catching.

backend/src/appointment/celery_app.py
```diff
@@ -59,6 +62,10 @@ def create_celery_app() -> Celery:
                 'task': 'appointment.tasks.health.heartbeat',
                 'schedule': 60.0,
             },
+            'renew-google-channels': {
+                'task': 'appointment.tasks.google.renew_google_channels',
+                'schedule': google_channel_renew_interval,
+            },
```
MelissaAutumn, line 69: We should probably store this outside of this object, but we can do that later.

backend/.env.test
```diff
@@ -70,6 +71,7 @@ SIGNED_SECRET=test-secret-pls-ignore
 SENTRY_DSN=
 # Possible values: prod, dev, test
 APP_ENV=test
+GOOGLE_INVITE_ENABLED=True
```
MelissaAutumn: Do we have the ability to handle watch channels locally? If not then we should default this to false. 

We might want to split the variables for google invites vs handling watch channels though.

MelissaAutumn (changes requested):
Looking better! Some more nits here, I need a chance next week to test this out. 😄

backend/.env.test
davinotdavid: Not out of the box, no. It is a bit painful to setup the local development (at the moment I am doing this with cloudflare tunnels) and I am not sure how we could automate that. I've changed it to `False` then, thanks!

backend/src/appointment/database/repo/__init__.py
davinotdavid: Hmmm seems like lint check / Ruff complains about it, so I left it there. Ref: https://github.com/thunderbird/appointment/actions/runs/24053974704/job/70155976996?pr=1529

davinotdavid:
@MelissaAutumn thanks for the review again! alright, other than the smaller changes, the biggest changes in the commit diff are:

- Now processing the google events in a celery task
- Stop watching google channel is also now a task

Also, unfortunately there's no enum exported from the library so I've added those manually and added a comment to the docs for future reference.

backend/src/appointment/migrations/versions/2026_02_23_1200-a1b2c3d4e5f6_add_google_calendar_channels_.py
```diff
@@ -0,0 +1,33 @@
...
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = 'a1b2c3d4e5f6'
+down_revision = '17792ef315c1'
```
MelissaAutumn, line 15: This is conflicting with the `repair external connection status` migration. Can you do a `alembic merge heads` to fix this?

backend/src/appointment/commands/backfill_google_channels.py
```diff
@@ -0,0 +1,101 @@
...
+        models.Calendar.id.in_(schedule_calendar_ids),
+    ).all()
+
+    candidates = []
+    for cal in all_calendars:
+        existing = repo.google_calendar_channel.get_by_calendar_id(db, cal.id)
+        if not existing:
+            candidates.append(cal)
+
+    print(f'Found {len(candidates)} connected Google calendar(s) without a watch channel.')
```
MelissaAutumn, line 49: These should actually be logs if we're calling it from a task.

backend/src/appointment/controller/apis/google_client.py
```diff
@@ -12,10 +14,32 @@
...
+    EventNotDeletedException,
+    EventNotPatchedException,
+    FreeBusyTimeException
+)
 from ...exceptions.google_api import GoogleScopeChanged, GoogleInvalidCredentials
 
 
+class SendUpdates(StrEnum):
+    """Maps to the Google Calendar API ``sendUpdates`` parameter."""
+    """Ref: https://developers.google.com/workspace/calendar/api/v3/reference/events/delete"""
```
MelissaAutumn: This can be one multi-line docstring.

backend/src/appointment/controller/apis/google_client.py
```diff
@@ -12,10 +14,32 @@
...
+    ALL = 'all'
+    EXTERNAL_ONLY = 'externalOnly'
+    NONE = 'none'
+
+
+class ResponseStatus(StrEnum):
+    """Maps to the Google Calendar API attendee ``responseStatus`` field."""
+    """Ref: https://developers.google.com/workspace/calendar/api/v3/reference/events"""
```
MelissaAutumn: Ditttooo.

backend/src/appointment/controller/apis/google_client.py
```diff
@@ -246,17 +280,150 @@ def save_event(self, calendar_id, body, token):
...
+        with build('calendar', 'v3', credentials=token, cache_discovery=False) as service:
+            try:
+                response = service.events().insert(
+                    calendarId=calendar_id, body=body, sendUpdates=send_updates
+                ).execute()
+            except HttpError as e:
+                logging.warning(f'[google_client.insert_event] Request Error: {e.status_code}/{e.error_details}')
+                raise EventNotCreatedException()
+
+        return response
```
MelissaAutumn: This is probably a copy and paste quirk and originally my bad, but can we consolidate on return in try/except vs return value is null to begin with and return after try/except?

backend/src/appointment/controller/apis/google_client.py
```diff
@@ -246,17 +280,150 @@ def save_event(self, calendar_id, body, token):
...
                 logging.warning(f'[google_client.delete_event] Request Error: {e.status_code}/{e.error_details}')
                 raise EventNotDeletedException()
 
         return response
 
+    def watch_events(self, calendar_id, webhook_url, token, state: str):
+        """Register a push notification channel for calendar event changes.
+        Ref: https://developers.google.com/calendar/api/v3/reference/events/watch"""
+        channel_id = str(uuid.uuid4())
+        ttl = os.getenv('GOOGLE_CHANNEL_TTL_IN_SECONDS', 604800) # 7 days
```
MelissaAutumn: Do we have a defines.py we can shove the 7 days in seconds into?

backend/src/appointment/controller/apis/google_client.py
```diff
@@ -246,17 +280,150 @@ def save_event(self, calendar_id, body, token):
...
+                logging.warning(f'[google_client.stop_channel] Request Error: {e.status_code}/{e.error_details}')
+
+    def list_events_sync(self, calendar_id, sync_token, token):
+        """Fetch events that changed since the last sync token.
+        Ref: https://developers.google.com/calendar/api/v3/reference/events/list (incremental sync)"""
+        items = []
+        next_sync_token = None
+
+        with build('calendar', 'v3', credentials=token, cache_discovery=False) as service:
+            page_token = None
```
MelissaAutumn: Wanna move this with next_sync_token? It looks a little lonely.

backend/src/appointment/controller/calendar.py
```diff
@@ -281,26 +285,58 @@ def save_event(
...
-                'displayName': organizer.name,
-                'email': self.remote_calendar_id,
-            },
-        }
-
-        new_event = self.google_client.save_event(
-            calendar_id=self.remote_calendar_id, body=body, token=self.google_token
-        )
+        if send_google_notification:
+            if booking_confirmation:
```
MelissaAutumn: This looks like a one line change, can we simplify this to like:
```python
organizer_response = ResponseStatus.NEEDS_ACTION if booking_confirmation else ResponseStatus.ACCEPTED
attendees = [
                    {'displayName': organizer.name, 'email': organizer_email,
                     'responseStatus': organizer_response},
                    {'displayName': attendee.name, 'email': attendee.email,
                     'responseStatus': ResponseStatus.NEEDS_ACTION},
                ]

```

backend/src/appointment/controller/calendar.py
```diff
@@ -281,26 +285,58 @@ def save_event(
...
+                     'responseStatus': ResponseStatus.NEEDS_ACTION},
+                ]
+
+            body = {
+                'summary': event.title,
+                'location': event.location.url if event.location.url else None,
+                'description': '\n'.join(description),
+                'start': {'dateTime': event.start.isoformat()},
+                'end': {'dateTime': event.end.isoformat()},
+                'status': 'tentative' if booking_confirmation else 'confirmed',
```
MelissaAutumn: Missing some enums 😄

backend/src/appointment/controller/apis/google_client.py
MelissaAutumn: Also I believe enums refer to an enum item. So SendUpdates.ALL is not strictly 'all', but str(SendUpdates.ALL) == 'all' / SendUpdates.ALL.value == 'all'. 

Can you double check that and update the enum checks accordingly?

backend/src/appointment/controller/calendar.py
```diff
@@ -309,9 +345,57 @@ def save_event(
...
+        event: schemas.Event | None = None,
+        organizer_language: str | None = None,
+    ):
+        """Patch a tentative event to confirmed status, notifying attendees.
+
+        When *event* is provided the patch also updates summary, location,
+        and description so that a newly-created meeting link is visible on
+        the Google Calendar event.
+        """
+        body = {'status': 'confirmed'}
```
MelissaAutumn: enuuum?

backend/src/appointment/database/repo/__init__.py
MelissaAutumn: oic

backend/src/appointment/routes/schedule.py
```diff
@@ -210,7 +209,25 @@ def update_schedule(
...
+    if os.getenv('GOOGLE_INVITE_ENABLED') == 'True':
+        _sync_watch_channels(db, google_client, subscriber, schedule.calendar_id)
+
+    return result
+
+
+def _sync_watch_channels(db: Session, google_client: GoogleClient, subscriber: Subscriber, default_calendar_id: int):
+    """Ensure a watch channel exists only for the schedule's default Google calendar.
+    Tears down channels on any other Google calendars for this subscriber."""
+    calendars = repo.calendar.get_by_subscriber(db, subscriber.id)
```
MelissaAutumn: At some point we should allow for filtering by calendar type on this.

backend/src/appointment/routes/schedule.py
```diff
@@ -605,46 +643,16 @@ def handle_schedule_availability_decision(
...
-                appointment_title=schedule.name,
-                lang=subscriber.language,
-            )
-        except OAuth2Error as err:
-            logging.error('OAuth flow error during zoom meeting creation: ', err)
-            if os.getenv('SENTRY_DSN') != '':
-                capture_exception(err)
-
-            # Notify the organizer that the meeting link could not be created!
+        zoom_url = zoom.create_meeting_link(db, slot, subscriber, title)
```
MelissaAutumn, line 646: In the future we should punt zoom link creation (along with the appointment_id so we can update it after) to a celery task.

backend/src/appointment/routes/webhooks.py
```diff
@@ -110,3 +116,73 @@ def zoom_deauthorization(
...
+        logging.warning(f'[webhooks.google_calendar] Unknown channel_id: {channel_id}')
+        return success_response
+
+    incoming_state = request.headers.get('X-Goog-Channel-Token')
+    if incoming_state != channel.state:
+        logging.warning(f'[webhooks.google_calendar] State mismatch for channel {channel_id}')
+        return success_response
+
+    calendar = channel.calendar
+    if not calendar:
```
MelissaAutumn, line 149: Double check the webhook spec on google's end. I know a lot of webhooks tend to want you to respond within a fixed time or else they'll queue for a resend. We don't want multiple notifications so we should ideally return 200 now and punt the processing to a task. 

That task can have some retry logic in case of failure.

backend/src/appointment/tasks/google.py
```diff
@@ -0,0 +1,250 @@
...
+
+            appointment = repo.appointment.get_by_calendar_and_external_id(db, calendar_id, google_event_id)
+            if not appointment:
+                continue
+
+            slot = appointment.slots[0] if appointment.slots else None
+            if not slot or not slot.attendee:
+                continue
+
+            if event.get('status') == 'cancelled':
```
MelissaAutumn: enuuuum?

backend/src/appointment/tasks/google.py
```diff
@@ -0,0 +1,250 @@
...
+                continue
+
+            response_status = google_attendee.get('responseStatus')
+            _handle_bookee_rsvp(
+                db, appointment, slot, response_status, google_client, google_token, remote_calendar_id
+            )
+    except Exception as e:
+        log.error(f'[tasks.google] Error processing event changes: {e}')
+        if sentry_sdk.is_initialized():
+            sentry_sdk.capture_exception(e)
```
MelissaAutumn: Does sentry_sdk noop their functions if it's not initialized?

MelissaAutumn (changes requested):
Sorry more buckets of comments 😅

backend/src/appointment/commands/backfill_google_channels.py
davinotdavid: Oh these are not going to be called from a task as this is a one-off command to be run once we get this to stage / prod for existing users. Good call though, I've changed the `renew_google_channels` task to log instead of print!

backend/src/appointment/migrations/versions/2026_02_23_1200-a1b2c3d4e5f6_add_google_calendar_channels_.py
davinotdavid: TIL, nice! I didn't know there was an `alembic` command for that

backend/src/appointment/controller/apis/google_client.py
davinotdavid: Yep, the idea of using`StrEnum` is to cope well with the plain strings from the Google API for comparison. Do you mean that it should be strictly 'all' / a standard Enum instead of StrEnum?

MelissaAutumn: Ah just gave it a run, and it seems like it all resolves okay. Was just worried it would trip up if there was a case where it didn't toStr itself correctly, but seems okay.

backend/src/appointment/routes/webhooks.py
davinotdavid: That's fair and indeed we should return 200 as fast as possible to avoid resend. Punted this into a celery task for processing now!

backend/src/appointment/tasks/google.py
davinotdavid: Whoops, indeed not needed, good catch

backend/src/appointment/routes/schedule.py
davinotdavid: Good point, created an issue here for tracking: https://github.com/thunderbird/appointment/issues/1619 and added a comment in place

davinotdavid:
@MelissaAutumn nice! Each review pass is making this PR better and better, thanks a lot! 🙏

devmount:
@MelissaAutumn @davinotdavid Wow you guys are awesome! Thanks so much for your work on this. Sorry I didn't have the time to review this 🙏🏻

backend/src/appointment/commands/renew_google_channels.py
```diff
@@ -0,0 +1,87 @@
...
+
+        external_connection = calendar.external_connection
+        if not external_connection or not external_connection.token:
+            repo.google_calendar_channel.delete(db, channel)
+            continue
+
+        token = Credentials.from_authorized_user_info(json.loads(external_connection.token), google_client.SCOPES)
+
+        # Stop the old channel (fire-and-forget via Celery with retries)
+        from ..tasks.google import stop_google_channel
```
MelissaAutumn, line 55: If it cannot live at the top, bump this to the top of the function please.

backend/src/appointment/controller/apis/google_client.py
```diff
@@ -241,28 +273,159 @@ def list_events(self, calendar_id, time_min, time_max, token):
...
+        channel_id = str(uuid.uuid4())
+        ttl = os.getenv('GOOGLE_CHANNEL_TTL_IN_SECONDS', SEVEN_DAYS_IN_SECONDS)
+
+        body = {
+            'id': channel_id,
+            'type': 'web_hook',
+            'address': webhook_url,
+            'token': state,
+            'params': {
+                'ttl': str(ttl)
```
MelissaAutumn, line 340: I believe os.getenv returns a string. If you want to str() wrap it around the os.getenv.

MelissaAutumn (approved):

backend/src/appointment/migrations/versions/2026_02_23_1200-a1b2c3d4e5f6_add_google_calendar_channels_.py
devmount: Oh nice, didn't know that either! 🎉

devmount (approved):
Thanks so much Davi for this heavy lift! Here are the things I successfully tested:

- Booking an event on a Google calendar with manual confirmation (Google event status was set and updated correctly, emails were correctly sent)
- Booking an event on a Google calendar with auto-confirmation (Google event status was set correctly, emails were correctly sent)
- Booking an event on a CalDAV calendar with manual confirmation (HOLD events were created and updated correctly)
- Booking an event on a CalDAV calendar with auto-confirmation (event was created correctly, emails were correctly sent)

I wasn't able to setup tests for the google notifications, sorry about that.

Issue #1035, opened 2025-05-22
Timezone Issue with UTC


MelissaAutumn:
ZoneInfoNotFoundError
'No time zone found with key +00:00'

Sentry error: https://thunderbird.sentry.io/issues/6608085565/

MelissaAutumn:
If a timezone conversion fails we should default to UTC, but in this case we should probably convert +00:00 to UTC.

MelissaAutumn:
Seems like dj.tz.guess() is basically just calling `Intl.DateTimeFormat.resolvedOptions()`.

We can re-create this specific issue here:
`new Intl.DateTimeFormat("en-US", { timeZone: "+00:00" }).resolvedOptions()`

It seems to be a valid timezone offset, but our zoneinfo database [tzdata](https://github.com/python/tzdata) doesn't contain it.

rwood-moz:
FYI: The UTC timezone shows up on the Appointment share link / booking page correctly on iOS as 'UTC' but on android it appears as '+00:00'.

<img width="145" height="94" alt="Image" src="https://github.com/user-attachments/assets/d71f4e23-d7ab-4d5e-9562-621b57ac731b" />

<img width="122" height="73" alt="Image" src="https://github.com/user-attachments/assets/3efb75f8-b7f1-4bba-b020-28d9bc9d90c5" />

rwood-moz:
Booking an appointment on android on BrowserStack fails with an error, I'm guessing it's because the BrowserStack android device is at UTC but the booking page is reporting the timzeone as `+00:00`; booking an appointment on iOS is working fine (where the booking page sees the timezone as `UTC`). Android screenshot:

<img width="177" height="356" alt="Image" src="https://github.com/user-attachments/assets/1319cdc1-ce73-4f24-9226-01b60cad0570" />

rwood-moz:
@davinotdavid @MelissaAutumn Any chance this could be fixed (is it complex)? This is blocking getting the book appointment E2E test running on Android Chrome (#1527).

davinotdavid:
@rwood-moz I can move this up in priority and start working on this today to unblock that. Doesn't seem to be complex!

MelissaAutumn:
@davinotdavid I would suggest since this is seemingly erroring out on non-standard UTC entries that you should just catch the error and yolo set them to UTC.

davinotdavid:
@MelissaAutumn I followed your suggestion and it worked well, thanks a lot!

PR #1531, opened 2026-02-25
Default to UTC for timezone keys outside of zoneinfo package

davinotdavid:
## Description of the Change

The `dayjs`'s `guess()` function that we use on the frontend may use unconventional timezone keys such as '+00:00' for UTC which doesn't exist in the [zoneinfo package](https://docs.python.org/3/library/zoneinfo.html) and it throws an exceptions like this one: https://thunderbird.sentry.io/issues/7291473530/?project=4505428124827648&query=is%3Aunresolved&referrer=issue-stream

This is a blocking problem for BrowserStack's E2E for Android devices as they are in UTC and set `+00:00` as timezone instead of UTC.

## Benefits

We won't have timezone errors anymore for unconventionally formatted timezones sent by the frontend

## Applicable Issues

Related https://github.com/thunderbird/appointment/issues/1035

### DIFF
--- a/backend/test/unit/test_calendar_tools.py
+++ b/backend/test/unit/test_calendar_tools.py
@@ -1,7 +1,7 @@
 from appointment.controller.calendar import Tools, GoogleConnector
 from appointment.database import schemas, models
 from datetime import datetime, timedelta, time, date, timezone
-from unittest.mock import Mock, MagicMock, PropertyMock, patch
+from unittest.mock import Mock, MagicMock, PropertyMock
 from starlette_context import request_cycle_context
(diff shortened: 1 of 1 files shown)


rwood-moz:
After the PR landed and was deployed to stage, I verified on a Google Pixel 10 on BrowserStack that the error is gone, and an appointment can now be booked on Android Chrome when the device's timezone is UTC. 🥇 Verified using my branch in PR (#1527) that is enabling the book appt test on mobile. Screenshot from BrowserStack Google Pixel 10:

Screenshot 2026-02-25 at 4 43 09 PM

Awesome! Thanks @davinotdavid!

Issue #1533, opened 2026-02-25
Mailer may fail sending a message (SMTPSenderRefused)


davinotdavid:
**Describe the bug**
We had instances where emails were not being delivered and in at least one of the instances it was because the gap of time between opening a connection to the SMTP server and actually sending a message took too long.

Sentry:
https://thunderbird.sentry.io/issues/7255683634/events/295bf53720a9457d910d7226f90d73bd/?project=4505428124827648

PR #1534, opened 2026-02-25
Build email before connecting to SMTP server

davinotdavid:
## Description of the Change

There were instances were we failed to build the email templates / message in time to actually send the email and while we are attempting to build it, the connection remains open and does a timeout.

In this PR, we are now building the email templates / message _before_ we connect to the server so that (hopefully) we don't run into timeouts anymore.

Sentry error for reference:
https://thunderbird.sentry.io/issues/7255683634/events/295bf53720a9457d910d7226f90d73bd/?project=4505428124827648

## Benefits

- One less reason for failing email delivery!

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/1533

### DIFF
--- a/backend/src/appointment/controller/mailer.py
+++ b/backend/src/appointment/controller/mailer.py
@@ -147,6 +147,7 @@ class Mailer:
             # url is not valid
             logging.error('[mailer.send] No valid SMTP url configured: ' + url)
 
+        message = self.build()
         server = None
         try:
             # if configured, create a secure SSL context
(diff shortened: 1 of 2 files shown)


davinotdavid:
@MelissaAutumn thanks for the review! It doesn't seem like we do this anywhere else and all the other mailers inherit from this one `Mailer`.

Good point! A task/queue would be great indeed, probably worth bumping https://github.com/thunderbird/appointment/issues/82 in priority so I will move it into Earlybird (also wow, that was opened in 2023 😱 )

PR #1539, opened 2026-02-26
Appointment data deletion: Password confirmation and new modal

### LINKED ISSUE #1538: Replace legacy modal "Delete all appointment data" with ModalDialog from services-ui

Current design from Zeplin:

Image

devmount:
## Description of the Change

This change contains:

- Extending the backend to confirm the password first before deleting all data
- Replacing the legacy modal in the frontend with the ModalDialog from services-ui
- Extended error handling with localized messages for the user if something went wrong (e.g. the wrong password was given)

Modal before user interaction:

image

Modal after user input:

image

Modal after wrong password:

image

## Benefits

UI/UX improvements, additional check before deleting everything.

## Applicable Issues

Closes #1538

### DIFF
--- a/frontend/src/views/SettingsView/components/AccountSettings.vue
+++ b/frontend/src/views/SettingsView/components/AccountSettings.vue
@@ -1,19 +1,18 @@
 <script setup lang="ts">
 import { computed, inject, ref, useTemplateRef } from 'vue';
-import { useRouter } from 'vue-router';
 import { useI18n } from 'vue-i18n';
 import { storeToRefs } from 'pinia';
 import { callKey } from '@/keys';
-import { TextInput, IconButton, ModalDialog, DangerButton, LinkButton, CheckboxInput, NoticeBar, NoticeBarTypes } from '@thunderbirdops/services-ui';
-import { PhCopySimple, PhArrowRight, PhDownloadSimple, PhX } from '@phosphor-icons/vue';
+import { TextInput, IconButton } from '@thunderbirdops/services-ui';
+import { PhCopySimple, PhArrowRight, PhDownloadSimple } from '@phosphor-icons/vue';
 import { createUserStore } from '@/stores/user-store';
 import { useSettingsStore } from '@/stores/settings-store';
-import { Alert, BlobResponse, BooleanResponse, Exception } from '@/models';
+import { BlobResponse } from '@/models';
 import { posthog, usePosthog } from '@/composables/posthog';
 import { MetricEvents } from '@/definitions';
+import DeleteAccountModal from './DeleteAccountModal.vue';
 
 const { t } = useI18n();
-const router = useRouter();
 
 const call = inject(callKey);
 const userStore = createUserStore(call);
@@ -23,10 +22,6 @@ const { currentState } = storeToRefs(settingsStore);
 const copyLinkTooltip = ref(t('label.copyLink'));
 
 const deleteModal = useTemplateRef('deleteModal');
-const consentToDeletion = ref(false);
-const confirmPassword = ref('');
-const supportUrl = import.meta.env?.VITE_SUPPORT_URL;
-const validationError = ref<Alert>(null);
 
 const displayName = computed({
   get: () => currentState.value.displayName,
@@ -44,39 +39,6 @@ const copyLink = async () => {
   setTimeout(() => {
     copyLinkTooltip.value = t('label.copyLink');
   }, 2000);
-};
-
-/**
- * Reset the form data.
- */
-const resetData = () => {
-  consentToDeletion.value = false;
-  confirmPassword.value = '';
-  validationError.value = null;
-};
-
-/**
- * Request an appointment data deletion, and then log out.
- */
-const actuallyDeleteAccount = async () => {
-  const pw = confirmPassword.value;
-
-  const { data, error }: BooleanResponse = await call('account/delete').delete({
-    password: pw,
-  }).json();
-
-  if (usePosthog) {
-    posthog.capture(MetricEvents.DeleteAccount);
-  }
-
-  if (error.value) {
-    validationError.value = { title: (data.value as Exception).detail as string };
-    return;
-  }
-
-  // We can't logout since we've deleted the user by now, so just delete local storage data.
-  userStore.$reset();
-  await router.push('/');
 };
 
 /**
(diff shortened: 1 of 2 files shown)


frontend/src/views/SettingsView/components/AccountSettings.vue
```diff
@@ -131,11 +145,75 @@ const actuallyDownloadData = async () => {
...
+          {{ t('heading.deleteAppointmentData') }}
+        </danger-button>
+      </div>
+    </div>
+
+    <template #actions>
+      <link-button
+        name="cancel"
+        @click="deleteModal?.hide()"
+        class="cancel-button"
```
davinotdavid: Seems like this class is unused at the moment and also differs a little bit from the styling in Zeplin:

Current implementation

Image

Zeplin

Image

frontend/src/views/SettingsView/components/AccountSettings.vue
```diff
@@ -229,6 +307,25 @@ h2 {
...
+.delete-modal-container {
+  display: flex;
+  flex-direction: column;
+  gap: 1rem;
+  padding-bottom: .25rem;
+
+  .password-confirmation {
+    display: flex;
+    gap: 1rem;
+    margin-top: 1.5rem;
```
davinotdavid: It seems like there shouldn't be a `margin-top` here at all as the gap of `1rem` from the parent should already cover the spacing.

Also, it seems that the password input should be slightly larger (perhaps something like a `flex-grow: 0.5` ?)

Current implementation:

Image

Zeplin:
Image

frontend/src/views/SettingsView/components/AccountSettings.vue
```diff
@@ -229,6 +307,25 @@ h2 {
...
+
+    button {
+      align-self: flex-start;
+      margin-top: 1.75rem;
+      line-height: 1.25;
+    }
+  }
+}
+
```
davinotdavid: Although we don't have official™ mobile design for this modal, I believe we should keep the checkbox's size and do some flex direction updates for the text input / button.

Current implementation:

Image

backend/src/appointment/routes/account.py
```diff
@@ -54,11 +59,17 @@ def download_data(db: Session = Depends(get_db), subscriber: Subscriber = Depend
...
+    form_data: schemas.CheckPassword,
     db: Session = Depends(get_db),
     subscriber: Subscriber = Depends(get_subscriber),
     posthog: Posthog = Depends(get_posthog),
 ):
     """Delete your account and all the data associated with it forever!"""
+    try:
+        utils.verify_password(form_data.password, subscriber.password)
+    except argon2.exceptions.VerifyMismatchError:
+        raise HTTPException(status_code=401, detail=l10n('password-mismatch'))
```
davinotdavid: We should also add an `except Exception` here for a catch all other exception types with a generic error message and a status_code of 500.

I've ran into this when using `OIDC` / `Keycloak` as `subscriber.password` is `None` in that case.

For now, if we want it to work, we can either:

- Skip this verification in case the auth scheme is oidc or;
- If it is oidc, implement the flow to re-auth through Keycloak (and probably remove the password confirmation bits from the frontend as well)

I'd say that we can do the first option for now and add a new ticket for implementing data deletion through Keycloak re-auth separately as that might involve slightly different copy / UI.

frontend/src/views/SettingsView/components/AccountSettings.vue
```diff
@@ -131,11 +145,75 @@ const actuallyDownloadData = async () => {
   </div>
 
   <!-- Delete Appointment Data modal -->
-  <confirmation-modal :open="deleteAppointmentDataModalOpen" :title="t('heading.deleteAppointmentData')"
-    :message="t('text.deleteAppointmentDataWarning')" :confirm-label="t('heading.deleteAppointmentData')"
-    :cancel-label="t('label.cancel')" :use-caution-button="true" @confirm="actuallyDeleteAccount"
-    @close="deleteAppointmentDataModalOpen = false">
-  </confirmation-modal>
+  <modal-dialog ref="deleteModal" class="delete-modal" @closed="resetData">
```
davinotdavid: Sorry to ask but since the code for this modal is already quite big considering the templates + function handlers + state, I believe it would be best to move it to its own `DeleteAccountModal.vue` component instead.

backend/src/appointment/routes/account.py
devmount: Definitely, thanks catching this! I implemented the first option, left a TODO in the code for the second and created #1545 for this matter.

devmount:
@rwood-moz Thanks so much for your review! 👏🏻 

1. **Localized messages:** This is kinda intended 😅 Since this message comes from the backend, we currently try to retrieve the language from the context (NOT from the subscriber lang setting). That means, if you set Appointment to German, but still use an English browser, you will get English notifications from the backend. If you use a German browser, you will indeed get German notifications:

    image

    For endpoints, that are entirely purposed for subscribers only, we could throw in the subscribers language setting into the `l10n()` calls. @davinotdavid What do you think about this? Should we file a ticket?

2. **Delete calendar events too when account is deleted:** Great question. Personally I would be very cautious automatically deleting calendar events when removing the Appointment account. Maybe we could add a checkbox saying "Remove related events from all remote calendars too" which is unchecked by default? I would leave the implementation as is for now, since it is kinda slightly out-of-scope here and requires a bit more thought. Should we file a ticket?

frontend/src/views/SettingsView/components/AccountSettings.vue
devmount: Thanks, yes I forgot to that, great catch!

frontend/src/views/SettingsView/components/AccountSettings.vue
devmount: Another leftover, thanks!

frontend/src/views/SettingsView/components/AccountSettings.vue
devmount: Oh that's right! That would also be relevant for services-ui to take care of. I'll make a note/ticket for that, but fix the responsiveness here still. Thanks!!

frontend/src/views/SettingsView/components/AccountSettings.vue
devmount: Absolutely! Thanks so much for the suggestion. I tend to make components a little bit too big at times 😅 So appreciated and fixed!

devmount:
@davinotdavid Thanks a lot for your great finds and suggestions! Mind giving it another spin? 😇

rwood-moz:
Thanks @devmount 

>     1. **Localized messages:** This is kinda intended 😅 Since this message comes from the backend, we currently try to retrieve the language from the context (NOT from the subscriber lang setting). That means, if you set Appointment to German, but still use an English browser, you will get English notifications from the backend. If you use a German browser, you will indeed get German notifications:

Oh sorry I didn't realize that, I thought that would come from the language selection in Appointment settings.

>     2. **Delete calendar events too when account is deleted:** Great question. Personally I would be very cautious automatically deleting calendar events when removing the Appointment account. Maybe we could add a checkbox saying "Remove related events from all remote calendars too" which is unchecked by default? I would leave the implementation as is for now, since it is kinda slightly out-of-scope here and requires a bit more thought. Should we file a ticket?

Yes I agree - we shouldn't remove the events - actually I meant to say "cancelled" not deleted. After deleting all of my Appointment data, the confirmed bookings for the future still remain in my calendar not cancelled - but the text on the delete data dialog says `all upcoming bookings will be cancelled` but they still exist in my calendar so that text is a little bit misleading IMO. It's true though I don't think we should delete events. Yes maybe a ticket would be good to map out the desired behaviour. When a person deletes their Appointment data do we send an event cancellation to all the attendees so that the person who selected the timeslot will know the event no longer exists?

devmount:
@rwood-moz 

> Oh sorry I didn't realize that, I thought that would come from the language selection in Appointment settings.

And that is completely understandable! From a users point of view it totally makes sense to make that depending on the language selection in the settings. If @davinotdavid agrees, we can create a ticket for it.

> but the text on the delete data dialog says all upcoming bookings will be cancelled but they still exist in my calendar so that text is a little bit misleading IMO

Ohhh I see!! Yes, that is indeed very misleading. I created #1546 to solve this. Thanks so much again!

devmount:
@rwood-moz I believe you need to approve this PR so that I can merge it in 😇 Otherwise it is still waiting for the requested changes :)

Issue #1352, opened 2025-11-24
Connect to CalDAV with OIDC token


MelissaAutumn:
Should be possible to use DAVClient's HTTPBearerAuth class https://github.com/python-caldav/caldav/blob/master/caldav/requests.py#L7 in the DAVClient's auth parameter and pass the current access token you got from Keycloak as the password (https://github.com/thunderbird/appointment/blob/main/backend/src/appointment/controller/calendar.py#L339)

davinotdavid:
As an update to this task, after some failed attempts by me of making it work, in the end @MelissaAutumn mentioned that it wouldn't be a good idea to use the token alone for the connection anyway since we would have to store it in the `ExternalConnection` DB table to interact with the calendar after the FTUE.

The new approach now is to do a joint effort with `thunderbird-accounts`:
- Add a `/appointment/setup/` route in `thunderbird-accounts` to be called only by the Appointment backend
- Appointment backend passes the current OIDC token to the endpoint
- `thunderbird-accounts` creates or retrieves a special / specific App Password on the user's Stalwart account
- Appointment then finally connects to the TB Pro Stalwart CalDAV calendar and stores the App Password instead of the OIDC token

Note: We'll probably want to add a secret to ensure only Appointment can call that route.

PR #1544, opened 2026-02-26
Automagically connect Thundermail Calendar in FTUE

davinotdavid:
## Description of the Change

- Updated `docker-compose.yml` container ports not to clash with `thunderbird-accounts` when running locally
- Added `TB_ACCOUNTS_CALDAV_URL` env var.
- Added `POST '/caldav/oidc/auth'` route to attempt to connect to the CalDAV through the OIDC token.
- Removed the `ConnectYourCalendarThundermailStep` as we don't need it anymore + related i18n and enum entries.

Here's the flow:
- Frontend makes a `POST` request to `/caldav/oidc/auth`
- Appointment's Backend makes a `POST` request to Account's Backend `/appointment/caldav/setup/` passing the `APPOINTMENT_CALDAV_SECRET` and the current logged in user's OIDC token
- Account's Backend deletes previously existing App Passwords that match a specific label and creates a new App Password in Stalwart. The password is returned in plain text as we will need it in Appointment's backend to connect to it but it is hashed when saved
- Appointment's Backend takes the subscriber's email + newly created App Password and connects to the CalDAV server + creates the ExternalConnection entry

> [!WARNING]
> Requires https://github.com/thunderbird/thunderbird-accounts/pull/616 to be merged in first for the best experience on testing this

## How to test it

Alright, first of all, this is a joint effort with the `thunderbird-accounts` repo so make sure you are in the branch `appointment-auto-setup-caldav` (https://github.com/thunderbird/thunderbird-accounts/pull/469) and you are able to run a `docker compose up` in both `thunderbird-accounts` and this branch in `Appointment`.

Make sure you have the .env var `APPOINTMENT_CALDAV_SECRET` set in both `thunderbird-accounts` and this branch to the same values.

#### Accounts Prerequisites

- Navigate to the Django Admin in `http://localhost:8087/admin` and login with user `admin@example.org` and password `admin`.
- Create a Subscription Plan manually in `http://localhost:8087/admin/subscription/plan/add/`. None of the information here actually matter except for the Product which has to be an active product.
- Now navigate again to `http://localhost:8087/` and you should be redirected to the `/subscribe` page.
- Use [Paddle's test cards](https://developer.paddle.com/concepts/payment-methods/credit-debit-card#test-payment-method) for the checkout.
- After you are redirected to the dashboard, you are all set!

PS: This whole step is required because only then we will have the `authenticate.User` model created + the actual Stalwart account to create the App Password.

#### Steps to test, now in Appointment-land

- Navigate to `http://localhost:8090`, if you are redirected to Keycloak, login with the same `admin@example.org` / `admin` as before.
- Go through the FTUE until the step to select which calendar to connect with and select `Thundermail Calendar`
- Click Next and see that you are now in the `Booking Page` step with the Stalwart Calendar pre-selected. All that without manually creating an App Password!

## Benefits

- 1 click configuration for Thundermail Calendar!

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/1352

### DIFF
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -87,6 +87,7 @@ TB_ACCOUNTS_HOST=http://localhost:8087
 TB_ACCOUNTS_CALLBACK=http://localhost:5000/accounts/callback
 TB_ACCOUNTS_CLIENT_ID
 TB_ACCOUNTS_SECRET
+# Might not work in all Docker setups, change this accordingly if need be!
 TB_ACCOUNTS_CALDAV_URL=http://host.docker.internal:8081
 
 # -- GOOGLE AUTH --
@@ -143,4 +144,4 @@ OIDC_TOKEN_INTROSPECTION_URL=
 OIDC_FALLBACK_MATCH_BY_EMAIL=
 
 # Required for Appointment's CalDAV auto-setup, needs to match the one in thunderbird-accounts
-APPOINTMENT_CALDAV_SECRET=changemeplz
+APPOINTMENT_CALDAV_SECRET=


backend/src/appointment/routes/caldav.py
```diff
@@ -135,6 +137,130 @@ def caldav_autodiscover_auth(
...
+        lookup_url = redis_client.get(dns_lookup_cache_key)
+
+    if lookup_url and 'http' not in lookup_url:
+        debug_obj = {'url': lookup_url, 'branch': 'CACHE'}
+        # Raise and catch the unexpected behaviour warning so we can get proper stacktrace in sentry...
+        try:
+            sentry_sdk.set_extra('debug_object', debug_obj)
+            raise UnexpectedBehaviourWarning(message='Cache incorrect', info=debug_obj)
+        except UnexpectedBehaviourWarning as ex:
+            sentry_sdk.capture_exception(ex)
```
MelissaAutumn: Since this is entirely thunderbird accounts we probably don't need the debug object information. This was just to capture information about failing caldav servers (and I think we've caught all those edge-cases too.)

backend/src/appointment/routes/caldav.py
```diff
@@ -135,6 +137,130 @@ def caldav_autodiscover_auth(
     return True
 
 
+@router.post('/oidc/auth')
+def oidc_autodiscover_auth(
+    db: Session = Depends(get_db),
+    subscriber: models.Subscriber = Depends(get_subscriber),
+    redis_client: Redis = Depends(get_redis),
+    token: str = Depends(oauth2_scheme),
```
MelissaAutumn: Be careful with oauth2_scheme because I think we're using it in a way it's not intended. (we used to generate local oauth tokens lol.) 

Calling it just retrieves the authorization header item so that's all fine, but in the future it might do somethings that are not intended. https://github.com/fastapi/fastapi/blob/0127069d47b54aedb179931a862036f0a49c7554/fastapi/security/oauth2.py#L536-L544

Might be worth creating a dep of our own that handles both oauth and local password similar to oauth2_scheme.

backend/src/appointment/routes/caldav.py
```diff
@@ -135,6 +137,130 @@ def caldav_autodiscover_auth(
...
+    # Do a dns lookup first
+    if lookup_url is None:
+        parsed_url = urlparse(connection_url)
+        lookup_url, ttl = Tools.dns_caldav_lookup(parsed_url.hostname, secure=secure_protocol)
+        # set the cached lookup for the remainder of the dns ttl
+        if redis_client and lookup_url:
+            redis_client.set(dns_lookup_cache_key, utils.encrypt(lookup_url), ex=ttl)
+    else:
+        # Extract the cached value
+        lookup_url = utils.decrypt(lookup_url)
```
MelissaAutumn: I guess this is still handy for the sake of self-hosting and if our caldavs ever change. Can you shove this into a shared function with the other caldav discovery route? I think maybe we can merge a lot of this code and just return the final url to look at.

backend/src/appointment/routes/caldav.py
```diff
@@ -135,6 +137,130 @@ def caldav_autodiscover_auth(
...
+            timeout=10,
+        )
+
+        response.raise_for_status()
+        result = response.json()
+
+        if not result.get('success'):
+            raise RemoteCalendarConnectionError()
+
+        app_password = result['app_password']
```
MelissaAutumn:
```suggestion
        app_password = result.get('app_password')
        if not app_password:
            raise RemoteCalendarConnectionError()
```
(Might need a lint format)

backend/src/appointment/routes/caldav.py
```diff
@@ -135,6 +137,130 @@ def caldav_autodiscover_auth(
...
+    try:
+        if not con.test_connection():
+            raise RemoteCalendarConnectionError()
+    except TestConnectionFailed as ex:
+        raise RemoteCalendarConnectionError(reason=ex.reason)
+
+    caldav_name = subscriber.email
+    caldav_id = json.dumps([connection_url, caldav_name])
+
+    external_connection = repo.external_connection.get_by_type(
```
MelissaAutumn: I feel like this code can be merged into with the existing caldav auto discover.

davinotdavid: True, lots of copy and paste around so extracted 2 the caldav url resolution and upsert of the external connection in their own functions. Thanks!

backend/src/appointment/routes/caldav.py
davinotdavid: Yes! Extracted this into a `_resolve_caldav_url` helper, thanks!

backend/src/appointment/routes/caldav.py
davinotdavid: Oh good call! I've copied over the bearer token extraction and replaced the usage with a new `get_bearer_token`. Thanks!

backend/src/appointment/routes/caldav.py
davinotdavid: Hmmm fair but since I've extracted this into a shared `_resolve_caldav_url` for both cases, perhaps it wouldn't hurt to leave it?

backend/src/appointment/dependencies/auth.py
```diff
@@ -20,7 +19,25 @@
 from ..exceptions.validation import InvalidTokenException, InvalidPermissionLevelException
 from ..utils import encrypt, decrypt, get_expiry_time_with_grace_period
 
-oauth2_scheme = OAuth2PasswordBearer(tokenUrl='token', auto_error=False)
+
+async def get_bearer_token(request: Request) -> str | None:
```
MelissaAutumn, line 23: Oh thanks, but this was okay to punt to another ticket 😅

backend/src/appointment/routes/auth.py
```diff
@@ -435,7 +435,7 @@ def fxa_callback(
         data={'sub': f'uid-{subscriber.id}', 'jti': secrets.token_urlsafe(16)}, expires_delta=access_token_expires
     )
 
-    return RedirectResponse(f'{os.getenv("FRONTEND_URL", "http://localhost:8080")}/post-login/{one_time_access_token}')
+    return RedirectResponse(f'{os.getenv("FRONTEND_URL", "http://localhost:8090")}/post-login/{one_time_access_token}')
```
MelissaAutumn, line 438: In a future ticket we should move these to a constant value.

davinotdavid:
@MelissaAutumn sorry for the ping on the re-review but when you have some time, could take another look on this PR? 🙏

backend/.env.example
```diff
@@ -140,3 +141,6 @@ OIDC_CLIENT_ID=
 OIDC_CLIENT_SECRET=
 OIDC_TOKEN_INTROSPECTION_URL=
 OIDC_FALLBACK_MATCH_BY_EMAIL=
+
+# Required for Appointment's CalDAV auto-setup, needs to match the one in thunderbird-accounts
```
MelissaAutumn, line 146: I recommend leaving this blank in the example env.

backend/.env.example
```diff
@@ -87,6 +87,7 @@ TB_ACCOUNTS_HOST=http://localhost:8087
 TB_ACCOUNTS_CALLBACK=http://localhost:5000/accounts/callback
 TB_ACCOUNTS_CLIENT_ID
 TB_ACCOUNTS_SECRET
+TB_ACCOUNTS_CALDAV_URL=http://host.docker.internal:8081
```
MelissaAutumn, line 91: Important to note this won't work for all docker setups.

frontend/src/views/FTUEView/steps/ConnectYourCalendarStep.vue
```diff
@@ -28,9 +31,30 @@ const onContinueButtonClick = async () => {
   ftueStore.clearMessages();
 
   switch (calendarProvider.value) {
-    case 'oidc':
-      await ftueStore.moveToStep(FtueStep.ConnectCalendarsThundermail);
+    case 'oidc': {
+      isLoading.value = true;
```
MelissaAutumn, line 35: You know I don't know if this actually works within the same function call. I know we do it in other places but I don't think I've tested this. 

It might work since the await might trigger a dom update eventually, but might be worth a test with a really long delay.

davinotdavid: You are spot on, it does seem to work because of the await triggering a dom update and Vue batching updates. Tested here with a long delay and it updates the buttons accordingly!

Issue #1547, opened 2026-02-27
CheckboxInput not responsive in Settings


davinotdavid:
**Describe the bug**
CheckboxInput components that are not disabled can't be interacted with.

**To Reproduce**

1. Go to Settings
2. Click on any checkbox from any calendar that is not disabled
3. See that it can't be interacted with

**Expected behavior**
Checkbox should work as expected

**Actual behavior**
Checkboxes don't work

PR #1548, opened 2026-02-27
Fix CheckboxInput not working properly

davinotdavid:
## Description of the Change

There were upstream changes in the `CheckboxInput` from `services-ui` that check for the attribute `disabled` to be present even if its value is false. While we don't fix the problem there (https://github.com/thunderbird/services-ui/issues/230) I've added a workaround to be deployed as a hotfix.

I've changed all instances of `checkbox-input`. They are mostly concentrated in Availability page in Settings page.

## Benefits

- CheckboxInputs should work again!

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/1547


davinotdavid:
Thanks @devmount and @rwood-moz for the fast reviews 🙏

PR #1556, opened 2026-03-04
Add extra_hosts to use a CalDAV running in docker

paulopassosjr:
## Description of the Change

The initial setup requests to connect a CalDAV Calendar, one alternative is to run Radicale in docker but the URL cannot be http://localhost:5232/ because appointment is also running in docker. The solution is to use the IP address or http://host.docker.internal:5232/ with this PR.

image

## Benefits

Facilitate the initial setup of appointment.

## Applicable Issues

Closes https://github.com/thunderbird/appointment/issues/1557

image

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,6 +11,8 @@ services:
     environment:
       - IS_LOCAL_DEV=yes
       - RELEASE_VERSION=localdev
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
 
     depends_on:
       postgresql:


docker-compose.yml
```diff
@@ -11,6 +11,8 @@ services:
     environment:
       - IS_LOCAL_DEV=yes
       - RELEASE_VERSION=localdev
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
```
davinotdavid, line 15: Nice! TIL!

PR #1564, opened 2026-03-10
Replace legacy confirmation modals on settings page

### LINKED ISSUE #1540: Replace legacy modals for ConnectedApplications with ModalDialog from services-ui

Replace the following modals:

- CalDAV connect
- Google disconnect confirmation
- CalDAV disconnect confirmation
- Zoom disconnect confirmation

devmount:
## Description of the Change

This change replaces the legacy ConfirmationModal component with the ModalDialog from services-ui for

- connecting CalDAV
- disconnecting CalDAV
- disconnecting Google
- disconnecting Zoom

Can be tested on the Settings page by creating CalDAV, Google and Zoom connections and disconnect them again.

Before the change:
appointment_connect_caldav_old appointment_disconnect_caldav_old

After the change:
appointment_connect_caldav_new appointment_disconnect_caldav_new

## Benefits

Reduction of Tailwindcss-based component usage.

## Known issues

There are no specific designs for those modals yet (at least I didn't found them), so I implemented the modals as separate components by intention and kept the styling to the services-ui defaults. We can iterate on those in a next step.

## Applicable Issues

Closes #1540

### DIFF
--- a/frontend/src/views/SettingsView/components/CaldavConnectModal.vue
+++ b/frontend/src/views/SettingsView/components/CaldavConnectModal.vue
@@ -42,7 +42,7 @@ defineExpose({ show, hide })
 
     <div class="caldav-connect-modal-container">
       <notice-bar v-if="errorMessage" class="notice-bar" :type="NoticeBarTypes.Critical">
-        {{ errorMessage.title }}
+        {{ errorMessage.details || errorMessage.title }}
         <template #cta>
           <icon-button @click="errorMessage = null" :title="t('label.close')">
             <ph-x />
--- a/frontend/src/components/CalDavProvider.vue
+++ b/frontend/src/components/CalDavProvider.vue
@@ -2,6 +2,7 @@
 import { useI18n } from 'vue-i18n';
 import {
   computed, inject, ref,
+  useTemplateRef,
 } from 'vue';
 import { callKey } from '@/keys';
 import {
@@ -40,7 +41,7 @@ withDefaults(defineProps<Props>(), {
   showSwitch: false,
 });
 const emits = defineEmits(['next', 'error', 'previous', 'switch']);
-const formRef = ref();
+const formRef = useTemplateRef<HTMLFormElement>('formRef');
 
 /**
  * Submit the credentials, this will perform a dns check to see if there's any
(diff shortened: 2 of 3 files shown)


frontend/src/views/SettingsView/components/CaldavConnectModal.vue
```diff
@@ -0,0 +1,71 @@
...
+
+    <div class="caldav-connect-modal-container">
+      <notice-bar v-if="errorMessage" class="notice-bar" :type="NoticeBarTypes.Critical">
+        {{ errorMessage.title }}
+        <template #cta>
+          <icon-button @click="errorMessage = null" :title="t('label.close')">
+            <ph-x />
+          </icon-button>
+        </template>
+      </notice-bar>
```
davinotdavid, line 51: Hmmm it would be nice to confirm with the design team but perhaps it would be nice to move this above the header? I think this would require changes to `services-ui` to have a slot dedicated for a notice-bar so no worries about it as of now.

Image

frontend/src/views/SettingsView/components/CaldavConnectModal.vue
```diff
@@ -0,0 +1,71 @@
...
+      <cal-dav-provider @next="afterCalDavConnect()" @error="(alert: Alert) => errorMessage = alert" />
```
davinotdavid, line 52: I think the `errorMessage` needs a bit of extra handling here. Unsure if this was happening before but before the Alert was passed to the `GenericModal` which forwarded the error to the `AlertBox` component. The error I am getting for authentication error is:

Image

But I just got the generic "An unknown error occurred" message instead. Could you double check to see if this is a regression or if it is not, it can be an easy win? Otherwise we can open another issue for it.

frontend/src/views/SettingsView/components/CaldavConnectModal.vue
devmount: Yes, I like that idea! Created https://github.com/thunderbird/services-ui/issues/234

frontend/src/views/SettingsView/components/CaldavConnectModal.vue
devmount: Oh a great catch! I analyzed this and pushed a fix. This is a regression that happened earlier, when we switched to the NoticeBar component. Before that, we had a custom component showing both, the title and the reason. I changed it now to show the reason first, and if none is given, show the titel (generic default).

davinotdavid (approved):
Awesome! lgtm, thanks! 🎉

PR #1566, opened 2026-03-10
Replace legacy confirmation modal on subscriber admin page

### LINKED ISSUE #1542: Replace legacy modal "Subscriber panel hard delete" with ModalDialog from services-ui

devmount:
## Description of the Change

This change replaces the legacy ConfirmationModal component with the ModalDialog from services-ui for the delete subscriber confirmation.

This can be tested on the admin page by deleting an existing subscriber.

<details><summary>How to quickly add test subscribers</summary>
<p>

Test subscribers can be created by temporarily adding the following endpoint to api.py:

```py
@router.get('/test-create-account')
def test_create_account(db: Session = Depends(get_db)):
    """Used to create a test account"""
    if os.getenv('APP_ENV') != 'dev':
        raise HTTPException(status_code=405)
    if os.getenv('AUTH_SCHEME') != 'password':
        raise HTTPException(status_code=405)
    password='password'
    subscriber = repo.subscriber.create(db, schemas.SubscriberBase(
        email='mail@test.org',
        username='mail@test.org',
        name='Jane Doe',
        timezone='Europe/Berlin'
    ))
    # Update with password
    subscriber.password = utils.get_password_hash(password)
    db.add(subscriber)
    db.commit()
    db.refresh(subscriber)
    return subscriber
```

And then calling http://localhost:5000/test-create-account

</p>
</details> 

Before the change:
appointment_delete_subscriber_old

After the change:
appointment_delete_subscriber_new

## Benefits

Reduction of Tailwindcss-based component usage.

## Known issues

There is no specific design for this modal yet (at least I didn't found one), so I kept the styling to the services-ui defaults. We can iterate on this in a next step.

## Applicable Issues

Closes #1542

### DIFF
--- a/frontend/src/views/admin/SubscriberPanelView.vue
+++ b/frontend/src/views/admin/SubscriberPanelView.vue
@@ -29,10 +29,10 @@ const loading = ref(true);
 const pageError = ref<Alert>(null);
 const pageNotification = ref<Alert>(null);
 
-const deleteSubscriberModal = useTemplateRef('subscriberDeleteModalRef');
+const deleteSubscriberModalRef = useTemplateRef('subscriberDeleteModalRef');
 const hardDeleteModalContext = ref<number>(null);
 
-  const emailFilter = ref<string>(null);
+const emailFilter = ref<string>(null);
 const subscriberList = computed(() => {
   if (emailFilter.value) {
     /** @ts-expect-error escape has no baseline support yet */
@@ -169,7 +169,7 @@ const onFieldClick = (_key: string, field: any) => {
     toggleSubscriberState(field.id.value, field.timeDeleted.value === '');
   } else if (_key === 'hardDelete') {
     hardDeleteModalContext.value = field.id.value;
-    deleteSubscriberModal.value.show();
+    deleteSubscriberModalRef.value.show();
   }
 };
 


frontend/src/views/admin/SubscriberPanelView.vue
```diff
@@ -30,9 +28,11 @@ const displayPage = ref(false);
 const loading = ref(true);
 const pageError = ref<Alert>(null);
 const pageNotification = ref<Alert>(null);
-const hardDeleteModalOpen = ref<boolean>(false);
-const hardDeleteModalContext = ref<Subscriber>(null);
-const emailFilter = ref<string>(null);
+
+const deleteSubscriberModal = useTemplateRef('subscriberDeleteModalRef');
```
davinotdavid: This name conflicts with the import of the modal itself so perhaps we should append a `TemplateRef` to it?

frontend/src/views/admin/SubscriberPanelView.vue
```diff
@@ -30,9 +28,11 @@ const displayPage = ref(false);
...
+const hardDeleteModalContext = ref<number>(null);
+
+  const emailFilter = ref<string>(null);
```
davinotdavid: Sorry for the super nit here but can we also remove the extra spacing on the left here? 😅

frontend/src/views/admin/SubscriberPanelView.vue
devmount: Upsi 😇 fixed, thanks!

frontend/src/views/admin/SubscriberPanelView.vue
devmount: Definitely!

Issue #1470, opened 2026-01-28
Confirm booking links work without login


kewisch:
Maybe this is a feature, but...

STR:

1. Other person books slot
2. You receive the booking email, and forward to the person booking with you noting to reschedule
3. Other person clicks on links from the forwarded message

Actual:
* The confirm booking link does not require login or validate it belongs to the right user.
* The booking can be confirmed from the forwarded email

Expected:
* The user needs to be logged in to confirm the booking

PR #1577, opened 2026-03-16
Prevent users other than Subscriber to confirm / deny appointment through email links

davinotdavid:
## Description of the Change

[Backend changes]
- The route `PUT public/availability/booking` now requires authentication and a match of the owner extracted from the `owner_url` with the logged in subscriber. Returns `Unauthorized` or `InvalidLinkException` accordingly.

[Frontend changes]
- In the `router.ts` there's a `beforeEach` guard that redirects the user to the OIDC login. However, in that brief moment where we are attempting to render some UI for the page, it always defaulted to the `NotFoundView` so users saw a "flash" of that before the redirect which is not great. Also, after logging in, they were not redirected to the URL that they wanted to go (they were always redirected to the Dashboard). This was necessary to be done so that an unauthenticated Subscriber can click on a confirm / deny link and, after successfully login in, go straight into the `BookingConfirmationView` instead of Dashboard.
- In the `onMounted` of the `BookingConfirmationView`, if there's a logged in user but doesn't match the owner of the URL, we show the invalid link error.

## How to Test

Preferably, test this with `AUTH_SCHEME=oidc` instead of `password` to match the actual stage / prod environment. Should work with password too though!

[Case 1 - Not logged in]
- Go through the FTUE with User A
- Make sure that in Availability the setting `Automatically confirm bookings if time is available` is OFF / Unchecked
- Copy the Booking URL and paste it in an incognito window
- Book any slot with any name / email
- Go to Mailpit and copy User A's "confirm" or "deny" links from the buttons
- Open another browser / incognito window and paste it
- See that you are redirected to login

[Case 2 - Logged in as a different user]
- Go through the FTUE with User A
- Make sure that in Availability the setting `Automatically confirm bookings if time is available` is OFF / Unchecked
- Copy the Booking URL and paste it in an incognito window
- Book any slot with any name / email
- Go to Mailpit and copy User A's "confirm" or "deny" links from the buttons
- Open another browser / incognito window and paste it
- See that you are redirected to login
- Login with User B
- See that you are shown the Invalid Link error

[Case 3 - Logged in as the owner of the confirm / deny link]
- Clicking confirm / deny should work as usual

## Benefits

- Subscribers that might forward the confirm / deny booking email to someone else or if the confirm / deny links leak by any chance, other people won't be able to confirm / deny the booking.

## Related tickets

Fixes https://github.com/thunderbird/appointment/issues/1470

### DIFF
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -135,7 +135,7 @@ const scheduleStore = createScheduleStore(call);
 
 // true if route can be accessed without authentication
 const routeIsPublic = computed(
-  () => route.meta?.isPublic,
+  () => !!route.meta?.isPublic,
 );
 const routeIsHome = computed(
   () => ['home'].includes(typeof route.name === 'string' ? route.name : ''),
@@ -154,6 +154,11 @@ const routeIsPostLogin = computed(
 );
 const routeIsLogout = computed(
   () => ['logout'].includes(typeof route.name === 'string' ? route.name : ''),
+);
+// The router.ts file has a beforeEach guard that redirects to the OIDC login
+// but we are flashing a NotFoundView briefly which is uh... not ideal 
+const routeNameUndefinedDueToAbortedNav = computed(
+  () => !route.name && !user?.authenticated && isOidcAuth,
 );
 
 // retrieve calendars and appointments after checking login and persisting user to db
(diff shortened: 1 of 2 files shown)


MelissaAutumn:
I think this was intentionally not protected to allow for folks who aren't logged in to simply accept a meeting request from their email. 

Just wanted to check in with @ryanleesipes before we merged, who I think requested this to be a feature in the first place.

kewisch:
I've checked with Ryan, we're ok to proceed gating this behind login. Originally it was because we had trouble retaining a logged in session, and it was annoying to have to log in for each request. Now that we have longer timeouts we'll be fine.

davinotdavid:
Thanks @kewisch ! Really appreciate the follow through!

PR #1578, opened 2026-03-16
Remove tailwindcss

### LINKED ISSUE #495: Remove tailwind
Now that we're getting some newer designs for pages, components, and flows we should think about moving away from tailwind. It was nice to have while prototyping but it's becoming a burden to override in some cases (the form package especially.)

This is a large task, but we can work on it in parts by writing new styles in <style scoped> tags and using main.css some more. We'll keep tailwind until we've moved everything over.

devmount:
## Description of the Change

It is finally time: Tailwindcss will be completely gone with this PR.

## Benefits

No magic usability classes anymore. Improved dev and maintenance experience.

## Screenshots

None. I hope everything still looks the same 😅 

To keep the default behaviour of Tailwind, I added a slightly modified [`preflight.css`](https://github.com/tailwindlabs/tailwindcss/blob/main/packages/tailwindcss/preflight.css) to our CSS stack. We can include those inside our main.css or keep it separate - let me know what you prefer.

## Related tickets

Closes #495

### DIFF
--- a/frontend/src/views/admin/SubscriberPanelView.vue
+++ b/frontend/src/views/admin/SubscriberPanelView.vue
@@ -4,6 +4,7 @@ import { useI18n } from 'vue-i18n';
 import { useRouter } from 'vue-router';
 import AlertBox from '@/elements/AlertBox.vue';
 import DataTable from '@/components/DataTable.vue';
+import LoadingSpinner from '@/elements/LoadingSpinner.vue';
 import DeleteSubscriberModal from './components/DeleteSubscriberModal.vue';
 import { TextInput } from '@thunderbirdops/services-ui';
 import { dayjsKey, callKey } from '@/keys';
@@ -221,6 +222,9 @@ onMounted(async () => {
 
     </data-table>
   </div>
+  <div v-else class="panel-loading">
+    <loading-spinner/>
+  </div>
 
   <!-- Refresh link confirmation modal -->
   <delete-subscriber-modal
@@ -231,7 +235,8 @@ onMounted(async () => {
 </template>
 
 <style scoped>
-.panel-alerts {
+.panel-alerts,
+.panel-loading {
   display: flex;
   justify-content: center;
   width: 100%;


frontend/src/router.ts
```diff
@@ -122,9 +123,8 @@ const routes: RouteRecordRaw[] = [
   {
     path: '/contact',
     name: 'contact',
-    component: ContactView,
-    meta: {
-      isPublic: true,
+    component: () => {
+      window.location.href = redirectUrl;
     },
```
davinotdavid, line 128: Nice!

frontend/src/views/admin/SubscriberPanelView.vue
```diff
@@ -222,9 +221,6 @@ onMounted(async () => {
 
     </data-table>
   </div>
-  <div v-else class="flex size-full min-h-[75vh] items-center justify-center">
-    <loading-spinner/>
-  </div>
```
davinotdavid, line 227: Perhaps we can leave the `loading-spinner` here? I honestly don't mind if it is not styled but might as well translate that to vanilla CSS while we are at it.

frontend/src/assets/styles/preflight.css
```diff
@@ -0,0 +1,370 @@
+/* Source: https://github.com/tailwindlabs/tailwindcss/blob/main/packages/tailwindcss/preflight.css */
```
davinotdavid, line 1: This looks like a nice CSS reset overall but we probably don't need some of the rules here. Also, it would be nice if we could author this ourselves and merge it with some of the declarations from `main.css`.

We can definitely keep it for now though, specially since this PR is a big change!

frontend/src/elements/arts/ArtSuccessfulBooking.vue
```diff
@@ -1,6 +1,16 @@
-<script setup lang="ts"></script>
+<script setup lang="ts">
+import { PhConfetti } from '@phosphor-icons/vue';
+</script>
 
 <template>
-  <!-- Pending new illustration -->
-  <div class="w-full text-center text-[10rem]">🎉</div>
+  <div>
+    <ph-confetti class="art-successful-icon" weight="thin" />
```
davinotdavid, line 7: I think we need to explicitly add a color here now, perhaps `--colour-primary-default` ?

This is the case of confirming a booking from the BookingSlidePanel in the Bookings page. The case from the BookingConfirmationView works well though 🤔 

Image

frontend/src/views/admin/SubscriberPanelView.vue
devmount: Alright, I added it again 👍🏻

frontend/src/elements/arts/ArtSuccessfulBooking.vue
devmount: Great catch! I fixed it

frontend/src/assets/styles/preflight.css
devmount: Agreed. I added #1581 for this!

PR #1599, opened 2026-03-30
Optimize CSS reset

### LINKED ISSUE #1581: Optimize CSS reset
**Describe the bug**
With #1578 a `preflight.css` was introduced to make a smooth transition from Tailwindcss to vanilla CSS. We should investigate, which rules we don't need and merge the rest with our main.css.

**Additional context**
@davinotdavid [stated](https://github.com/thunderbird/appointment/pull/1578#discussion_r2949132867):

> This looks like a nice CSS reset overall but we probably don't need some of the rules here. Also, it would be nice if we could author this ourselves and merge it with some of the declarations from main.css.

devmount:
## Description of the Change

This change moves all rules from `preflight.css` to main.css that still have an impact to our current codebase. Styles that are already taken care of in Services UI were removed. Also some more clean up was done which is now possible after the Removal of TailwindCSS.

## Benefits

Less styles scattered around.

## Screenshots

No visual changes. In the best case 😇 

Not that I know of. Of course there could be style glitches I missed, but since we're relying on Services UI already a lot, those should occur only very rarely if at all.

## Related tickets

Closes #1581

### DIFF
--- a/frontend/eslint.config.js
+++ b/frontend/eslint.config.js
@@ -46,7 +46,6 @@ export default [
       // Disable full warning, and customize the typescript one
       // Warn about unused vars unless they start with an underscore
       'no-unused-vars': 'off',
-      '@typescript-eslint/no-require-imports': 'off', // can be removed when we got rid of Tailwind or updated to v4
       '@typescript-eslint/no-unused-vars': [
         'warn',
         {


frontend/src/views/BookingsView/components/AppointmentSlidingPanelModify.vue
```diff
@@ -158,21 +158,21 @@ defineExpose({
...
-            data-testid="appointment-modal-modify-notes-input">
-          </textarea>
-        </label>
+        <text-area
+          name="notes"
+          :label="t('label.notes')"
+          v-model="form.notes"
+          :placeholder="t('placeholder.bookingNotesPlaceholder')"
+          data-testid="appointment-modal-modify-notes-input"
+        />
```
MelissaAutumn, line 175: Was this an intentional change? Looks fine, just wasn't mentioned 😄

devmount: Ah sorry, yes, this was necessary because it was the only occurrence of a native textarea element! Everything else comes directly from Services UI. Forgot to mention this 😇

PR #1601, opened 2026-03-30
Fix E2E nightly jobs Part 2: Add new E2E workflows

### LINKED ISSUE #1594: Appointment Nightly E2E test jobs aren't running automatically
In #1585 I separated the nightly E2E test desktop and mobile workflows into separate workflows for each platform/browser so that we can manually trigger a test job via the actions menu for only one single platform. Manually triggering each of the jobs works fine (actions tab, select one of the nightly-tests* jobs, and click to run on main); however the actual scheduled nightly test jobs themselves stopped running after that change.

Apparently sometimes after renaming a workflow it can take awhile for it to be recognized / scheduled again so I was hoping after a follow-up PR had landed then the jobs would startup again, but that's not the case. Investigate why I can manually trigger the nightly test workflows successfully but they won't run automatically at their scheduled (cron) time.

rwood-moz:
Fixes #1594. Old E2E nightly test workflows were removed in #1600, now add new workflows for the nightly E2E test jobs (hopefully this will fix the GHA scheduling issue).

### DIFF
--- a/.github/workflows/e2e-desktop-nightly-chrome.yml
+++ b/.github/workflows/e2e-desktop-nightly-chrome.yml
@@ -28,11 +28,11 @@ jobs:
       APPT_MY_SHARE_LINK: ${{ secrets.E2E_APPT_PROD_MY_SHARE_LINK }}
       APPT_BOOKEE_EMAIL: ${{ secrets.E2E_APPT_PROD_BOOKEE_EMAIL }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
 
-      - uses: actions/setup-node@v4
+      - uses: actions/setup-node@v6
         with:
-          node-version: 20
+          node-version: 24
           cache: 'npm'
           cache-dependency-path: 'test/e2e/package-lock.json'
 
--- a/.github/workflows/e2e-desktop-nightly-firefox.yml
+++ b/.github/workflows/e2e-desktop-nightly-firefox.yml
@@ -28,11 +28,11 @@ jobs:
       APPT_MY_SHARE_LINK: ${{ secrets.E2E_APPT_PROD_MY_SHARE_LINK }}
       APPT_BOOKEE_EMAIL: ${{ secrets.E2E_APPT_PROD_BOOKEE_EMAIL }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
 
-      - uses: actions/setup-node@v4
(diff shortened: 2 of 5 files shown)


.github/workflows/e2e-desktop-nightly-chrome.yml
```diff
@@ -0,0 +1,58 @@
...
+      name: production
+      deployment: false
+    env:
+      TB_ACCTS_EMAIL: ${{ secrets.E2E_APPT_PROD_TB_ACCTS_EMAIL }}
+      TB_ACCTS_PWORD: ${{ secrets.E2E_APPT_PROD_TB_ACCTS_PWORD }}
+      APPT_DISPLAY_NAME: ${{ secrets.E2E_APPT_PROD_DISPLAY_NAME }}
+      APPT_MY_SHARE_LINK: ${{ secrets.E2E_APPT_PROD_MY_SHARE_LINK }}
+      APPT_BOOKEE_EMAIL: ${{ secrets.E2E_APPT_PROD_BOOKEE_EMAIL }}
+    steps:
+      - uses: actions/checkout@v4
```
devmount: We might want to got to v6 here:
https://github.com/actions/checkout

.github/workflows/e2e-desktop-nightly-chrome.yml
```diff
@@ -0,0 +1,58 @@
...
+
+      - uses: actions/setup-node@v4
```
devmount: This action is also already on v6:
https://github.com/actions/setup-node

.github/workflows/e2e-desktop-nightly-chrome.yml
```diff
@@ -0,0 +1,58 @@
...
+        with:
+          node-version: 20
```
devmount: Are we on v20 intentionally? At some point we should go to 22 or 24, or what do you think?

.github/workflows/e2e-desktop-nightly-chrome.yml
rwood-moz: Good point, I'll update that (and `setup-node` below too) thank you

.github/workflows/e2e-desktop-nightly-chrome.yml
rwood-moz: Yes good idea, I'm running `24` locally and it seems to work ok, so I'll give it a try.

devmount:
@rwood-moz Awesome, thanks so much!

PR #1604, opened 2026-03-30
Fix E2E nightly GHA workflow typos

rwood-moz:
Follow-up to #1601.


rwood-moz:
> Good catch!

Thank you :)

PR #1610, opened 2026-04-02
Bump version to 1.3.0

davinotdavid:
It's been a while to deploy Appointment to prod so let's do it!

Last release was on February 27 2026!
https://github.com/thunderbird/appointment/releases/tag/r-0772

### DIFF
--- a/backend/uv.lock
+++ b/backend/uv.lock
@@ -1,5 +1,5 @@
 version = 1
-revision = 3
+revision = 2
 requires-python = ">3.12"
 resolution-markers = [
     "python_full_version >= '3.14'",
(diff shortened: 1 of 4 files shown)


backend/uv.lock
```diff
@@ -1,5 +1,5 @@
 version = 1
-revision = 3
```
kewisch, line 2: Specific reasons to downgrade here? Are the remaining package changes in this file a consequence of the lock file version change?

davinotdavid: hmm that was a result of a fresh run of a `uv sync` on my machine but not strictly necessary so I can revert the uv.lock just in case

aatchison: still need to address this? otherwise lgtm

aatchison (approved):
bump it

backend/uv.lock
MelissaAutumn: I believe this is due to uv version variance. It's a revision of the lockfile format not the contents iirc. Should be good though.

MelissaAutumn (approved):

MelissaAutumn:
Bit late for a deploy, but lgtm!

Issue #1615, opened 2026-04-08
Add Flower for queue visibility


davinotdavid:
Now that we have a basic Celery setup up, we should also add Flower for queue visibility!

ryanjjung:
Both this ticket and #1611 will have to be slightly redesigned. Leaving details in that ticket.

Issue #1611, opened 2026-04-02
Deploy Celery workers for Appointment


ryanjjung:
## Background

Appointment demands some background workers. We'll do that with Celery. Some of this has landed in local dev already, but we need to get prepared for deployment in our live environments.

## Deployment Details

For each of stage and prod:

- Deploy a new `AutoscalingFargateCluster` for this purpose.
- Install `celery` and `flower` services using the images made for Celery to run in.
- Set up communication between the publisher (the Appointment containers), the consumer (Celery), the monitor (Flower), and the Redis cluster they all talk over. Use a custom channel for communication so there is no clobbering with other uses Appointment has for Redis.

ryanjjung:
Redis for this application is an AWS Serverless Redis instance, which only works in cluster mode. Celery doesn't support clustered Redis at all, period. There's an [open PR](https://github.com/celery/kombu/pull/1021) that has been touched as recently as three weeks ago. I found this via a [discussion about this problem](https://github.com/celery/kombu/pull/1021) that says this is to be released in v5.6. However, the latest version I can pull with pip is 5.6.3 and the feature isn't there. So we need to investigate another way to get Celery working.

Seems the simplest solution is to build an SQS queue alongside this and route tasks through that. Configuration for this [is supported](https://docs.celeryq.dev/en/stable/getting-started/backends-and-brokers/sqs.html#predefined-queues). Will take a small amount of reconfiguration here.

PR #1617, opened 2026-04-08
Add Flower alongside live environment Celery workers

ryanjjung:
This PR adds Celery workers to our live environments alongside Flower for worker visibility.

## What's in the PR

- Flower added to local dev configuration
- Added explicit `CONTAINER_ROLE` variable value: "api" to indicate the regular backend as opposed to Celery, Flower, or something else.
- Built new AutoscalingFargateCluster with Celery and Flower running in dev, stage, and prod.
- Restructured all of our Pulumi configurations regarding task and container definitions to be more "DRY". This is because we now have three task definitions (which tend to be large/unwieldy) that are all 99% the same. So now we have variables that get reused across task/container defs that themselves get reused in our cluster configs.
- Allow the Celery and Flower containers in dev and stage to talk to the Redis cluster.

## What's Not in the PR

I have intentionally left the new prod Celery and Flower services scaled down to zero instances. This is because they will not work if brought online. And *that* is because we have to allow these containers access to the Neon DB PrivateLink security group (which is manual), and to the Redis cluster (which will be codified after the prod security groups have been created).

So there will be a small future PR coming after this is fully deployed.

**Relevant Tickets**
- #1611
- #1615

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,6 +9,7 @@ services:
       - ./backend/.env:/app/.env
       - ./backend/src/appointment:/app/appointment
     environment:
+      - CONTAINER_ROLE=api
       - IS_LOCAL_DEV=yes
       - RELEASE_VERSION=localdev
     extra_hosts:


docker-compose.yml
```diff
@@ -36,6 +36,17 @@ services:
       - backend
       - redis
 
+  celery-flower:
+    <<: *backend
+    ports:
+      - 5555:5555
```
davinotdavid: As a reminder to myself as well, can we switch this to `5556:5555` instead? It is becoming more common for me to run both `thunderbird-accounts` and `appointment` locally so the ports here would collide with ports there.

```suggestion
      - 5556:5555
```

ryanjjung: Yeah, I'm about to sit back down to this. Just making sure, but this doesn't need to be changed in live environments, right? Only local dev?

Also, doesn't @MelissaAutumn have a page somewhere where we track ports? I should add this to the list.

.github/workflows/deploy-production.yml
```diff
@@ -98,17 +98,10 @@ jobs:
...
-          cat << EOF > newimage.yaml
-          resources:
-            tb:fargate:FargateClusterWithLogging:
-              backend:
-                task_definition:
-                  container_definitions:
-                    backend:
-                      image: "${{ steps.pulumi-tag-extract.outputs.pulumi_tag }}"
-          EOF
+          echo ".apmt_image: &APMT_IMAGE ${{ steps.pulumi-tag-extract.outputs.pulumi_tag }}" > newimage.yaml
```
ryanjjung, line 104: This is simpler now due to the DRYing out of the task definitions.

.github/workflows/deploy-production.yml
```diff
@@ -118,8 +111,9 @@ jobs:
           pulumi login
           pulumi stack select prod
           TBPULUMI_DISABLE_PROTECTION=True \
-            pulumi up -y --diff --target \
-            'urn:pulumi:prod::appointment::tb:fargate:FargateClusterWithLogging$aws:ecs/taskDefinition:TaskDefinition::appointment-prod-fargate-backend-taskdef' \
+            pulumi up -y --diff \
+            --target 'urn:pulumi:stage::appointment::tb:fargate:FargateClusterWithLogging$aws:ecs/taskDefinition:TaskDefinition::appointment-stage-fargate-backend-taskdef' \
+            --target 'urn:pulumi:stage::appointment::tb:fargate:AutoscalingFargateCluster::appointment-stage-afc-appointment' \
```
ryanjjung: This makes sure that images get deployed to both clusters when we do a release.

pulumi/config.prod.yaml
```diff
@@ -1,9 +1,137 @@
 ---
+
+### Special variables used throughout this file
+
+# Update this value to update all containers based on the thunderbird/appointment image
+.apmt_image: &APMT_IMAGE 768512802988.dkr.ecr.eu-central-1.amazonaws.com/thunderbird/appointment:7de6f16bdd309937caa186c8f5a269ea00118e5e
```
ryanjjung, line 6: This is the variable referenced in the workflow changes. This gets used in the task definition below, and that task definition gets used for all three services. So you change this in one place and it goes out to All The Things.

pulumi/config.prod.yaml
```diff
@@ -1,9 +1,137 @@
...
+
+
 resources:
 
   domains:
     redis: cache.appointment.tb.pro
 
+  tb:cloudwatch:LogDestination:
+    appointment:
+      org_name: tb
```
ryanjjung, line 133: This creates a privacy policy compliant CloudWatch Log Group called `/tb/prod/appointment` that these containers will now produce logs in.

pulumi/config.prod.yaml
```diff
@@ -75,6 +203,16 @@ resources:
               protocol: tcp
               from_port: 6380
               to_port: 6380
+            # - description: Allow Redis traffic from Celery
+            #   protocol: tcp
+            #   from_port: 6379
+            #   to_port: 6379
+            #   source_security_group_id:
```
ryanjjung, line 210: This is where I will add the security groups to grant access to Redis later on.

pulumi/config.prod.yaml
```diff
@@ -349,8 +625,11 @@ resources:
       enable_fargate_deployments: True
       fargate_clusters:
         - appointment-prod-fargate-backend
+        - appointment-prod
       fargate_task_role_arns:
         - arn:aws:iam::768512802988:role/appointment-prod-fargate-backend
+        - arn:aws:iam::768512802988:role/appointment-prod-afc-appointment-celery
+        - arn:aws:iam::768512802988:role/appointment-prod-afc-appointment-flower
```
ryanjjung, line 632: These represent new permissions needed for the CI process to deploy to the new cluster and tasks.

pulumi/requirements.txt
```diff
@@ -1,3 +1,3 @@
-tb_pulumi @ git+https://github.com/thunderbird/pulumi.git@v0.0.16
+tb_pulumi @ git+https://github.com/thunderbird/pulumi.git@v0.0.18
```
ryanjjung, line 1: This contains many fixes to both the `LogDestination` class and `AutoscalingFargateCluster` class that we need for these new resources to come out right. Ref: https://github.com/thunderbird/pulumi/blob/main/CHANGELOG.md

pulumi/security_groups.py
```diff
@@ -69,7 +69,8 @@ def security_groups(
     )
     backend_cache_sg_ingress_rules = backend_cache_sg_opts.get('rules', {}).get('ingress', {})
     for rule in backend_cache_sg_ingress_rules:
-        rule['source_security_group_id'] = container_sgs.get('backend').resources.get('sg').id
+        if 'source_security_group_id' not in rule and 'cidr_blocks' not in rule:
+            rule['source_security_group_id'] = container_sgs.get('backend').resources.get('sg').id
```
ryanjjung, line 73: Normally, the way we use these SGs in code is to automatically link up the Appointment backend container to Redis. If we want to link up any other source, we would need to specify that in code or config somewhere. Rather than bog this code down in lots of conditions based on expected strings in the config, I've changed this to allow us to specify a source in config, and to fall back on the Appointment backend container if no more explicit source is defined.

docker-compose.yml
```diff
@@ -36,6 +36,17 @@ services:
       - backend
       - redis
 
+  celery-flower:
+    <<: *backend
+    ports:
+      - 5556:5555
```
ryanjjung, line 43: Davi requested this port exposure change so that this does not conflict with a simultaneously running Flower container for a local dev instance of Accounts.

.github/workflows/deploy-production.yml
davinotdavid: This is currently targetting `appointment-stage-*` but shouldn't it be `appointmento-prod-*`?

docker-compose.yml
davinotdavid, line 11: I think we need to add a `CONTAINER_ROLE=api` here since we've changed the `entry.sh`

pulumi/config.stage.yaml
```diff
@@ -313,6 +356,240 @@ resources:
...
+                - *VAR_SMTP_SECURITY
+                - *VAR_TB_ACCOUNTS_CALDAV_URL
+                - *VAR_TB_ACCOUNTS_HOST
+                - *VAR_TIER_BASIC_CALENDAR_LIMIT
+                - *VAR_TIER_PLUS_CALENDAR_LIMIT
+                - *VAR_TIER_PRO_CALENDAR_LIMIT
+                - *VAR_ZOOM_API_ENABLED
+                - *VAR_ZOOM_API_NEW_APP
+                - name: CONTAINER_ROLE
+                  value: celery
```
davinotdavid: Same here, shouldn't this be `beat` / `worker` ? If so, I wonder if we need 2 sets of this configs, one for each?

pulumi/config.prod.yaml
```diff
@@ -312,6 +354,240 @@ resources:
...
+                - *VAR_SMTP_SECURITY
+                - *VAR_TB_ACCOUNTS_CALDAV_URL
+                - *VAR_TB_ACCOUNTS_HOST
+                - *VAR_TIER_BASIC_CALENDAR_LIMIT
+                - *VAR_TIER_PLUS_CALENDAR_LIMIT
+                - *VAR_TIER_PRO_CALENDAR_LIMIT
+                - *VAR_ZOOM_API_ENABLED
+                - *VAR_ZOOM_API_NEW_APP
+                - name: CONTAINER_ROLE
+                  value: celery
```
davinotdavid: Same here, just double checking!

.github/workflows/deploy-production.yml
ryanjjung: Yes, I copied those things in and didn't scan for the environment name. I'll push a fix right now.

pulumi/config.stage.yaml
ryanjjung: I just fixed "celery" to "worker". As far as I'm aware, the beat container worker doesn't need to be deployed to live environments. That will be ultimately removed in favor of real tasks, right?

pulumi/config.prod.yaml
ryanjjung: I fixed this in all three files just now.

pulumi/config.stage.yaml
davinotdavid: There's going to be a task [in another PR](https://github.com/thunderbird/appointment/pull/1529) that needs to be run periodically every week in production as well so perhaps we still need the beat container there too? Not sure if we need a separate container though as @Sancus added something called celery-redbeat to Accounts with a single container (ref https://github.com/thunderbird/thunderbird-accounts/pull/696) so maybe that's also an option

ryanjjung: I had not heard of redbeat, so I went and found [its docs](https://redbeat.readthedocs.io/en/latest/config.html). It looks like you just configure your normal Celery container with this and a Redis key to use for its distributed lock. Then you [define tasks](https://redbeat.readthedocs.io/en/latest/tasks.html) as part of your Celery config. So it sounds to me like we wouldn't need a separate container to emit these events on a schedule.

docker-compose.yml
ryanjjung: I pushed this change this morning :)

ryanjjung:
Merged, will keep an eye on the workflows and follow through with the prod work now.

PR #1622, opened 2026-04-15
Update stage image; fix boolean -> string in all envs

ryanjjung:
When we merged [this PR](https://github.com/thunderbird/appointment/pull/1617), [this deploy to stage](https://github.com/thunderbird/appointment/actions/runs/24460284235/job/71472228939) failed with this error:

> error: aws:ecs/taskDefinition:TaskDefinition resource '[secret]-stage-afc-[secret]-taskdef-flower' has a problem: ECS Task Definition container_definitions is invalid: json: cannot unmarshal bool into Go struct field KeyValuePair.Environment.Value of type string. Examine values at '[secret]-stage-afc-[secret]-taskdef-flower.containerDefinitions'.

That's because I was missing some quotes around some "true" values in the Flower config. That's fixed here. I was able to reproduce the problem locally and then fix it with these changes locally.

### DIFF
--- a/pulumi/config.dev.yaml
+++ b/pulumi/config.dev.yaml
@@ -384,7 +384,7 @@ resources:
                 - name: CONTAINER_ROLE
                   value: flower
                 - name: FLOWER_UNAUTHENTICATED_API
-                  value: true
+                  value: "true"
       
       targets:
         flower:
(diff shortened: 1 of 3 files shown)


davinotdavid (approved):
lgtm, thanks!

PR #1631, opened 2026-04-16
Build a Redis cache replica set to migrate to

ryanjjung:
This PR builds new Redis replica sets which we will soon migrate the Appointment backend containers to. It also outlines through commentary what the next steps are here.

Just to create some parity, I fleshed out the upper variable section of the dev config as well based on the recent stage config changes, but I haven't gone out and set all the secrets and updated the task definitions yet. That's some tedious work that's irrelevant to this issue.

I put a condition in around DNS in prod so that rolling these changes out will not result in DNS getting swapped out before I'm ready to make that change. However, with this PR, we will get these changes in prod (summary of preview):

- New log destination
- Appointment app logs go to that log destination
- New Redis cache gets built (in a way that I've verified in stage works with both the backend API and Celery/Flower)

After approval of this PR, I'll merge and apply these changes to prod. Then I will prep the next (much smaller) PR to swap DNS over. I will force a rollout of the services without changing the image and will verify both the continued functionality of the app and that we see a corresponding shift in load through the CloudWatch metrics for the caches.

Ref: #1611

### DIFF
--- a/pulumi/config.dev.yaml
+++ b/pulumi/config.dev.yaml
@@ -231,14 +231,6 @@ resources:
       engine: redis
       major_engine_version: "7"
 
-  tb:elasticache:ElastiCacheReplicaGroup:
-    backend:
-      num_cache_nodes: 1
-      automatic_failover_enabled: False
-      transit_encryption_mode: required
-      transit_encryption_enabled: True
-      apply_immediately: True
-  
   tb:fargate:FargateClusterWithLogging:
     backend:
       assign_public_ip: True
--- a/pulumi/config.dev.yaml
+++ b/pulumi/config.dev.yaml
@@ -543,6 +543,14 @@ resources:
           min_capacity: 1
           max_capacity: 1
 
+  # Additional properties for this environment's cache
+  tb:elasticache:ElastiCacheReplicationGroup:
+    backend:
+      num_cache_nodes: 1
+      # Settings below should be "yes" when (num_cache_nodes > 1)
+      automatic_failover_enabled: no
+      multi_az_enabled: no
+
   tb:cloudfront:CloudFrontS3Service:
     frontend:
       service_bucket_name: tb-appointment-dev-frontend
--- a/pulumi/config.dev.yaml
+++ b/pulumi/config.dev.yaml
@@ -8,16 +8,20 @@
 # These variables are common to Accounts application environments. Some tasks will require additional configuration.
 .app_env: &VAR_APP_ENV {name: "APP_ENV", value: "dev"}
 .auth_scheme: &VAR_AUTH_SCHEME {name: "AUTH_SCHEME", value: "oidc"}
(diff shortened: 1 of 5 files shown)


ryanjjung:
Holding off. Need to fix a few features on Redis.

pulumi/config.dev.yaml
```diff
@@ -10,14 +10,17 @@
...
 .database_port: &VAR_DATABASE_PORT {name: "DATABASE_PORT", value: "5432"}
+.frontend_url: &VAR_FRONTEND_URL {name: "FRONTEND_URL", value: "https://appointment-dev.tb.pro"}
+.fxa_callback: &VAR_FXA_CALLBACK {name: "FXA_CALLBACK", value: "https://appointment-dev.tb.pro/fxa"}
+.google_auth_callback: &VAR_GOOGLE_AUTH_CALLBACK {name: "GOOGLE_AUTH_CALLBACK", value: "https://appointment-dev.tb.pro/api/v1/google/callback"}
 .jwt_algo: &VAR_JWT_ALGO {name: "JWT_ALGO", value: "HS256"}
 .jwt_expire_in_mins: &VAR_JWT_EXPIRE_IN_MINS {name: "JWT_EXPIRE_IN_MINS", value: "10000"}
 .log_level: &VAR_LOG_LEVEL {name: "LOG_LEVEL", value: "ERROR"}
 .log_use_stream: &VAR_LOG_USE_STREAM {name: "LOG_USE_STREAM", value: "True"}
-.oids_exp_grace_period: &VAR_OIDC_EXP_GRACE_PERIOD {name: "OIDC_EXP_GRACE_PERIOD", value: "60"}
+.oidc_exp_grace_period: &VAR_OIDC_EXP_GRACE_PERIOD {name: "OIDC_EXP_GRACE_PERIOD", value: "60"}
```
davinotdavid, line 20: I think we should also change this to `oidc_exp_grace_period` for stage and prod for consistency

mzeier (approved):
Limited context aside,

1. Dev cuts over immediately along with stage. Uses `if project.stack == 'prod' else redis_replica_group_primary_endpoint` → maybe intentional.

1. Empty valueFrom: "" secret placeholders (`SECRET_APP_ADMIN_ALLOW_LIST`, `SECRET_DB_SECRET`, `SECRET_FXA_*`, etc). Maybe harmless.

davinotdavid (approved):
LGTM after comment!

pulumi/config.dev.yaml
ryanjjung: Did I miss it? I'll double-check.

ryanjjung:
@mzeier dev isn't an environment that normally exists, and I tested the cutover procedure there yesterday. The reason the variables are not used is because this env is not part of our normal deployment flows and we don't run fully working services here; normally just testing infra changes. The values are empty because I would otherwise have to go and create all of those secrets, which is tedious work that didn't progress this issue at all, so I opted not to go that far with the work. And harmless because those variables are not referenced in the container definitions.

PR #1632, opened 2026-04-17
Fix "invalid grant" error during Google Oauth

### LINKED ISSUE #1629: "Internal Server Error" when adding a Google calendar
## Describe the bug

I'm trying to add my work Google calendar, but after completing the Google authentication/authorization flow I'm getting this page:

Image

## To Reproduce

1. During onboarding, try adding a Google calendar
2. Authenticate and tell Google to give Appointment all the permissions it asks for
3. Confirm and wait to be redirected back to Appointment

## Expected behavior

I'm able to continue/finish the onboarding.

## Actual behavior

The "Internal Server Error" page shown above appears.

## Screenshots

See above

## System

- OS: ArchLinux (btw)
- Browser Version: Firefox 151.0a1 (2026-04-16)

## Additional context

None

davinotdavid:
## What changed?

`backend/src/appointment/routes/google.py`
- We are now storing the OAuth `code_verifier` in the session object.

`backend/src/appointment/controller/apis/google_client.py`
- Instead of re-using an in-memory `Flow` instance for the OAuth exchange, we are now creating a fresh one. Paired with storing the `code_verifier` in the session object, we can make sure that the `Flow` object keeps the same `code_verifier` throughout the OAuth process.

## Why?

The error that we were getting back was "Invalid grant" which seems to happen due to a PKCE (Proof Key for Code Exchange) error on the `code_verifier. [Stage and Prod share the same Google Cloud app / project](https://console.cloud.google.com/apis/credentials?authuser=2&project=thunderbird-appointment), which is puzzling on why stage works well and prod doesn't or is at best intermittent.

Example issue in Sentry:
https://thunderbird.sentry.io/issues/7302416178/?environment=prod&project=4505428124827648&query=is%3Aunresolved&referrer=issue-stream

One possible explanation is that since we are storing the `Flow` instance in memory, the request to start the OAuth and the callback request happen in two different Fargate instances so the `code_verifier` check fails when attempting to get credentials.

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/1629

### DIFF
--- a/backend/src/appointment/controller/apis/google_client.py
+++ b/backend/src/appointment/controller/apis/google_client.py
@@ -25,7 +25,6 @@ class GoogleClient:
         'https://www.googleapis.com/auth/userinfo.email',
         'openid',
     ]
-    client: Flow | None = None
 
     def __init__(self, client_id, client_secret, project_id, callback_url):
         self.config = {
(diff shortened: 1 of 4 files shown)


devmount (approved):
Thanks so much Davi!

Code looks good and I can confirm, that I can add Google calendars again locally (didn't work on main, so I could reproduce the issue).

PR #1633, opened 2026-04-17
Fix cluster settings

ryanjjung:
This PR resolves an issue where the `REDIS_USE_CLUSTER` environment variable was not being interpreted correctly because it was being treated as a boolean value when all environment variable values are actually strings. This caused various conditions to always evaluate to `True` even when we were trying to get it to evaluate to `False`.

I solved this by adding a utility function, `get_envvar_as_bool`, which interprets a string as a boolean, then wrapping our retrievals of this variable with it.

I also fixed an issue where the backend container did not have a security group allowing it into the new Redis replicaset.

Additionally, I've swapped dev and stage over to using this. I will prepare prod and coordinate that deployment with @davinotdavid and the rest of the team in a separate PR.


davinotdavid (approved):

backend/src/appointment/dependencies/database.py
```diff
@@ -42,7 +42,8 @@ def get_db():
 def boot_redis_cluster():
     """Open a connection to a redis cluster"""
     global _redis_instance
-    if not os.getenv('REDIS_URL') or not os.getenv('REDIS_USE_CLUSTER'):
+
+    if not os.getenv('REDIS_URL') or not get_envvar_as_bool('REDIS_USE_CLUSTER'):
```
davinotdavid: Instead of having `get_envvar_as_bool` at all, wouldn't it suffice to mark it as `os.getenv('REDIS_USE_CLUSTER', False) == 'True'` instead?

backend/src/appointment/utils.py
```diff
@@ -19,6 +19,12 @@
 log = logging.getLogger(__name__)
 
 
+def get_envvar_as_bool(envvar):
+    """Returns False unless the environment variable is present and indicates truth."""
+    value = os.getenv(envvar, 'false')
+    return value in ['true', 'yes']
```
davinotdavid: I don't think we need this although I have been bitten before with these truth-y / false-y checks.
Some examples of how we handle this in other places:


https://github.com/thunderbird/thunderbird-accounts/blob/main/src/thunderbird_accounts/settings.py#L473

https://github.com/thunderbird/thunderbird-accounts/blob/main/src/thunderbird_accounts/settings.py#L547

davinotdavid: Also, in case this is `True`, I think this would fail as:

'True' in ['true', 'yes'] == False

backend/src/appointment/dependencies/database.py
ryanjjung: Yes, this is more efficient/Pythonic, I'll update.

backend/src/appointment/utils.py
ryanjjung: Ugh, this is missing the call to `.lower()`, will fix per the other comment, though.

backend/src/appointment/dependencies/database.py
ryanjjung: The downside here is this doesn't handle casing or alternative strings, so we'll just have to be careful in our configs.

davinotdavid (approved):
lgtm! thanks!

PR #1638, opened 2026-04-20
Destroy old Redis clusters

ryanjjung:
This wraps up the Redis migration for Appointment. Currently, the new Redis caches are online and taking all of the traffic in our environments. The old clusters' metrics have plummeted to zeroes. It's time to tear down the old, now-unused clusters. This PR removes the code that builds them and the YAML that configures them.


davinotdavid (approved):
Thank you so much for all the journey here! 🚀

PR #1642, opened 2026-04-21
FTUE remove CreateYourProfile (first) step

davinotdavid:
## What changed?

- Removed the CreateYourProfile step in FTUE (first step) as we can derive all the information from the OIDC
- In case there's no name set up (it is optional in Keycloak), we use the username but capitalized for the `booking page name`. For example, `admin@example.org` would be `Admin's Availability`.
- For the timezone, we are also prioritizing what is being set in Keycloak, otherwise we use DayJS's guess()

## Why?

- Subscribers won't have to input the same information again and there's a better cohesion between User records between Appointment <-> Accounts.
- Reduced FTUE steps might improve conversion!

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/1620


devmount (approved):
Looks great to me! I couldn't find any left-overs after the removal 🙌🏻

rwood-moz (approved):
Worked nicely for me locally.

PR #1649, opened 2026-04-23
Update OIDC external connection with preferred_username instead of email

### LINKED ISSUE #1647: Update the oidc token during login
We're remapping the token's email field, which accounts only uses for the name of the oidc external connection.

I think if an external connection is found we should update the name during login. (Using the preferred_username field.) Additionally we should change that create statement to also use the preferred_username field. 

https://github.com/thunderbird/appointment/blob/4acb53f620e1f9b5cbd2cdd10554719fa5984810/backend/src/appointment/routes/auth.py#L614-L632

davinotdavid:
## What changed?

Upon login, we are updating the OIDC external connection's name to be the `preferred_username` or `username` fields from the token.

## Why?

We can't use `email` anymore since this is being remapped (used to be the recovery email field during signup).

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/1647

### DIFF
--- a/backend/src/appointment/routes/auth.py
+++ b/backend/src/appointment/routes/auth.py
@@ -624,9 +624,7 @@ def oidc_token(
     if oidc_connection:
         ec = oidc_connection[0]
         if ec.name != email:
-            ec.name = email
-            db.commit()
-            db.refresh(ec)
+            repo.external_connection.update_name(db, ec, email)
     else:
         external_connection_schema = schemas.ExternalConnection(
             name=email,
--- a/backend/src/appointment/routes/auth.py
+++ b/backend/src/appointment/routes/auth.py
@@ -621,9 +621,15 @@ def oidc_token(
 
         raise HTTPException(403, l10n('invalid-credentials'))
 
-    if not oidc_connection:
+    if oidc_connection:
+        ec = oidc_connection[0]
+        if ec.name != email:
+            ec.name = email
+            db.commit()
+            db.refresh(ec)
+    else:
         external_connection_schema = schemas.ExternalConnection(
-            name=token_data.get('email'),
+            name=email,
             type=ExternalConnectionType.oidc,
             type_id=oidc_id,
             owner_id=subscriber.id,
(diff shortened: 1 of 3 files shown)


backend/src/appointment/routes/auth.py
```diff
@@ -621,9 +621,15 @@ def oidc_token(
 
         raise HTTPException(403, l10n('invalid-credentials'))
 
-    if not oidc_connection:
+    if oidc_connection:
+        ec = oidc_connection[0]
+        if ec.name != email:
+            ec.name = email
+            db.commit()
+            db.refresh(ec)
```
MelissaAutumn: Wow no repo function? Gasp!

backend/test/integration/test_auth.py
```diff
@@ -1005,14 +1005,15 @@ def test_oidc_token_existing_subscriber(self, with_client, make_pro_subscriber,
...
+            subscriber_id=subscriber.id, type=models.ExternalConnectionType.oidc, type_id=oidc_id, name='stale@old.com'
         )
 
         # Mock OIDCClient to return valid token data
         with patch('appointment.controller.apis.oidc_client.OIDCClient.introspect_token') as mock_introspect:
             mock_introspect.return_value = {
                 'sub': oidc_id,
-                'email': subscriber.email,
+                'email': 'different@email.com',
+                'preferred_username': subscriber.email,
```
MelissaAutumn: These values will actually be the same soon. We're mapping username over both preferred_username and email.

MelissaAutumn (approved):
Looks good, I'd take a look at the comments and fix the test though.

backend/src/appointment/routes/auth.py
davinotdavid: Architecture? What's that? 😅 

Moved to a repo function! Good catch!

backend/test/integration/test_auth.py
davinotdavid: Hmmm good to know, updated the test. Thanks!

PR #1652, opened 2026-04-24
Auto-refresh zoom tokens via celery task + implement task locks

### LINKED ISSUE #1580: Auto-refresh Zoom token access
Even though we are running health checks for Zoom and other external connections, it still requires manual user intervention to re-auth. This causes pain because if the user decides not to do so, accepted bookings will not generate zoom links and then they will need to manually go to the invite in their calendars, manually generate a link and add there.

It would be great if the Zoom token access auto refreshed in a cron job of sorts!

devmount:
## What changed?

This change implements a celery task that triggers a command to check all Zoom token expiry dates. If a token is expired or under a given time threshold, a token refresh is triggered by initializing the OAuth2 Session and making a lightweight `get_me()` call.

A new env var `ZOOM_TOKEN_TTL_IN_SECONDS` got introduced to specify, how long we assume an auth token to be valid before checking and renewing it.

If there is an error during the process, the external connection is marked with `status = error` so that at least it is shown up in the UI. Otherwise, it is marked as `status = ok`.

Also, we've added some repository functions for external connection token and status handling.

## Why?

This improvement should help users reducing manual effort to keep a Zoom connection active.

## Limitations and Notes

Added a comment with a TODO that we should proactively send an email to the Subscriber in case there's a problem with renewing the token: https://github.com/thunderbird/appointment/issues/1662

## How to test

Run `docker compose exec backend run-command main refresh-zoom-tokens` and then `docker compose logs celery-worker --tail 30` to check the celery worker logs.

## Applicable Issues

Closes #1580

### DIFF
--- a/backend/.env.test
+++ b/backend/.env.test
@@ -63,6 +63,7 @@ ZOOM_API_ENABLED=False
 ZOOM_AUTH_CLIENT_ID=
 ZOOM_AUTH_SECRET=
 ZOOM_AUTH_CALLBACK=http://localhost:8090/zoom/callback
+ZOOM_TOKEN_TTL_IN_SECONDS=604800 # 7 days
 
 # -- SIGNED URL SECRET --
 # Shared secret for url signing (e.g. create it by running `openssl rand -hex 32`)
--- a/backend/src/appointment/commands/refresh_zoom_tokens.py
+++ b/backend/src/appointment/commands/refresh_zoom_tokens.py
@@ -21,6 +21,7 @@ def run():
     zoom_connections = []
     refreshed = 0
     failed = 0
+    cleaned = 0
 
     try:
         zoom_connections = repo.external_connection.get_zoom(db)
@@ -33,7 +34,10 @@ def run():
             try:
                 subscriber = repo.subscriber.get(db, owner_id)
                 if not subscriber:
-                    raise ValueError('No subscriber found for external connection')
+                    # Orphaned external connection so let's just remove it.
+                    repo.external_connection.delete(db, connection)
+                    cleaned += 1
+                    continue
 
                 zoom_client = get_zoom_client(subscriber)
                 token = json.loads(connection.token)
@@ -73,5 +77,5 @@ def run():
 
     logging.info(
         f'[refresh_zoom_tokens] Zoom token refresh complete: '
-        f'{refreshed} refreshed, {failed} failed, {len(zoom_connections)} total processed'
+        f'{refreshed} refreshed, {failed} failed, {cleaned} cleaned, {len(zoom_connections)} total processed'
     )
--- a/backend/src/appointment/commands/refresh_zoom_tokens.py
+++ b/backend/src/appointment/commands/refresh_zoom_tokens.py
@@ -4,10 +4,10 @@ Run periodically (e.g., weekly) to ensure tokens don't expire.
 Zoom tokens typically last ~90 days, so weekly renewal keeps a buffer.
 """
 
+import json
 import logging
-from datetime import datetime, timedelta, timezone
 
-from ..database import repo
+from ..database import repo, models
 from ..dependencies.database import get_engine_and_session
 from ..dependencies.zoom import get_zoom_client
 from ..main import _common_setup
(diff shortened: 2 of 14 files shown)


backend/src/appointment/commands/refresh_zoom_tokens.py
```diff
@@ -0,0 +1,77 @@
...
+
+        for connection in zoom_connections:
+            if not connection or not connection.token:
+                continue
+
+            owner_id = connection.owner_id
+            try:
+                subscriber = repo.subscriber.get(db, owner_id)
+                if not subscriber:
+                    raise ValueError('No subscriber found for external connection')
```
MelissaAutumn: Shouldn't we mark these for clean up? 🤔 (I guess this can be punted to another ticket)

backend/src/appointment/routes/commands.py
```diff
@@ -69,3 +70,11 @@ def backfill_channels():
             backfill_google_channels.run()
     except FileExistsError:
         print('backfill-google-channels is already running, skipping.')
+
+@router.command('refresh-zoom-tokens')
+def refresh_tokens():
+    try:
+        with cron_lock('refresh_zoom_tokens'):
+            refresh_zoom_tokens.run()
```
MelissaAutumn: .run() for a beat scheduler will essentially hold up the beat scheduler until this task is finished. 

I think you'll want to move this to a .delay(). We'll want to move the lock into the task itself to avoid double firing. (https://docs.celeryq.dev/en/main/tutorials/task-cookbook.html) (we don't use memcache so just redis 🤔)

MelissaAutumn (changes requested):
Looks good, I have a few suggestions though. I can't test this locally right now, but I'm sure someone else can 😄

backend/src/appointment/commands/refresh_zoom_tokens.py
davinotdavid: Ah this should be straightforward enough so added the cleanup here (just deleting the orphan external connection). Good point, thanks!

backend/src/appointment/routes/commands.py
davinotdavid: Oh TIL, I was using the `cron_lock` in other tasks so I will cut tickets to update those too.

Is this a good pattern?

MelissaAutumn: cron_lock is good in systems that don't have a way to off-load work to a worker/queue system or for crons that run very frequently (every minute or so.) It ensures that no other cron will be able to run while the tmp file is created. 

Since we have celery now we can move these locks into out jobs using redis to store the lock (so other celery workers can see the lock.)

backend/src/appointment/tasks/zoom.py
```diff
@@ -0,0 +1,49 @@
...
+    task_name = refresh_zoom_tokens.__name__
+    redis_instance = get_redis(os.getenv('REDIS_CELERY_DB'))
+    lock_token = None
+
+    if redis_instance is not None:
+        lock_token = acquire_task_lock(redis_instance, task_name, DEFAULT_LOCK_TTL_SECONDS)
+        if lock_token is None:
+            log.info('Zoom token check is already running, skipping.')
+            return
+    else:
```
MelissaAutumn: This logic should probably be wrapped in a context manager so we can just use

```python
try:
    with task_lock('refresh_zoom_tokens'):
        run()
except TaskLockFailed:
    print('ah shoot it failed to acquire the lock')

... etc ...
```

With task_lock pulling redis instance, locking the task, and has a finally that unlocks the task. Default values can include the default lock ttl seconds.

MelissaAutumn (changes requested):
Thanks for the change, just wanted to suggest a better flow for it though.

backend/src/appointment/tasks/locks.py
```diff
@@ -0,0 +1,31 @@
...
+
+    # Only delete lock if still owned by this task instance.
+    release_script = """
+if redis.call("get", KEYS[1]) == ARGV[1] then
+    return redis.call("del", KEYS[1])
+else
+    return 0
+end
+"""
+    redis_instance.eval(release_script, 1, lock_key, lock_token)
```
MelissaAutumn: I hope this can be done without an eval 🤔

davinotdavid: Hmm I've replaced this with [Redis' Transactions](https://redis.io/docs/latest/develop/using-commands/transactions/#optimistic-locking-using-check-and-set), what do you think?

backend/src/appointment/tasks/zoom.py
davinotdavid: Ah nice, indeed much better to do that! Changed, thanks!

backend/src/appointment/tasks/locks.py
```diff
@@ -0,0 +1,68 @@
...
+    lock_key = _task_lock_key(task_name)
+
+    with redis_instance.pipeline() as pipe:
+        pipe.watch(lock_key)
+        if pipe.get(lock_key) == lock_token:
+            pipe.multi()
+            pipe.delete(lock_key)
+            pipe.execute()
+        else:
+            pipe.unwatch()
```
MelissaAutumn, line 38: I haven't tested this but looking at the docs this seems like it should work. I'm not sure why we're using pipeline here since we're essentially just deleting one key, but I guess this avoids us from having to scan for it?

backend/src/appointment/tasks/locks.py
```diff
@@ -0,0 +1,68 @@
...
+    if redis_instance is not None:
+        lock_token = acquire_task_lock(redis_instance, task_name, ttl_seconds)
+        if lock_token is None:
+            raise TaskLockFailed(f'Failed to acquire lock for {task_name}')
+    else:
+        logging.warning(f'Redis unavailable; running {task_name} without distributed lock.')
+
+    try:
+        yield
+    finally:
```
MelissaAutumn, line 66: We might want to log all uncaught errors as well.

backend/src/appointment/tasks/locks.py
```diff
@@ -0,0 +1,68 @@
...
+        if lock_token is not None:
+            try:
+                release_task_lock(redis_instance, task_name, lock_token)
+            except Exception as e:
+                logging.error(f'Failed to release {task_name} lock: {e}')
+                sentry_sdk.capture_exception(e)
```
MelissaAutumn, line 72: I noticed sentry creates "issues" for logging.error as well as capture_exception. I'm not sure what would be the best solution here, maybe drop the log down to warning?

MelissaAutumn (commented):
Just have some questions, no real changes. Looks good! Once they're answered I can approve.

backend/src/appointment/tasks/locks.py
MelissaAutumn: Ohhh I should have seen this before commenting. That works for me!

backend/src/appointment/tasks/locks.py
davinotdavid: Ah yes, added an `except` there, thanks!

backend/src/appointment/tasks/locks.py
davinotdavid: Hmmm I dropped the logging down to warning in this case and in the uncaught errors from the comment above too! Thanks!

MelissaAutumn:
I'm out today due to cat stuff, but will test tonight or tmw morning. Thanks!

devmount:
@MelissaAutumn / @davinotdavid Is there any update on this? The current state seems to be good to go, if I didn't miss anything 😇

backend/.env.test
```diff
@@ -63,6 +63,7 @@ ZOOM_API_ENABLED=False
 ZOOM_AUTH_CLIENT_ID=
 ZOOM_AUTH_SECRET=
 ZOOM_AUTH_CALLBACK=http://localhost:8090/zoom/callback
+ZOOM_TOKEN_TTL_IN_SECONDS=604800 # 7 days
```
MelissaAutumn, line 66: Do we want this in the pulumi configs?

devmount:
@MelissaAutumn @davinotdavid Just a little ping on this thread: Should we take a look at the changes once more? Should I test again? Let me know if there's any status update on this. I can happily review again if needed.

davinotdavid:
@devmount Whoops, misclicked here lol! Just waiting on @MelissaAutumn 's review as it is blocked by a requested changes that happened before.

MelissaAutumn (approved):
Looks good! I realize I have a pending comment from 200 years ago but should be good tbh.

backend/.env.test
davinotdavid: Ah yes, thanks! Also realized that I forgot to add the `GOOGLE_CHANNEL_TTL_IN_SECONDS` there so I've added that too heh

PR #1669, opened 2026-05-07
Fix LegalView content update toggle + CSS fixes

### LINKED ISSUE #1668: Privacy Policy / Terms page missing CSS styles
We've forgot a spot in the removal of TailwindCSS so now the terms page look less than ideal 😅

davinotdavid:
## What changed?

- Updated `LegalView` component to use vanilla CSS instead of TailwindCSS
- Fixed content update toggle (moving directly from Legal to Privacy was not working)

## Why?

- We recently removed TailwindCSS from Appointment but we missed a spot here
- There were two bugs in the `LegalView` component:
  - When navigating from Legal to Privacy Policy and vice-versa, it wasn't updating the content since we are using the same component for both views but were only updating the content `onMounted`
  - Also, the scroll position stayed stuck at the bottom so now scrolling back to top

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/1668

## Screenshots

Before (stage / prod)

image

--

After (this PR)

image

### DIFF
--- a/frontend/src/assets/legal/en/privacy.html
+++ b/frontend/src/assets/legal/en/privacy.html
@@ -1,13 +1,13 @@
-<h1 id="thunderbird-pro-services-privacy-notice"><strong>Thunderbird Pro Services Privacy Notice</strong></h1>
+<h1 id="thunderbird-pro-services-privacy-notice">Thunderbird Pro Services Privacy Notice</h1>
 <p>Last updated December 3, 2024</p>
 <p>Thunderbird Services (the “Services”) are a collection of online services that integrate with Thunderbird and other applications to enable users to manage their communications, appointments, and files.</p>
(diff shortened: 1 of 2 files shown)


rwood-moz (approved):
Much nicer! 🥇

kewisch:
fwiw if we have an appointment specific set of terms, we can simply replace this with a link to tb.pro/terms . Those terms are intended to cover all products.

davinotdavid:
@kewisch that's a good point, I thought about removing these pages altogether from Appointment but iirc there's some Zoom specific guidelines for the Zoom App approval (or was it Google?) that required us to have specific verbiage around that.

Maybe we need to update the broader terms / policy to include those before we axe this

devmount (approved):
Ohhhhh, sorry about that 😂 Totally missed that page 😇 Thanks for updating it Davi!

PR #1677, opened 2026-05-19
Initialize custom availabilities when first activating them

### LINKED ISSUE #1657: Schedules with use_custom_availabilities as true aren't bookable
## Describe the bug

When a user has their Availability / Schedule with `use_custom_availabilities` as `true` but no changes have been made to the defaults, it causes the Booking Page to show slots correctly but errors out when attempting to book any slot.

## To Reproduce

1. As a brand new user, go through FTUE with the all default settings (any calendar is fine)
2. Go to the Availability page, check the `Set custom times for each day` but don't change any of the pre-populated values
3. Save the Availability
4. Check in the DB that the `availabilities` table is empty
5. Copy the booking link URL, open an incognito tab and attempt to book any slot
6. See the error
7. Go back to the Availability page
8. Update one of the custom availabilities to any valid value
9. Save the Availability (at this point the booking page works as expected)
10. Move it back to the previous custom availability
11. Save the Availability (check that the booking page works as expected)

## Expected behavior

Turning on custom availabilities without changing any of the pre-populated values and saving the schedule shouldn't cause the booking page to not be bookable.

## Actual behavior

Booking a time under the conditions above errors out with the message: "The time slot you have selected could not be found. Please try again."

## Screenshots

Image

## Additional context

https://thunderbird.sentry.io/issues/7230911272/?environment=prod&project=4505428124827648&query=is%3Aunresolved&referrer=issue-stream

devmount:
## What changed?

This change emits an update when activating custom availability for the first time. Also, `onMounted` and `watch` were moved to the bottom of the script section, to keep logical definition-call order of functions intact.

## Why?

This change emits the default availabilities to the store. Otherwise, the store would just keep being empty if no change was made and the time slot would not be available for bookees.

## Limitations and Notes

None.

## Applicable Issues

Closes #1657 

## Screenshots

None.

### DIFF
--- a/frontend/src/views/AvailabilityView/components/AvailabilitySettings/components/AvailabilitySelect.vue
+++ b/frontend/src/views/AvailabilityView/components/AvailabilitySettings/components/AvailabilitySelect.vue
@@ -64,20 +64,6 @@ const defaultAvailabilitySet = () => Object.fromEntries(isoWeekdays.map((d) => {
   return [d.iso, existing.length ? existing.sort(compareAvailabilityStart) : [defaultAvailability(d.iso)]];
 })) as AvailabilitySet;
 
-// Prefill existing availabilities from schedule
-onMounted(() => {
-  availabilitySet.value = defaultAvailabilitySet();
-});
-
-// Track schedule reset
-watch(
-  () => props.availabilities,
-  () => {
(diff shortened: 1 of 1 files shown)


davinotdavid (approved):
Oh nice! LGTM ty 🎉

PR #1684, opened 2026-05-25
Pin package versions + npm ci in GH workflows

### LINKED ISSUE #1681: Pin package versions and use 'npm ci' instead
We should explicitly pin our packages and use npm ci during our project deployments

davinotdavid:
## What changed?

- Pinned NPM packages in both /frontend & /test/e2e
- Updated GitHub Workflows to run `npm ci` instead of `npm install`
- As part of the pinning, there was a need to update some of packages for peer dependency conflicts and/or security issues during `npm audit`:
  - `pinia` from `3.0.1` to `3.0.4`
  - `vue-i18n` from `11.1.2` to `11.1.10`
  - `eslint` from `9.23.0` to `9.32.0`
  - `eslint-plugin-import` from `2.25.2` to `2.32.0`
  - `msw` from `2.0.11` to `2.2.14`
  - `postcss` from `8.4.45` to `8.5.10`
- Updated the `post-install` npm script to remove the manual update to `browserstack-node-sdk` in test/e2e
 

## Why?

- Not having pinned versions for packages increase our changes of running through unintentional problems such as security vulnerabilities and different results in different environments

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/1681


test/e2e/package.json
```diff
@@ -18,18 +18,20 @@
...
+  "overrides": {
+    "uuid": "11.1.1"
+  },
   "devDependencies": {
     "@playwright/test": "1.59.1",
-    "@types/node": "^24.1.0",
-    "browserstack-node-sdk": "1.52.3",
-    "dotenv": "^17.2.1",
+    "@types/node": "24.1.0",
+    "browserstack-node-sdk": "1.55.1",
```
davinotdavid: @rwood-moz highlighting this and the removal of the `post-install` script that manually updated the browserstack-node-sdk package!

rwood-moz: Hey, thanks for removing the post-install step (and the npm ci) that's great! However please don't change the `browserstack-node-sdk` version - we need to keep it pinned to `1.52.3` (and the playwright/test version pinned also). I spent a bunch of time last week trying out the newest sdk version (and upgrading playwright/test) and the only one that works on BrowserStack for mobile and desktop browsers is `1.52.3`. Unfortunately BrowserStack and Playwright don't always work well together and upgrading one or the other can often cause BrowserStack issues. Thanks :)

rwood-moz (changes requested):

test/e2e/package.json
rwood-moz: It's strange though when the post-install step would update the node-sdk version; but when I tested this out last week from my local machine with the latest BrowserStack it wouldn't work on some platforms (in both appointment and other services) so not sure how it is currently working in CI, maybe the post-install isn't running there or something.

davinotdavid:
@rwood-moz ooh I see! I vaguely remember that there was a manual step to pin versions for browserstack so that they all play together but good to know that they won't work if we upgrade the package. Reverted, thanks!

test/e2e/package.json
davinotdavid: Hmmm I wonder if we should pin a different version then and update altogether? Seems like `npm update` would update minor and patch versions for caret packages so perhaps it was secretly updating it to something else 🤔

rwood-moz: Oh that's right npm update just does minor not major. Let me try running your branch with the tests in CI but I really prefer not to change the pinned versions because I spent hours last week going through the same thing. With some versions they work fine on desktop but then on mobile they won't even start the browser; some work on desktop and one mobile but not the other, etc. it's not fun haha

davinotdavid: Oh definitely! We can take a look at that later since it is working as expected now. I reverted the version to what we had before

rwood-moz: Cool thanks, I'm running your PR in CI now on BrowserStack desktop and mobile

rwood-moz (approved):
> @rwood-moz ooh I see! I vaguely remember that there was a manual step to pin versions for browserstack so that they all play together but good to know that they won't work if we upgrade the package. Reverted, thanks!

Thanks @davinotdavid. I ran this branch in BrowserStack ([desktop](https://automate.browserstack.com/projects/Thunderbird+Appointment/builds/Appointment+Nightly+Tests+Chromium+Desktop/214?testListView=spec), [android](https://automate.browserstack.com/projects/Thunderbird+Appointment/builds/Appointment+Nightly+Tests+Android+Chrome/227?tab=sessions&testListView=spec), and [iOS](https://automate.browserstack.com/projects/Thunderbird+Appointment/builds/Appointment+Nightly+Tests+iOS+Safari/173?testListView=spec)) and it looks good.

Issue #1679, opened 2026-05-19
Shut down appointment.day infrastructure


kewisch:
We're ready to shut down our old infrastructure for appointment.day. We should redirect both appointment.day and apmt.day to https://tb.pro/appointment/ so users will know what the new thing is and can reach out to their link owners if necessary.

I'm not sure if there is any data we need to retain?

If there is any clean up on the AWS account necessary let's also get that done.

ryanjjung:
# Obituary/Autopsy for `appointment.day`

I tore down all the Terraform-managed resources associated with the "stage" and "prod" installations of Appointment to the `appointment.day` domain, working in reverse order from the build/apply instructions that used to be in the `tofu` directory of this repo. There will be a PR coming very shortly to delete the relevant code.

## Roadblocks, Issues, Manual Things, General Notes

- This project required setting several environment variables. Some mattered (environment name), most didn't (because I was destroying resources, I didn't have to match up correct values). The Tofu Readme had several listings I was able to compile a full list from. Without these, terragrunt would fail at the `init` step and not let me proceed to a `destroy`.
- We were using an RDS Terraform module provided by AWS (above and beyond the provider, this is a larger pattern for building databases with Terraform), and this was not version-locked. There was an incompatibility with the latest version of this module and our declared version of Terraform. I was able to find release notes about the version that bumped the minimum required version for the module and pin that module to that version to get through the destruction.
- I took final backup snapshots of the databases, but these snapshots were still associated with the option groups for the databases that we needed them for. The Terraform code manages those option groups, so after creating the snapshots and creating a new manual dependency, Terraform wasn't able to delete them. To resolve this, I manually created some alternate option groups to survive with these manual snapshots, made copies of the snapshots using the new options groups, and deleted the old snapshots. This removed the association between the snaps and the options.
- There were also some old manual snapshots of these databases (one as old as 2024) associated with those option groups. I just deleted those old snapshots, opting to keep only the most recent version of the data. Along with the above fix, this allowed Terraform to delete the option groups it was managing.
- I found an old jumphost instance blocking VPC subnet deletion. Deleted that by hand to unblock Terraform.
- I found three manually created subnets blocking VPC deletion. These had rules referring to each other, so I had to delete those rules first and then the subnets.
[...]

PR #1686, opened 2026-05-26
Delete old Tofu/Terraform infra code

ryanjjung:
I destroyed all of the resources this code refers to, and that was the last thing we needed this code for. You can find the [full autopsy over here](https://github.com/thunderbird/appointment/issues/1679#issuecomment-4548724159).

We have renewed the ownership of `appointment.day` (and maybe `apmt.day`?) for the time being (to avoid them being used for phishing attempts, etc.), but we intend on dropping those when they run out again. In the meantime, those domains all point to https://tb.pro/appointment.


MelissaAutumn (approved):
This looks good. Do we have any lingering github actions for terraform/tofu?

PR #1688, opened 2026-05-29
Standardize code formatter: Prettier

### LINKED ISSUE #1430: Define & Standardize code formatter settings for better DX
We currently don't have a standardized code formatter which causes unnecessary bloat in code changes per PR just for different preferences on how the auto formatting is set in each dev's environment.

Regardless of the IDE / code editor used, this would be super helpful to define so all devs can set a "format on save" option of sorts.

This can be quite opinionated but I believe once we settle on a good standard, it will greatly reduce the cognitive overload on PR reviews!

devmount:
## What changed?

This PR adds the `prettier` and `prettier-plugin-sort-json` packages. A command to apply Prettier was added to the scripts section:

```bash
npm run format
```

The following configuration was applied to all frontend code:

```json
{
  "overrides": [
    {
      "files": ["./src/locales/**"],
      "options": {
        "plugins": ["prettier-plugin-sort-json"],
        "jsonRecursiveSort": true
      }
    }
  ],
  "plugins": ["prettier-plugin-sort-json"],
  "printWidth": 120,
  "semi": true,
  "singleQuote": true,
  "tabWidth": 2,
  "trailingComma": "es5"
}
```

In short: We indent using 2 spaces, quote with single quotes, break code at 120 chars and sort JSON lang keys in all JSON files under `src/locales/`.

## Why?

We want to prevent unnecessary bloat in code changes by different devs.

## Limitations and Notes

- This is of course opinionated. Let's discuss those settings and tweak them until it fits our teams needs.
- We need to decide, *when* to run `npm run format` exactly, e.g. on save or in a pre-commit hook.
- Heads-up: We had node version mismatches. Some of our workflow files were still using Node 20, whereas some were already using 24. I unified all versions to 24 and fixed the lockfile accordingly.
- The timezone test failed. I couldn't find the exact reason why, but I could replicate the issue locally. I fixed it by changing the expected value for now, but please take a look at that issue, maybe this was introduced with the node version change?

## Applicable Issues

Closes #1430 

## Screenshots

No UI changes.


frontend/package.json
```diff
@@ -53,6 +54,8 @@
     "postcss-custom-media": "12.0.0",
     "postcss-mixins": "12.1.2",
     "postcss-nesting": "14.0.0",
+    "prettier": "^3.8.3",
+    "prettier-plugin-sort-json": "^4.2.0",
```
davinotdavid: We should pin the version / remove the `^` for the packages here

frontend/.prettierignore
```diff
@@ -0,0 +1,27 @@
...
+# Logs
+logs
+*.log
+
+# Environment files
+.env
+.env.*
+
+/.pnpm-store/
+.pnpm-lock.yaml
```
davinotdavid: I think we can remove this as we don't use pnpm currently

frontend/test/stores/appointment-store.test.js
```diff
@@ -120,7 +114,7 @@ describe('Appointment Store', () => {
   test('timezone', async () => {
     const apmt = createAppointmentStore(createFetch({ baseUrl: API_URL }));
     await apmt.fetch();
-    expect(apmt.appointments[0].slots[0].start.toISOString()).toBe('3000-01-01T01:00:00.000Z');
+    expect(apmt.appointments[0].slots[0].start.toISOString()).toBe('3000-01-01T02:00:00.000Z');
   });
```
davinotdavid, line 121: Ah the timezone is currently guessed here so this change actually makes the test fail for me when running locally. Nothing related with the Node changes! I believe we can remove the flaky-ness / locale dependent by importing `useUserStore` and doing:

```suggestion
test('timezone', async () => {
  const user = useUserStore();
  user.data.settings.timezone = 'America/Vancouver';

  const apmt = createAppointmentStore(createFetch({ baseUrl: API_URL }));
  await apmt.fetch();
  expect(apmt.appointments[0].slots[0].start.toISOString()).toBe('3000-01-01T01:00:00.000Z');
});
```

davinotdavid (changes requested):
Great work! I like the defaults you've set and the addition of the `prettier-plugin-sort-json`. Other than the comments below, there are a few extra things that I believe it would be worthwhile to add to this PR:

[.vscode specific stuff]
For those who don't have the Prettier extension installed, we could add a `.vscode/extensions.json`:

```json
{
  "recommendations": ["esbenp.prettier-vscode", "editorconfig.editorconfig"]
}
```

I believe we should add the `.vscode/settings.json` with some defaults like:

```json
{
  "editor.formatOnSave": true,
  "editor.defaultFormatter": "esbenp.prettier-vscode"
}
```

An example of this + `.gitignore` changes can be found in the `thunderbird-accounts` PR https://github.com/thunderbird/thunderbird-accounts/pull/852/changes

Lastly, since we also currently have an [.editorconfig](https://github.com/thunderbird/appointment/blob/main/.editorconfig) that may not be 1:1 with the prettier configs (and keeping them both in sync is not ideal), I would lean towards having a pre-commit hooks that run the formatter but that can be punted into a different PR.

rwood-moz (approved):
This is awesome and I am glad that you also are using the same prettier formatting settings that are already [implemented in accounts](https://github.com/thunderbird/thunderbird-accounts/blob/main/.prettierrc), great to be consistent across the services.

frontend/test/stores/appointment-store.test.js
devmount: Ohh thank you for this suggestion! It's way better like this!

devmount: Huh, GitHub wouldn't let me apply this suggestion 🤷🏻 I'll add it manually then.

image

frontend/.prettierignore
devmount: Yes, agreed

frontend/package.json
devmount: Ahh running `npm add ...` defaults to `^`, great catch!

devmount:
> Great work! I like the defaults you've set and the addition of the `prettier-plugin-sort-json`. Other than the comments below, there are a few extra things that I believe it would be worthwhile to add to this PR:

@davinotdavid  Nice, I added those! Thanks so much, Davi! I kept the `.editorconfig` for now, let's improve that in a follow-up PR.

> This is awesome and I am glad that you also are using the same prettier formatting settings that are already [implemented in accounts](https://github.com/thunderbird/thunderbird-accounts/blob/main/.prettierrc), great to be consistent across the services.

@rwood-moz Oh hahaha I didn't know that 😂 Thanks Rob! I was mostly copying from the `services-ui` Prettier config! I'm glad all those repos are getting closer to consistency perfection 😇

davinotdavid:
@devmount thanks for the changes! that timezone test is still failing in CI (but passing locally for me), weird. I am also checking here what's up

devmount:
@davinotdavid I just saw that too, locally passing, CI failing again, that's weird indeed 🤔

davinotdavid:
@devmount hmmm AI says that we should have the timezone configuration in [vitest.config.ts](https://github.com/thunderbird/appointment/blob/main/frontend/vitest.config.ts#L15-L17) instead of `vitest.global-setup.ts` and also update the slot date in `restHandlers` to be a closer date like 2024 instead of year 3000 🤔

devmount:
@davinotdavid Ok I tried like 5 different approaches, the only thing working was indeed just using another testing date (I tried with 2024-02), wtf?? 😂

davinotdavid:
@devmount that's wild lol I honestly don't know why either but hey, since it is stable now, great! thanks for all the attempts!

davinotdavid (approved):
lgtm! thanks! 🚀 this will very much improve the diffs in the frontend 🎉

devmount:
Awesome, thanks guys for the quick reviews 🙇🏻 Merging in 🙏🏻

PR #1689, opened 2026-05-29
fix(auth): use standard OIDC email claim for subscriber email

kvdb:
## What changed?

The OIDC token endpoint (`backend/src/appointment/routes/auth.py`) now derives a
new subscriber's email from the standard OIDC `email` claim, falling back to
`preferred_username`/`username` only when it's absent.

Adds two integration tests in `backend/test/integration/test_auth.py` for both
paths (email claim present; falls back to `preferred_username`).

## Why?

The scheme derived the email from `preferred_username`, assuming
`preferred_username == email`. That only holds for Thunderbird/Firefox Accounts.
For other IdPs (Keycloak, Kanidm, ...) `preferred_username` is a display handle,
not an email, so subscribers got a bogus email, breaking host notification
emails. Per OIDC Core 1.0 §5.1 `preferred_username` "MUST NOT be relied upon" as
an identifier; the `email` claim (§5.4) is the authoritative source.

Ref: https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims

## Limitations and Notes

Trusts the introspected `email` claim without checking `email_verified` (matching
prior behavior). Verified locally: full backend suite passes (403 passed, 1
skipped), `ruff check` clean.

### DIFF
--- a/backend/src/appointment/routes/auth.py
+++ b/backend/src/appointment/routes/auth.py
@@ -597,8 +597,12 @@ def oidc_token(
     oidc_id = token_data.get('sub')
     name = token_data.get('name')
 
-    # preferred_username is the thundermail address (@thundermail.com)
-    email = token_data.get('preferred_username', token_data.get('username'))
+    # Per OIDC Core 1.0 §5.1, the standard `email` claim is the authoritative email
+    # address, while `preferred_username` is only a display/login handle (it MUST NOT
+    # be relied upon to hold an email). Prefer `email`, falling back to
+    # `preferred_username`/`username` for Thunderbird Accounts, where the username is
+    # the (@thundermail.com) email address.
+    email = token_data.get('email') or token_data.get('preferred_username') or token_data.get('username')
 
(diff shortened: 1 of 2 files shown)


davinotdavid (approved):
LGTM! Thank you so much @kvdb ! Really appreciate the contribution here 🙏 

In the recent-ish past, we were relying on the `email` claim to be where we stored the recovery email / not the @thundermail.com address. This was done so that we could reach out to users in case they had problems accessing their Thundermail address.

This caused confusion and since then we have moved to use a `recovery_email` field instead and moved to map the `email` to also be the thundermail address so we can now be compliant with the OIDC Core. Thanks again!

Also, if you are interested and not there yet, we'd love to have you in the Thunderbird Services Matrix channel where we actively chat about current / future developments:

https://matrix.to/#/#tb-services:mozilla.org

markstos:
@davinotdavid I've rebased and pushed. Is this ready to merge?

PR #1690, opened 2026-05-29
Fix short links targetting root to redirect to the root page instead of /user/

### LINKED ISSUE #638: consolidate apmt.day with appointment.day routes
a user notes that "appointment" being quite a long word to type, [they] will try to use apmt.day to go to appointment.day; but we currentl redirect `apmt.day` (with or without trailing slash) to `appointment.day/user/` which goes to a "Sorry, this page could not be found." instead of the splash page. also applies to the empty/root path, which results in an error page. If it's not too difficult it would be neat to redirect https://apmt.day/ to https://appointment.day/.
```

apmt.day should go to appointment.day. We can look at other routes that we may need to consolidate

### PR DESCRIPTION
**Pull request by davinotdavid**
## What changed?

Updated the `cloudfront-rewrite.js` handler function to check if the request URI is `/` or empty `''` and redirect to the root of the app instead of the `/user/` route.

## Why?

We can assume that a `/` or `''` uri means root. The "else" case is left unchanged (we assume that everything else is probably a user's short link). This might be worth taking a closer look at in case we add more redirects in the future but there's no plans for that in the foreseeable future AFAIK.

## Applicable Issues

Closes https://github.com/thunderbird/appointment/issues/638

### CODE REVIEW

File: pulumi/cloudfront-rewrite.js [hunk 1]
```diff
@@ -21,13 +21,19 @@
...
     }
 
-    // If we got a short URL, rewrite to the long domain with '/user' prepended
+    // Short URLs omit /user/; prepend it when redirecting booking paths to the long domain.
+    // Root requests (apt.mt/) should land on the main site home, not /user/.
     if (Object.keys(domainRewrites).includes(host)) {
+      const longHost = domainRewrites[host];
+      const location = request.uri === '/' || request.uri === ''
+        ? `https://${longHost}/`
+        : `https://${longHost}/user${request.uri}`;
```
aatchison: **Preserve query strings on the redirect, and harden the path against open-redirect drift.**

Two things bundled here (happy to split if you'd prefer):

1. **Query strings are dropped.** `location` is built from `request.uri` only, so `apt.mt/john?month=2026-06` loses `?month=2026-06`. In CloudFront Functions (`cloudfront-js-2.0`), `request.querystring` is an **object** (`name -> {value, multiValue?}`), not a string — so appending it directly would emit `?[object Object]`. It needs to be serialized (expanding `multiValue`, preserving raw already-encoded values).
2. **Open-redirect hardening (defense-in-depth).** Today the redirect can't leave the domain only *incidentally*, because `/user` is always prepended. Making it an explicit invariant — treat only a single-slash-rooted path as a booking path; send root/empty/`//host`/`/\host` to the home page — keeps a future edit from re-opening it.

```suggestion
    // Rebuild the request's query string into a redirectable string. CloudFront Functions
    // expose request.querystring as an object (not a string), so we serialize it ourselves
    // and preserve raw values (they arrive already URL-encoded). Returns '' when empty.
    const buildQueryString = (querystring) => {
      const params = [];
      Object.keys(querystring).forEach((key) => {
        const param = querystring[key];
        if (param.multiValue) {
          param.multiValue.forEach((entry) => params.push(`${key}=${entry.value}`));
        } else {
          params.push(`${key}=${param.value}`);
        }
      });
      return params.length ? `?${params.join('&')}` : '';
    };

    // Short URLs omit /user/; prepend it when redirecting booking paths to the long domain.
    // Root requests (apt.mt/) should land on the main site home, not /user/.
    if (Object.keys(domainRewrites).includes(host)) {
      const longHost = domainRewrites[host];
      const queryString = buildQueryString(request.querystring);
      // Only a single-slash-rooted path is a booking path. Root, empty, or anything that could
      // escape the domain ('//host', '/\host') falls back to the home page.
      const isBookingPath = request.uri.charAt(0) === '/'
        && request.uri.charAt(1) !== '/'
        && request.uri.charAt(1) !== '\\'
        && request.uri !== '/';
      const path = isBookingPath ? `/user${request.uri}` : '/';
      const location = `https://${longHost}${path}${queryString}`;
```

Notes from reviewing this against the AWS docs:
- **Don't add `encodeURIComponent`** to the values — CloudFront forwards querystring values "without changing them" (already percent-encoded), so re-encoding would double-encode (`%20` -> `%2520`). Worth confirming once via `aws cloudfront test-function` with e.g. `?q=a%20b%26c`.
- All constructs used are supported by `cloudfront-js-2.0` (no `for...of`, no `async` closures).
- Verified locally across root, multi-param, `NoValue=`, `multiValue`, pre-encoded values, and the `//host`/`/\host` guard cases.

aatchison: **How this was tested** (logic-level, not on live CloudFront):

I loaded the actual file's `handler` into Node and ran it against mock events shaped like the [documented CloudFront Functions event](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/functions-event-structure.html):

```js
const code = fs.readFileSync('pulumi/cloudfront-rewrite.js', 'utf8');
const handler = new Function(code + '\nreturn handler;')(); // runs the real file, not a copy
const ev = (host, uri, querystring = {}) => ({
  request: { method: 'GET', uri, querystring, headers: { host: { value: host } } },
});
```

11 cases asserting on `res.headers.location.value`, all PASS:

| Case | Input | Expected `location` |
|---|---|---|
| root `/` | `apt.mt`, `/` | `https://appointment.tb.pro/` |
| empty uri | `apt.mt`, `''` | `https://appointment.tb.pro/` |
| short link | `apt.mt`, `/john` | `…/user/john` |
| stage host | `stage.apt.mt`, `/jane` | `https://appointment-stage.tb.pro/user/jane` |
| single param | `/john` + `month=2026-06` | `…/user/john?month=2026-06` |
| multi params | `/john` + `a=1`,`b=2` | `…/user/john?a=1&b=2` |
| empty-value param | `/john` + `NoValue=''` | `…/user/john?NoValue=` |
| `multiValue` | `/john` + `tag=[x,y]` | `…/user/john?tag=x&tag=y` |
| pre-encoded value | `/john` + `q=a%20b` | `…/user/john?q=a%20b` (no double-encode) |
| root + param | `/` + `ref=newsletter` | `…/?ref=newsletter` |
| non-rewrite host | `appointment.tb.pro`, `/index.html` | passes through (no redirect) |

The `pre-encoded value` and `multiValue` cases are the ones that would catch a naive `'?' + request.querystring`.

**Caveat:** this verifies the logic against AWS's documented event contract; it does not run on the real `cloudfront-js-2.0` engine. I limited the code to constructs already used and working in this file (`Object.keys`, `forEach`, `push`, template literals, arrow functions). Full runtime confirmation would need AWS's `test-function` console.

davinotdavid: Hmmm while that's a good point, AFAIK there's no current use case at the moment for handling query params. Wouldn't hurt to do this now, so I will commit the suggestion.

aatchison:
### Suggested follow-up: a regression test for the redirect logic

Now that the query-string handling is in, it'd be good to lock the redirect behavior down with a test — there's currently no coverage for this function. The repo already runs **vitest** in `frontend/`, so it can host this with no new tooling.

The function is deployed by reading the file verbatim (`pulumi/cloudfront.py`), so it must stay export-free; the test loads the source as text and materializes `handler` via `new Function` rather than importing it.

I verified this **green against the committed code (`f7054e94`)** — 12 passing tests:

<details>
<summary><code>frontend/test/pulumi/cloudfront-rewrite.test.js</code></summary>

```js
import { describe, it, expect } from 'vitest';
import { readFileSync } from 'node:fs';
import { resolve } from 'node:path';

// The CloudFront Function is deployed by reading this file verbatim (see pulumi/cloudfront.py),
// so it must stay a bare `async function handler(event) {...}` with no import/export. We can't
// import it; instead we load the source and materialize `handler` via `new Function`.
//
// Caveat: this runs on Node's engine, NOT the cloudfront-js-2.0 runtime, so it verifies the
// redirect/serialization logic against AWS's documented event contract — not runtime parsing.
// Definitive runtime validation needs `aws cloudfront test-function`.
// vitest runs from frontend/ (its `test` script lives in frontend/package.json), so the
// CloudFront Function source sits one directory up under pulumi/.
const source = readFileSync(resolve(process.cwd(), '../pulumi/cloudfront-rewrite.js'), 'utf8');
// eslint-disable-next-line no-new-func
const handler = new Function(`${source}\nreturn handler;`)();

// Build a fresh event per call (the handler mutates request.uri in place on the passthrough path).
const makeEvent = (host, uri, querystring = {}) => ({
  request: {
    method: 'GET',
    uri,
    querystring,
    headers: { host: { value: host } },
  },
});

const locationOf = async (event) => {
  const res = await handler(event);
  return res?.headers?.location?.value ?? null;
};

describe('cloudfront-rewrite handler — short-domain redirects', () => {
  it('redirects the root path to the long-domain home, not /user/', async () => {
    const res = await handler(makeEvent('apt.mt', '/'));
    expect(res.statusCode).toBe(302);
    expect(res.statusDescription).toBe('Found');
    expect(res.headers.location.value).toBe('https://appointment.tb.pro/');
  });

  it('treats an empty uri as root', async () => {
    expect(await locationOf(makeEvent('apt.mt', ''))).toBe('https://appointment.tb.pro/');
  });

  it('prepends /user to a short booking link', async () => {
    expect(await locationOf(makeEvent('apt.mt', '/john'))).toBe('https://appointment.tb.pro/user/john');
  });

  it('maps the stage short domain to the stage long domain', async () => {
    expect(await locationOf(makeEvent('stage.apt.mt', '/jane'))).toBe('https://appointment-stage.tb.pro/user/jane');
  });

  describe('query string preservation', () => {
    it('keeps a single query param', async () => {
      expect(await locationOf(makeEvent('apt.mt', '/john', { month: { value: '2026-06' } })))
        .toBe('https://appointment.tb.pro/user/john?month=2026-06');
    });

    it('keeps multiple query params in order', async () => {
      expect(await locationOf(makeEvent('apt.mt', '/john', { a: { value: '1' }, b: { value: '2' } })))
        .toBe('https://appointment.tb.pro/user/john?a=1&b=2');
    });

    it('keeps a valueless param as key= (AWS NoValue shape)', async () => {
      expect(await locationOf(makeEvent('apt.mt', '/john', { NoValue: { value: '' } })))
        .toBe('https://appointment.tb.pro/user/john?NoValue=');
    });

    it('expands a multiValue param into repeated keys', async () => {
      expect(await locationOf(makeEvent('apt.mt', '/john', {
        tag: { value: 'x', multiValue: [{ value: 'x' }, { value: 'y' }] },
      }))).toBe('https://appointment.tb.pro/user/john?tag=x&tag=y');
    });

    it('preserves already-encoded values without double-encoding', async () => {
      expect(await locationOf(makeEvent('apt.mt', '/john', { q: { value: 'a%20b' } })))
        .toBe('https://appointment.tb.pro/user/john?q=a%20b');
    });

    it('carries the query string through on the root redirect', async () => {
      expect(await locationOf(makeEvent('apt.mt', '/', { ref: { value: 'newsletter' } })))
        .toBe('https://appointment.tb.pro/?ref=newsletter');
    });
  });
});

describe('cloudfront-rewrite handler — non-rewrite hosts pass through', () => {
  it('does not redirect a long-domain request and rewrites SPA paths to /index.html', async () => {
    const res = await handler(makeEvent('appointment.tb.pro', '/dashboard'));
    expect(res.statusCode).toBeUndefined();
    expect(res.uri).toBe('/index.html');
  });

  it('strips the /api/v1 prefix from API requests', async () => {
    const res = await handler(makeEvent('appointment.tb.pro', '/api/v1/me'));
    expect(res.uri).toBe('/me');
  });
});
```
</details>

Happy to push this to the branch if you'd like it in the PR.

> Side note: the open-redirect guard from the earlier suggestion (treating only single-slash-rooted paths as booking paths, so `//host` / `/\host` go to the home page) didn't make it into `f7054e94` — only the query-string part was committed. It's optional defense-in-depth; let me know if you'd like me to re-raise it.

aatchison (approved):
LGTM

PR #1700, opened 2026-06-05
Fix slug generation & improve error handling

### LINKED ISSUE #1696: Slugs are scope-generated on the Subscriber but indexed globally in DB
## Describe the bug

We are generating slugs for schedules / availabilities based on whether or not there's already a schedule with the generated slug [[ref](https://github.com/thunderbird/appointment/blob/main/backend/src/appointment/database/repo/schedule.py#L153)].

This would be fine, except that we are indexing in the DB globally so collisions of slugs would fail a PUT to schedule.

## To Reproduce

- Go through FTUE
- Navigate to /availability and copy the slug of the booking link
- Logout
- Go through FTUE again with another user
- Navigate to /availability and paste the slug of the booking link of the first user
- Attempt to save the schedule

## Expected behavior

Slugs shouldn't be indexed globally OR we should compare the generated slug with all the other slugs, independently of the user

## Actual behavior

Conflicting slugs block user availability to be saved

## Additional context

https://thunderbird.sentry.io/issues/7525937415/?project=4505428124827648&query=is%3Aunresolved&referrer=issue-stream

gz#6044

davinotdavid:
## What changed?

- Slugs are now generated considering global scope instead of subscriber's scope
- If the subscriber manually input a slug that conflicts during availability / schedule update, we are catching this error and we show a better error message (screenshot below) instead of a generic cryptic 500
- Added a slew of tests

## Why?

Slugs are unique globally but they were generated scoped on the requester's / owner's / subscriber's id which already caused conflicts / support tickets.

## Limitations and Notes

We might want to increase the number of characters of the slug in the near-ish future or consider a better strategy for collision.

## Applicable Issues

Closes https://github.com/thunderbird/appointment/issues/1696

## Screenshots

When a manually input slug collides with an existing slug:

image

### DIFF
--- a/backend/src/appointment/database/repo/schedule.py
+++ b/backend/src/appointment/database/repo/schedule.py
@@ -7,6 +7,7 @@ import uuid
 import zoneinfo
 
 from datetime import datetime, time, timezone
+from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import Session
 from .. import models, schemas, repo
 from ... import utils
@@ -152,22 +153,24 @@ def generate_slug(db: Session, schedule_id: int) -> str | None:
     if schedule.slug:
         return schedule.slug
 
-    # If slug isn't provided, give them the last 8 characters from a uuid4
-    # Try up-to-3 times to create a unique slug
+    # If slug isn't provided, give them the last 8 characters from a uuid4.
+    # Try up-to-3 times to create a unique slug.
     for _ in range(3):
         slug = uuid.uuid4().hex[-8:]
-        if not slug_exists(db, slug):
-            schedule.slug = slug
-            break
+        if slug_exists(db, slug):
+            continue
 
-    # Could not create slug due to randomness overlap
-    if schedule.slug is None:
-        return None
+        schedule.slug = slug
+        db.add(schedule)
 
-    db.add(schedule)
-    db.commit()
+        try:
+            db.commit()
+            return schedule.slug
+        except IntegrityError:
+            db.rollback()
+            schedule.slug = None
 
-    return schedule.slug
+    return None
 
 
 def hard_delete(db: Session, schedule_id: int):
--- a/backend/src/appointment/database/repo/schedule.py
+++ b/backend/src/appointment/database/repo/schedule.py
@@ -44,13 +44,20 @@ def get_by_slug(db: Session, slug: str, subscriber_id: int) -> models.Schedule |
     )
 
 
-def slug_exists(db: Session, slug: str) -> bool:
-    """Check if a slug is already in use (globally unique in the database)."""
+def slug_taken(db: Session, slug: str) -> bool:
+    """True if any schedule already uses this slug."""
+    return db.query(models.Schedule).filter(models.Schedule.slug == slug).first() is not None
+
+
+def slug_available_for_schedule(db: Session, slug: str, schedule_id: int) -> bool:
+    """True if this slug is unused or already belongs to the given schedule."""
     return (
         db.query(models.Schedule)
-        .filter(models.Schedule.slug == slug)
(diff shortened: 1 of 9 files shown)


backend/src/appointment/l10n/en/main.ftl
```diff
@@ -35,6 +35,7 @@ slot-not-auth = You are not authorized to view or modify this time slot.
 account-delete-fail = There was a problem deleting your data. This incident has been logged and your data will manually be removed.
 protected-route-fail = No valid authentication credentials provided.
 username-not-available = This username has already been taken.
+schedule-slug-taken = This link slug is already in use. Please choose a different one.
```
rwood-moz, line 38: Will this error message make enough sense to an average Appointment user (what if they don't know what a `slug` is)?

backend/src/appointment/database/repo/schedule.py
```diff
@@ -43,6 +44,21 @@ def get_by_slug(db: Session, slug: str, subscriber_id: int) -> models.Schedule |
...
+
+
+def slug_available_for_schedule(db: Session, slug: str, schedule_id: int) -> bool:
+    """True if this slug is unused or already belongs to the given schedule."""
+    return (
+        db.query(models.Schedule)
+        .filter(models.Schedule.slug == slug, models.Schedule.id != schedule_id)
+        .first()
+        is None
+    )
```
rwood-moz, line 59: If a Thundermail user has deleted their account, but previously had an Appointment schedule and a slug, would that still be found in this query (i.e. is there a separate table for deleted accounts or anything that needs to be checked)? I'm guessing not, but just a thought.

davinotdavid: The Thundermail DB is separated from the Appointment DB and the off boarding process is still under discussion iirc.

Ideally both DBs would be automatically cleaned and deleting a Subscriber in Appointment also deletes their Schedule and would free up the slug for usage!

backend/src/appointment/l10n/en/main.ftl
davinotdavid: Fair point! Although this would only happen now if the user manually inputs a slug that is the exact same of another users'. If they leave it for the auto generation, it shouldn't happen!

If they manually input a slug there in the link I'd assume they know what it is but we can update the copy later in case it's confusing

backend/src/appointment/database/repo/schedule.py
rwood-moz: Ok, makes sense, thank you

rwood-moz (changes requested):
Code LGTM, awesome tests!

- Verified that in your PR checks the integration tests all passed and the unit tests all passed too including your new addition
- Did some testing locally:
   - Went through the FTUE for a new user, booking link worked great, changed the booking link slug to a new unique on, it saved and worked great
   - Refreshed link with link refresh button, verified new booking link worked
   - Created a second Appointment user and attempted to change the booking link to the same slug as the first on, and the `This link slug is already in use` error appeared as expected
   - Then changed the slug and was able to save it and the new link worked fine

I did find one small issue:
   - I repeated the test - attempted to change booking link slug to one that already exists, got the slug in use error message as expected, but this time instead of manually changing the slug in the input field I clicked the `refresh` button to have it change, and then confirmed; the slug changed fine however the `This link slug is already in use` error message still stays at the top of the screen even though the slug was changed and saved

davinotdavid:
@rwood-moz thanks for the review! great point, added a call to `clearNotice` to remove the notices upon success!

rwood-moz (approved):
Thanks for the update, gave it a try locally and confirmed the `slug already in use` error does indeed disappear now if you change the slug via the link refresh button. Nice!

PR #1707, opened 2026-06-15
Upgrade vitest and tornado (for flower) and bump version to 1.4.7

davinotdavid:
- Bump version to `1.4.7`
- Upgrade vitest to `4.1.8`
- Tighten fastapi dependency to `0.136.*` for better version consistency between local / CI
- Fresh `package-lock.json` without esbuild

### DIFF
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "appointment"
-version = "1.4.6"
+version = "1.4.7"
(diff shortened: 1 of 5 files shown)


MelissaAutumn:
I don't believe we need esbuild anymore if we're using Vite 8. It should pull in rolldown which is meant to replace it.

MelissaAutumn (approved):
Container boots and frontend looks fine on my end!

PR #1709, opened 2026-06-15
Upgrade @babel/core and PyJWT to fix security vulnerabilities

davinotdavid:
Further package bumps :(


MelissaAutumn (approved):
Ok we're switching to c, but we're hand writing everything. No more CVE's besides the hidden ones we write ourselves along the way. 😊 (Important note: This is a joke.)

Thanks for the dep bump!

PR #1711, opened 2026-06-15
Make slug generation / validation be constrained on the owner_id

### LINKED ISSUE #1706: Make username be part of the slug to avoid global collisions of common slugs
We recently had support tickets about booking link slug generation collision during the auto-generation button flow / FTUE. This is a problem since it blocks any changes to the availability / schedule.

[Recently, there was an improvement made where the slug auto-generation is guaranteed to be global](https://github.com/thunderbird/appointment/pull/1700). However, collisions could still happen if users were to type the slug manually and said slug was being used by another user.

@MelissaAutumn suggested that we should make the username part of the slug so that two users could have a slug of `meeting` for instance.

davinotdavid:
## What changed?

- Added a migration that:
  - Adds `owner_id` to `schedules` table
  - Backfills `owner_id` to existing schedules
  - Replaces the global constraint on the slug only to a combined constraint with slug and owner_id
- Updates schedule model with the new `owner_id` where applicable
- Add tests

## Why?

Before, we had a DB constraint on the slug being global across all users which is ok except that the generation of the slug was not checking it globally. This already caused collisions on users and cryptic errors on the frontend.

Later, we've changed the generation to be considered globally, to avoid said collisions and added better errors on the frontend. However, since slugs can be typed manually, a user could add `meeting` as a slug essentially making it taken globally, which is not ideal.

This PR now makes it so that the DB constraint is actually on the owner and the slug together. Allowing two users to have the slug `meeting` for example. Slug generation also should never collide.

## How to test

- Go through FTUE with 2 users (if you're using the password AUTH_SCHEME, remember to change backend/src/appointment/routes/auth.py#L469 locally)
- Manually input `meeting` as a slug and save the availability, check that it has been saved correctly
- Logout and login with the other user, manually input `meeting` as a slug and save the availability, check that it has been saved correctly
- Generate a few slugs to make sure that auto-generation is also working as before

Bonus points, when running locally, please try to do the upgrade and downgrade of the migration instead of starting from scratch to make sure that there's no problems there!

## Applicable Issues

Closes https://github.com/thunderbird/appointment/issues/1706

### DIFF
--- a/backend/src/appointment/database/models.py
+++ b/backend/src/appointment/database/models.py
@@ -11,7 +11,20 @@ import uuid
 import zoneinfo
 from functools import cached_property
 
-from sqlalchemy import Column, ForeignKey, Integer, String, DateTime, Enum, Boolean, JSON, Date, Time, Uuid
+from sqlalchemy import (
+    Column,
+    ForeignKey,
+    Integer,
+    String,
+    DateTime,
+    Enum,
+    Boolean,
+    JSON,
+    Date,
+    Time,
+    Uuid,
+    UniqueConstraint,
+)
 from sqlalchemy_utils import StringEncryptedType, ChoiceType
 from sqlalchemy_utils.types.encrypted.encrypted_type import AesEngine
 from sqlalchemy.orm import relationship, as_declarative, declared_attr, Mapped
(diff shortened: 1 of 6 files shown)


rwood-moz (commented):
Hey, I am testing out your branch locally and am seeing some strange behaviour. I'm not sure if it's just a front-end refresh issue or something else with the actual values. To repro:

- I sign in to my local stack as my first user
- Changed my booking availability slug (to one I enter or just with refresh button)
- Save changes, all good, sign out
- Sign in to my local stack as my second user
- Go to availability page
- My booking page link slug (in customize and share fields) has changed to the one that I saved in my other local account
- Refresh the browser page, and the correct link slug appears for the account I'm currently signed in as

Maybe this is just an issue having two accounts on the local stack or some kind of refresh bug? I'll switch to main and see if I see the same behaviour there, maybe it has already existed and is unrelated to this PR.

rwood-moz:
> Maybe this is just an issue having two accounts on the local stack or some kind of refresh bug? I'll switch to main and see if I see the same behaviour there, maybe it has already existed and is unrelated to this PR.

Tried on main, yep happens there too, so not related to this patch.

rwood-moz (approved):
Code and new tests LGTM. I also tested the branch out locally and it worked great.

- Signed in as my first user, changed my booking link slug, saved
- Tested out booking page, able to book an appt
- Signed in as my second user, changed my booking link slug to the same as my first user, was able to save
- Tested out the booking page link, able to book an appt
- Refreshed the booking page link worked also

Issue #1678, opened 2026-05-19
Repeating Thundermail Calendar Events Cutting Off Last Day


madhattermattic:
## Describe the bug
I have a user that is setting repeated events in Thundermail calendar that last two, three, five days strait. However, those show up in Appointment after syncing as one, two, and four days only, cutting the last repeating day off.

Perhaps it's related to time zones?

## To Reproduce

1. Set a repeating event in Thundermail calendar that lasts more than two days
2. Wait for that event to sync to Appointment
3. View how many days on the calendar the repeating event happens

## Expected behavior
All repeated days should be included

## Actual behavior
The final day in the series does not show up in Appointment after sync

## Screenshots

<img width="862" height="268" alt="Image" src="https://github.com/user-attachments/assets/448797e8-3806-4e59-8612-c940bd9e01aa" />
<img width="1269" height="425" alt="Image" src="https://github.com/user-attachments/assets/a7442d61-9be4-44d2-b4cb-1bb36dfd7472" />

(Ticket 5516)

gz#5516

lisajill:
@madhattermattic In your reproduction, was this aesthetic only? Or did it impact availability? e.g. were people able to book into what would should have been blocked day/times?

madhattermattic:
@lisajill Apologies, I should have mentioned in this request that I was unable to reproduce, that's my mistake, I see now that I didn't mention that above. _These screenshots and replication steps are directly from the user._

I was unable to reproduce - Appointment works as expected for me, but that may be due to my time zone if the hunch is correct.

The user's calendar was able to be booked on the days that did not end up with the last day in the sequence - The events did not set for the full amount of days, they all ended up one day short. However, interestingly, single day events in the future set to the calendar just fine.

madhattermattic:
The user says this is still happening for them and has added two more screenshots to show their system time zone and calendar settings, if that helps:

<img width="513" height="743" alt="Image" src="https://github.com/user-attachments/assets/a9b17211-7fc9-446c-85db-5b331a225a18" />
<img width="773" height="347" alt="Image" src="https://github.com/user-attachments/assets/c477d080-ac7d-4c9b-b6d6-bfe5f2c132cc" />

madhattermattic:
We've got some more clues here that _seem_ to point to a time zone problem, not sure since I cannot recreate this. I'm going to directly quote the user on this next bit:

[...]

PR #1715, opened 2026-06-16
Fix remote events being cut off when adjacent to selected date border

devmount:
## What changed?

This fixes a bug where remote events were not shown on the dashboard when they were adjacent to the selected / active date range.

## Steps to test this

0. Don't checkout this branch yet. Have a daily repeated remote event ready that spans over at least a week
1. Go to availability settings and activate all weekdays
2. Go to the dashboard and see, that there are events from the repeated remote event, but the event on the far right is missing (otherwise the missing event might not be visible because of the day being blocked in general)
3. Now checkout this branch and refresh the dashboard
4. See, that the dashboard now shows all repeated events.

## Why?

This fixes missing events reported by users.

## Limitations and Notes

This is a fix for just *one* issue I was able to reproduce. There might be more issues since there are reports about missing events, but also for repeated remote events that are not adjacent as described here.

## Applicable Issues

Related to #1678 

## Screenshots

Before the fix:

image

After the fix:

image

### DIFF
--- a/frontend/src/views/DashboardView/components/WeekCalendar.vue
+++ b/frontend/src/views/DashboardView/components/WeekCalendar.vue
@@ -288,7 +288,7 @@ const filteredRemoteEventsForGrid = computed(() => {
 
   const { start, end } = props.activeDateRange;
   const remoteEventsWithinActiveWeek = props.events.filter((remoteEvent) =>
-    dj(remoteEvent.start).isBetween(start, end)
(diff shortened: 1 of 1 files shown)


davinotdavid (approved):
Thank you so much! 🎉 Works well on my local here following the test steps you've provided.

rwood-moz (approved):
Also pulled branch locally and verified the fix (thanks for the instructions!). Good find and nice fix. 😸

devmount:
Perfect, thanks you two! 🫶

PR #1724, opened 2026-06-24
Fix farthest booking calculation being cut by 1d

### LINKED ISSUE #1719: Changing the booking notice causes inappropriate availability
## Describe the bug

* With a booking notice of 7 days from Wednesday, it shows next Wednesday (July 24) basically totally available (not true)
* With a booking notice of 14 days from Wednesday, next Wednesdsay is correct, but the following Wednesday. (July 1) is again totally available (again, not true)
* Did not test other notice periods, likely the same bug?

## To Reproduce

1. https://appointment.tb.pro/availability
2. Booking window, switch between 7 and 14 days. Probably 21 and a month, too.
3. My tests were all done on a Wednesday

**Booking window 7 days**

Image

**Booking window 14 days**

Image

Image

## Expected behavior
Honors busy blocks no matter what the notice period is. My Wednesdays are very similar, and right now I could be double-booked on July 1 with a 14 day notice period.

## Actual behavior
The final day in the notice period shows entirely available and ignores the connected calendar's busies.

## System

- OS: Tahoe 26.4.1
- Browser Version Firefox 152.0

devmount:
## What changed?

This change fixes a bug where the last day of an availability time window wasn't considered for remote event retrieval.

## How to test

0. Have remote events at the last day of the configured booking window (e.g. if the window is 7d, have events at that exact day).
1. Don't checkout this branch yet and go to your bookers page and see, that the last day in the booking window is completely bookable
2. Now checkout this branch and reload the bookers page
3. See that there are blockers now on the last available day where the remote events are

## Why?

Prevention of bookings at potentially already blocked times.

## Limitations and Notes

It would be good to have tests for those event edge cases.

## Applicable Issues

Closes #1719

## Screenshots

None

### DIFF
--- a/backend/src/appointment/controller/calendar.py
+++ b/backend/src/appointment/controller/calendar.py
@@ -1058,7 +1058,7 @@ class Tools:
         now = datetime.now()
 
         earliest_booking = now + timedelta(minutes=schedule.earliest_booking)
-        farthest_booking = now + timedelta(minutes=schedule.farthest_booking)
+        farthest_booking = now + timedelta(days=1, minutes=schedule.farthest_booking)
 
         start = max([datetime.combine(schedule.start_date, schedule.start_time), earliest_booking])
         end = (


davinotdavid (commented):
Hmm, I don't see this problem happening in the `main` branch? Here's what I did to try to reproduce:

I've changed my Availability to 7 days and I have tried with all day events and regular events and it still shows blocked (correct) in the booking page 🤔

rwood-moz:
> Hmm, I don't see this problem happening in the `main` branch? Here's what I did to try to reproduce:
> 
> I've changed my Availability to 7 days and I have tried with all day events and regular events and it still shows blocked (correct) in the booking page 🤔

I was able to reproduce this on main, but I have my Availability set to 14 days. On main on the 14th day of availability my existing remote appointment is not blocked off on my bookee page:

Screenshot 2026-06-24 at 1 27 28 PM

But when I switched to this branch the remote appointment time is now blocked out on my bookee page, so I see your fix worked for my case:

Screenshot 2026-06-24 at 1 30 32 PM

devmount:
> @davinotdavid: I've changed my Availability to 7 days and I have tried with all day events and regular events and it still shows blocked (correct) in the booking page 🤔

Well that's interesting. The 7th day shows correctly blocked events? What happens if you switch to 14 and look at the last day?

> @rwood-moz: But when I switched to this branch the remote appointment time is now blocked out on my bookee page, so I see your fix worked for my case:

Good to know that you could reproduce that Rob. So it seems to only happen to some, there might be another variable at play here I'm not aware of 😅

davinotdavid (approved):
Interesting. I was able to reproduce this only with the 14-day availability block somehow. There might be something move to this but I'd say we can give this a try since it is an improvement over the previous code. Thanks!

PR #1733, opened 2026-07-02
Update E2E tests to ensure settings values are populated before proceeding

### LINKED ISSUE #1734: Improve E2E test stability when settings pages are loaded
Sometimes on production (and especially in BrowserStack) the Appointment Settings page loads but the actual values in the fields may take a couple of seconds to be populated. This can cause issues with the E2E tests as they might try to proceed when the values aren't actually populated yet. The tests should ensure the settings field values are populated before proceeding.

rwood-moz:
## What changed?
Improve the Appointment E2E tests by ensuring that the settings page field values are actually populated before proceeding.

# Why
Sometimes production Appointment can run quite slow (especially in BrowserStack) and after navigating to the Appointment Settings page, the DOM is loaded however the actual values in the settings fields aren't populated for several seconds. I previously filed an issue about Settings being slow to load in #1371. Here [is an example](https://automate.browserstack.com/projects/Thunderbird+Appointment/builds/Appointment+Nightly+Tests+Firefox+Desktop/241?testListView=spec&details=3713593457) BrowserStack video where you see the settings page opens but the actual field values aren't populated and the tests attempts to proceed.

Allow more time for the Settings pages to be populated with the actual Settings values; and check actual field values and ensure they're not blank, before proceeding to the test's next step.

## Applicable Issues
Fixes #1734.

## QA Log
- Ran the E2E tests on desktop locally
- Ran the updated E2E tests on stage on Firefox desktop [in BrowserStack](https://automate.browserstack.com/projects/Thunderbird+Appointment/builds/Appointment+Tests+Firefox+Desktop/42?tab=sessions&testListView=spec)
- Ran the updated E2E tests on stage on Android Chrome [in BrowserStack](https://automate.browserstack.com/projects/Thunderbird+Appointment/builds/Appointment+Tests+Android+Chrome/56?bls_localTimezone=true&testListView=spec)

### DIFF
--- a/test/e2e/pages/availability-page.ts
+++ b/test/e2e/pages/availability-page.ts
@@ -95,13 +95,21 @@ export class AvailabilityPage {
     await this.page.goto(APPT_AVAILABILITY_PAGE, { timeout: TIMEOUT_30_SECONDS });
     await this.page.waitForTimeout(TIMEOUT_10_SECONDS);
 
-    // the bookable toggle especially seems to take a long time to load in BrowserStack
-    if (! await this.bookableToggle.isEnabled()) {
-      await this.page.waitForTimeout(TIMEOUT_10_SECONDS);
-    }
+    // ensure that the bookable toggle is loaded/enabled
+    await expect
+      .poll(async () => (await this.bookableToggle.isEnabled()), {
+        timeout: TIMEOUT_30_SECONDS,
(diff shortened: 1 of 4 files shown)


davinotdavid (approved):
lgtm, ty! also, sad that it takes so long.. we definitely need to profile and improve performance :(

Issue #580, opened 2024-07-24
All day events bleed into the next day 


kewisch:
**Describe the bug**
All day events bleed into the next day

**To Reproduce**

1. Create an all day event on Friday in your calendar (in my case it was an all day OOO event on gcal)
2. Go to calendar in appointment.day
3. Observe that the event shows on both Friday and Saturday (depending on timezones)

**Expected behavior**
End dates are exclusive, and date-only events are in floating timezone. Make sure they are displayed as such.

**Actual behavior**
Event appears in multiple days 

**Screenshots**
![image](https://github.com/user-attachments/assets/c55ea63f-2f68-498d-9abf-8e810f8b4c26)

**System (please complete the following information):**

- OS: macOS
- Browser Version Firefox Nightly

**Additional context**
This happens in a few different calendar views

kewisch:
Sounds like the same issue! This isn't the event on the PTO calendar, but rather a custom "Out of Office" event I set for all day Friday so my calendar is blocked for the rest of Mozilla. I was wrong on this being a date-only event, it appears that Google makes this a midnight to midnight event in your timezone when exporting via ICS (and also via the API).

```
BEGIN:VEVENT
CREATED:20240603T122629Z
LAST-MODIFIED:20240603T123049Z
DTSTAMP:20240603T123049Z
UID:20t8vd6tshj89ek6tsoohg4h69@google.com
SUMMARY:Wellbeing Friday
STATUS:CONFIRMED
ORGANIZER;CN=Unknown Organizer:mailto:unknownorganizer@calendar.google.com
ATTENDEE;CN=me@thunderbird.net;PARTSTAT=ACCEPTED;CUTYPE=INDIVIDUAL;ROL
 E=REQ-PARTICIPANT;X-NUM-GUESTS=0:mailto:me@thunderbird.net
RRULE:FREQ=WEEKLY;BYDAY=FR;UNTIL=20240809T215959Z;WKST=SU
DTSTART;TZID=Europe/Berlin:20240705T000000
DTEND;TZID=Europe/Berlin:20240706T000000
CLASS:PUBLIC
SEQUENCE:0
TRANSP:OPAQUE
END:VEVENT
```
If you've fixed timezone conversion, then it should work.

MelissaAutumn:
Hey there @kewisch!

I haven't had a chance to test this, but I've just pushed up the latest changes to production. Can you give it another try?

Thanks!

kewisch:
[...]

PR #1744, opened 2026-07-15
Handle midnight to midnight all day events

devmount:
## What changed?

This PR adds a temporary fix for CalDAV servers serving all day events in datetime format with 00:00 time. The fix only applies if:

- **dtstart** and **dtend exist**
- both are of type **datetime** and have hours, minutes, seconds and ms set to **0**
- the diff of both is **exactly 24h**
- the CalDAV domain is **thundermail.com**

Also 12 tests were added to cover CalDAV events and this issue in particular.

## Why?

For some reason, Thundermail (Stalwart) serves events marked as all-day not in the usual date format, but as datetime. This is an attempt to fix that quickly while waiting for a possible upstream fix. On the other hand, there might always be other CalDAV servers handling it like this. So we can discuss keeping this fix longer, but there are trade-offs.

## How to test

0. Don't check out this branch yet and have a Thundermal CalDAV connection active with at least one all-day event
1. Check the dashboard and see, the event is not positioned on top of the column but actually spans over all rows as normal event (wrong).
2. Now check out this branch and reload the dashboard page
3. See, that the same event now is listed on top of the column (correct).

## Limitations and Notes

- This fix fails, if a DST switch happens on the event day
- This fix is currently scoped to only apply to thundermail.com domains

## Applicable Issues

Related to #580 

## Screenshots

Before the change:

image

After the change:

image

### DIFF
--- a/backend/src/appointment/controller/calendar.py
+++ b/backend/src/appointment/controller/calendar.py
@@ -652,6 +652,8 @@ class CalDavConnector(BaseConnector):
 
             if is_midnight_one_day_span(vevent) and has_domain(self.url, MIDNIGHT_SPAN_DOMAIN_WHITELIST):
                 all_day = True
+                start = start.replace(tzinfo=None)
+                end = end.replace(tzinfo=None)
             # -- fix end --
 
             events.append(
--- a/backend/src/appointment/controller/calendar.py
+++ b/backend/src/appointment/controller/calendar.py
@@ -608,6 +608,41 @@ class CalDavConnector(BaseConnector):
             # if start doesn't hold time information (no datetime), it's a whole day
             all_day = not isinstance(start, datetime)
 
+            # FIXME: Temporary Workaround for known CalDAV servers returning all day events
+            # as datetime spanning from midnight to midnight. This WILL FAIL, if a DST happens
+            # in between start and end times.
+            # -- fix start --
+            def is_midnight_one_day_span(vevent):
+                dtstart = vevent.dtstart.value
+                dtend = vevent.dtend.value
+
+                # Must both be actual datetimes (not plain dates) to have a meaningful "time"
+                if not isinstance(dtstart, datetime) or not isinstance(dtend, datetime):
+                    return False
+
+                starts_at_midnight = (
+                    dtstart.hour == 0 and dtstart.minute == 0 and
+                    dtstart.second == 0 and dtstart.microsecond == 0
+                )
+                ends_at_midnight = (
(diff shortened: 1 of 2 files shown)


backend/src/appointment/controller/calendar.py
```diff
@@ -608,6 +608,41 @@ def list_events(self, start, end):
...
+
+                return starts_at_midnight and ends_at_midnight and exactly_one_day
+            
+            def has_domain(url, domain):
+                hostname = urlparse(url).hostname
+                if hostname is None:
+                    return False
+                return hostname == domain or hostname.endswith('.' + domain)
+            
+            if not all_day and is_midnight_one_day_span(vevent) and has_domain(self.url, 'thundermail.com'):
```
davinotdavid: Can we expand the domain here to include `stage-thundermail.com` as well? Also, for local environment, we could even potentially return `True` directly for the `has_domain` method.

backend/src/appointment/controller/calendar.py
```diff
@@ -608,6 +608,41 @@ def list_events(self, start, end):
...
+                all_day = True
```
davinotdavid, line 654: During my local testing here with TB Desktop in the "Local Timezone" setting + Appointment having the same timezone, it turns out that the all day event was being set in the Dashboard's calendar **one day before** I have set in TB Desktop.

I believe this happens because of timezone conversion from the frontend side. SInce genuine all day events carry a timezone-less date, perhaps we should remove the `tzinfo` for the start and `end`:


```suggestion
                all_day = True
                start = start.replace(tzinfo=None)
                end = end.replace(tzinfo=None)
```

davinotdavid (changes requested):
Great job, thanks for implementing this! I have a few comments here before we are good to go.

For the timezone bit, please test on your end as well to make sure that it is not biased to my timezone.

backend/src/appointment/controller/calendar.py
devmount: Great suggestion, I'll rewrite this using a whitelist of domains, in case we have additions in the future

backend/src/appointment/controller/calendar.py
devmount: Great call, I will test this. I will also try to add a test covering this!

devmount:
@davinotdavid Thanks a lot for testing and your great suggestions. Mind giving this another spin?

davinotdavid (approved):
lgtm! thanks! 🎉

Issue #693, opened 2024-09-26
"Have an invite code" is confusing when you only need to enter an email


kewisch:
**To Reproduce**

1. Go to appointment.day
2. Click on Sign up for the beta
3. See text

**Expected behavior**
The heading talks about exciting reasons why people should really enter their email and continue and what they are about to experience.

**Actual behavior**
It says "Have an invite code? Enter it here", which is confusing if you can only enter an email

**Screenshots**
![image](https://github.com/user-attachments/assets/eabba186-eb74-4653-bf5f-94099a98c37f)

kewisch:
Also, while we're at it, "Your privacy is important to us" isn't really meaningful as it is. Should this be linking to the privacy policy? 

The only way to escape that dialog is to click "Plan less, do more". This might be on purpose if you want to capture people right there, but the usual pressing escape or clicking outside the dialog don't work for me.

devmount:
Good catches and great suggestions, thank you Philipp! We'll look into that.

@inattee What would be a good copy here? "Have an invite code? First enter your email here"?

@MelissaAutumn I'm wondering why the input field for the actual invitation code is missing at that point 😅 Does it appear when entering the email address?

inattee:
Yes good catch! I'll see what Mel decides about including the invite code field and proceed with the copy for this page.  I'd also love to spruce up the invite email with some excitement while we're at it! :)

MelissaAutumn:
I'm good with adding an x in the corner to dismiss this modal (aka return to home) but I'll just ping @laurelterlesky for a quick check.

And the header part is a bug it seems. Oops.

inattee:
@MelissaAutumn can you confirm what the input field(s) on that modal should be? Then I can make some copy suggestions for it. 

ps. I would add in addition to an x in the corner to close, clicking in the grey area outside the modal is always intuitive for closing it

laurelterlesky:
I can see from Mel's response that some of the issues here are due to a bug. 

The design file is located here: https://www.figma.com/design/L7WgGR3JfGJ6tzFOY1oKdJ/TBA_2024_Q2_Sign-Up?node-id=4503-10396&t=NjWgyOhzAJHy6Elg-1

(which includes the email that is sent when a person signed up)

[...]

PR #1748, opened 2026-07-21
feat(frontend): runtime SPA config via /config.js (build-time -> runtime)

aatchison:
## Summary

Parameterize the SPA so **one built bundle runs unchanged in every environment**, configured at **runtime** instead of build time, and add a **decoupled multi-arch GHCR image publish** for the EKS/Kargo path. Prerequisite for the Appointment-on-EKS matched-set promotion pipeline (thunderbird/platform-infrastructure#779, design in `docs/superpowers/specs/2026-07-20-appointment-eks-kargo-promotion-pipeline-design.md`).

> **Draft** — opened for the Appointment app team to own/complete/merge.

## Design principle: additive, existing pipeline untouched

The existing env-specific **ECS + S3/CloudFront** deploy (`deploy-staging.yml` / `deploy-production.yml`) is **left untouched**. The SPA source change is **backwards compatible on its own**: with an empty `/config.js`, the runtime accessor falls back to the `--mode`-baked `import.meta.env` values, so the existing S3 deploy behaves exactly as before. The EKS path gets images from a **separate, additive GHCR publish workflow** (no cloud creds, no Pulumi, no S3/ECS).

## Runtime config mechanism

- `src/config.ts` — typed accessor: reads `window.__APP_CONFIG__`, **falls back to `import.meta.env`** (dev/tests/existing-S3 keep working).
- `window.__APP_CONFIG__` is set by `/config.js`, loaded synchronously in `index.html` before the bundle.
- `/config.js` is supplied per environment **at run time, not rebuilt**: the EKS nginx entrypoint (`docker/docker-entrypoint.d/40-appointment-config.sh`) generates it from `APP_*` pod env. (Existing S3 keeps baking values via `--mode`; the empty `public/config.js` falls through to them.)
- Backend needs no change (FastAPI already reads runtime env).

## Additive GHCR image publish

- `publish-images.yml` — on merge to main, builds the **backend** and **frontend (nginx SPA)** as native amd64+arm64 manifest lists → GHCR (`ghcr.io/thunderbird/thunderbird-appointment[-frontend]`), tagged `:<sha>`/`:v<version>`/`:latest`, via `GITHUB_TOKEN`. Path-filtered so each image builds only when its dir changes.
- `build-multiarch-image.yml` — reusable builder (native runners, push-by-digest, `imagetools create` manifest list), mirroring thunderbird-accounts.

## Changes

- `src/config.ts` + `test/config.test.ts` (new); `public/config.js` (new); `index.html` loads it.
- `docker/docker-entrypoint.d/40-appointment-config.sh` (new); `deploy.dockerfile` builds env-agnostic + wires the entrypoint.
- Migrated all `import.meta.env.VITE_*` app-config call-sites to `config` (main.ts, App.vue, router.ts, utils.ts, 3 composables). `MODE`/`BASE_URL` left as Vite built-ins.
- `.github/workflows/{publish-images,build-multiarch-image}.yml` (new). **No change to `deploy-staging.yml` / `deploy-production.yml`.**

## Verified

`npm run lint` clean, `npm run test` **97 pass**, `npm run build` green, both new workflows parse. **Local container test** (built `deploy.dockerfile`, ran nginx): entrypoint generates a JSON-safe `config.js` from `APP_*` (a `"` in a value is correctly escaped → valid JS), `config.js` loads before the deferred module bundle, SPA serves with client-side-route fallback (`/booking/xyz` → 200), and `APP_API_UPSTREAM` rewrites the proxy target. This caught the Node 22→24 fix (npm 10 couldn't `npm ci` the lockfile).

## Reviewed

Ran a 20-lens parallel review and applied the fixes: Node 18→22 (Vite 8 build was broken), `jq`-safe `config.js` generation, matched-set build (both images every push so both carry `:<sha>`), moving tags (`:latest`/`:v`) gated to `main`, arm64-count guard before publish, fail-loud `assertConfigured()` on the container path, table-driven tests over all 15 config keys, and corrected/accurate comments. Backwards-compat of the untouched S3/ECS deploy, supply-chain/token scopes, and secrets exposure all reviewed clean.

## Remaining (for the app team / later phases)

- **EKS wiring (appointment-deploy repo / ExternalSecrets):** supply the `APP_*` env for the frontend nginx pod, and set **`APP_API_UPSTREAM`** to the backend Service DNS (the baked `127.0.0.1:5000` is a sidecar/dev default). Note `APP_API_URL` is host+`/api/v1` path, not just a host.
- **nginx resolves the static `proxy_pass` upstream at startup** — if the backend Service DNS isn't resolvable when the frontend pod boots, nginx crashloops. On EKS either guarantee the backend Service exists first, or switch `appointments.conf` to a `resolver` + variable `proxy_pass` so the name resolves per-request (deploy-repo refinement).
- **Supply chain (spec Phase 4/6):** cosign sign + SLSA attest the GHCR indexes; ECR pull-through cache — deliberately deferred to the later pipeline phases.
- **S3 `config.js` cache headers** if the S3 build ever adopts runtime config (not today; fallback covers it).
- Optional later: drop `--mode` from the existing frontend build and source config from `/config.js` there too.

Relates to thunderbird/platform-infrastructure#779, #693, #731.


frontend/deploy.dockerfile
```diff
@@ -3,26 +3,34 @@ FROM nginx:stable
...
-# Add Node 18 support
+# Add Node 22 (Vite 8 requires Node >=20.19; matches the CI Node used by the S3 build)
+# and jq (used by the runtime-config entrypoint to emit JSON-safe config.js).
 RUN apt-get update
-RUN apt-get install -y ca-certificates curl gnupg
+RUN apt-get install -y ca-certificates curl gnupg jq
 RUN mkdir -p /etc/apt/keyrings
 RUN curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg
-RUN echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_18.x nodistro main" | tee /etc/apt/sources.list.d/nodesource.list
+RUN echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_22.x nodistro main" | tee /etc/apt/sources.list.d/nodesource.list
```
davinotdavid: Just double checking with @rwood-moz around the node version & browserstack as that's my main concern about this.

aatchison: Thanks @davinotdavid! Clarifications on both fronts:

**Node version** — this `deploy.dockerfile` is the new, *additive* image for the EKS/GHCR path (built by `publish-images.yml`); it isn't used by the current stage/prod deploy, which builds the frontend on the CI runner's Node (`setup-node` → Node 24). So the Node here only affects the static Vite build — nginx serves the built assets at runtime, no Node involved. I've bumped it `22 → 24` to match CI: it's actually required, because Node 22's npm 10 rejects this `lockfileVersion: 3` lockfile as out-of-sync (`npm ci` EUSAGE), while npm 11 (Node 24) installs it cleanly. Verified the image builds + serves locally.

**Browserstack** — the stage e2e run against the deployed stage app in real browsers and are unaffected: the existing stage deploy is untouched (still `npm run build -- --mode`), and the SPA runtime-config change is backwards-compatible — with the committed empty `public/config.js`, the app falls back to the `--mode`-baked `import.meta.env` values exactly as before. The GHCR publish is decoupled and doesn't touch the stage pipeline.

Happy to walk through it with you and @rwood-moz if that helps.

aatchison: @rwood-moz — flagging you directly per David's note. The two questions are (1) the Node version (now `node_24.x`, matching CI, and only for this new **additive** EKS/GHCR image — the current stage/prod build is untouched) and (2) browserstack (stage e2e run against the deployed app and are unaffected, since the SPA change is backwards-compatible and the GHCR publish is decoupled from the stage pipeline). Would appreciate your read on whether that puts your concern to rest.

davinotdavid (approved):
lgtm but honestly I am not super familiar with the intricacies of the change when it touches the infra / devops land so I'd say let's give it a try!

reviewing the code diff seems correct / reasonable to me and running it locally didn't produce any errors on my end

aatchison:
Post-merge on `main`, both workflows are green: `publish-images` published the backend + frontend multi-arch (amd64+arm64) images to GHCR (validating the Node 24 build), and `deploy-staging` deployed stage + **browserstack e2e passed**. @rwood-moz — that covers the node-version + browserstack concern end-to-end on real stage.

rwood-moz:
> Post-merge on `main`, both workflows are green: `publish-images` published the backend + frontend multi-arch (amd64+arm64) images to GHCR (validating the Node 24 build), and `deploy-staging` deployed stage + **browserstack e2e passed**. @rwood-moz — that covers the node-version + browserstack concern end-to-end on real stage.

Nice! Yeah when running on BrowserStack we don't need to install the playwright browsers, that's where the issue is with Node 24 and Playwright 1.59.

PR #1754, opened 2026-07-22
Fix missing repeated event instances

### LINKED ISSUE #1678: Repeating Thundermail Calendar Events Cutting Off Last Day

devmount:
## What changed?

This PR changes the event retrieval for CalDAV calendars. We used to rely on `calendar.search(expand=True)`, assuming this would provide all instances of recurring events. But some CalDAV servers were cutting of the last instances when the recurring event was based on an end date. So this PR introduces a manually approach of expanding recurring events and modifies existing tests accordingly.

## Why?

Retrieving and processing recurring event instances from CalDAV servers more reliably.

## How to test

0. Have a recurring (e.g. daily) all-day event and a recurring normal event available in Thundermail and have the Thundermail CalDAV connected to Appointment. **Important: The issue only occurs when the recurring event was configured with an end date. Specifying a recurrence count was fine on my end.**
1. Don't checkout this branch yet and go to the dashboard.
2. See that the last occurrence of each event chain is missing
3. Now checkout this branch (and reload containers to clear event caches)
4. Reload the dashboard and see that the last occurrence now shows up for each event chain

## Limitations and Notes

- We now use `recurring_ical_events.of()` to expand events. This slightly changes the event processing, since we now get actual vevent instances directly instead of the caldav events returned by `calendar.search()`. That made some changes on existing tests necessary.
- The issue of multiple day events is out of scope was reported here: #1755
- Implementation was done by me. Tests were modified by Claude Code.

## Applicable Issues

Closes #1678 

## Screenshots

Before:

Image

After:

image

### DIFF
--- a/backend/src/appointment/controller/calendar.py
+++ b/backend/src/appointment/controller/calendar.py
@@ -614,8 +614,7 @@ class CalDavConnector(BaseConnector):
             tentative = status == EventStatus.TENTATIVE
 
             start = vevent['DTSTART'].dt
-            # get_duration grabs either end or duration into a timedelta
-            end = (start + vevent['DURATION'].dt if 'DURATION' in vevent else vevent['DTEND'].dt)
+            end = vevent['DTEND'].dt
             # if start doesn't hold time information (no datetime), it's a whole day
             all_day = not isinstance(start, datetime)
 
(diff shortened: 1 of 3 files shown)


davinotdavid (approved):
Nice, thanks! 🎉 

Manually tested here and can confirm that it does now include the end event for repeated events. Tested with TB Desktop, Accounts and Appointment all running locally and pointing to the same Thundermail CalDAV calendar.

davinotdavid:
Merging already as I plan to do a deploy today still!

PR #1758, opened 2026-07-22
Introspect OIDC token to fix automatic Thundermail CalDAV connection

### LINKED ISSUE #1757: In FTUE, Thundermail CalDAV auto-connect should use preferred_username
In local dev and when running `thunderbird-accounts` and `appointment` in tandem, the standard initial admin user (admin@example.org / admin) that gets created during the Accounts side boot up doesn't have a matching `email` and `preferred_username`.

This is a problem because if you try to go through FTUE in Appointment using the initial admin user, we will then create a Subscriber with an email that doesn't match the Thundermail CalDAV credentials.

We should follow the change done in https://github.com/thunderbird/appointment/pull/1689 to error proof the credentials here by introspecting the OIDC token and getting the username from there instead of Subscriber.email.

FYI: This is not a problem in stage / production since the `subscriber.email` matches the username auth for the CalDAV connection.

davinotdavid:
## What changed?

During the automatic Thundermail CalDAV connection, instead of using `Subscriber.email` for the auth, we are introspecting the OIDC token and using a fallback of `preferred_username` -> `username` -> `Subscriber.email`.

## Why?

We shouldn't rely on `Subscriber.email` to attempt to connect with the Thundermail CalDAV. The credentials should be derived / introspected from the OIDC token instead.

## How to test

This might be a bit troublesome to test but you'd need to run both `thunderbird-accounts` and `appointment` in your local dev environment and change the `.env` vars from `appointment` to point to `thunderbird-accounts` ports. Also don't forget to change the `AUTH_SCHEME` to `oidc`.

1. Navigate to `http://localhost:8090
2. Sign up with `admin@example.org / admin` (if this works, you have successfully setup both projects together)
3. In the FTUE, click on "Connect Thundermail Calendar" and click Continue
4. Check that there shouldn't be any errors

## Applicable Issues

Closes https://github.com/thunderbird/appointment/issues/1757
Related with https://github.com/thunderbird/appointment/pull/1689

### DIFF
--- a/backend/src/appointment/routes/caldav.py
+++ b/backend/src/appointment/routes/caldav.py
@@ -11,6 +11,7 @@ from redis import Redis
 from sqlalchemy.orm import Session
 
 from appointment import utils
+from appointment.controller.apis.oidc_client import OIDCClient
 from appointment.controller.calendar import CalDavConnector, Tools
 from appointment.database import models, schemas, repo
 from appointment.dependencies.auth import get_subscriber, get_bearer_token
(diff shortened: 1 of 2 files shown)


devmount (approved):
Code lgtm, tests are passing (in CI and locally for me too), CalDAV still works as expected when navigating the frontend.

I wasn't able to go through the tests steps though, since I don't have a working tb-accounts instance running locally 😇

PR #1759, opened 2026-07-22
Better guards for admin checks in get_admin_subscriber

### LINKED ISSUE #1760: Improve admin checks

davinotdavid:
Closes https://github.com/thunderbird/appointment/issues/1760

### DIFF
--- a/backend/src/appointment/dependencies/auth.py
+++ b/backend/src/appointment/dependencies/auth.py
@@ -213,8 +213,11 @@ def get_admin_subscriber(
     if not admin_emails or not user:
         raise InvalidPermissionLevelException()
 
-    admin_emails = admin_emails.split(',')
-    if not any([user.email.endswith(allowed_email) for allowed_email in admin_emails]):
+    allowed_emails = {email.strip().lower() for email in admin_emails.split(',') if email.strip()}
+
+    email = user.email.lower()
+    domain = email.rpartition('@')[2]
+    if email not in allowed_emails and f'@{domain}' not in allowed_emails:
         raise InvalidPermissionLevelException()
 
     return user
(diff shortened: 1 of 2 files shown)


Sancus:
The same issue still exists in utils.py:211 and fxa_client.py:97

backend/src/appointment/dependencies/auth.py
```diff
@@ -213,8 +213,11 @@ def get_admin_subscriber(
     if not admin_emails or not user:
         raise InvalidPermissionLevelException()
 
-    admin_emails = admin_emails.split(',')
-    if not any([user.email.endswith(allowed_email) for allowed_email in admin_emails]):
+    allowed_emails = {email.strip().lower() for email in admin_emails.split(',') if email.strip()}
+
+    email = user.email.lower()
+    domain = email.rpartition('@')[2]
+    if email not in allowed_emails and f'@{domain}' not in allowed_emails:
```
Sancus, line 220: This whole thing seems a bit verbose? And what's the purpose of checking the domain after an exact match check?

```
allowed_emails = {email.strip().lower() for email in admin_emails.split(',') if email.strip()}
if user.email.lower() not in allowed_emails:
    raise InvalidPermissionLevelException()
```

Sancus (changes requested):

backend/src/appointment/dependencies/auth.py
davinotdavid: ah this is to allow for "@thunderbird.net" (all users) too as we are using this for staging at the moment

Sancus: ah I see, nevermind then that's OK.

Sancus (approved):

PR #1768, opened 2026-07-24
Remove FXA

### LINKED ISSUE #1761: Remove deprecated FXA auth flows
We currently have quite a bit of code for `AUTH_SCHEME=fxa` which is deprecated and should be removed.

devmount:
## What changed?

This change removes all code related to the Firefox Accounts Authentication (FXA). See exceptions described below.

## Why?

Housekeeping

## How to test

1. Rebuild containers
2. Check FTUE
3. Check external connections (Google, CalDAV, Zoom)
4. Check dashboard and booking

## Limitations and Notes

- I didn't touch any file under `pulumi/` at all, since that is @ryanjjung's domain.
- There is an env var `FXA_ALLOW_LIST` we still actively use outside of fxa context, maybe we want to rename that in the future.
- I couldn't verify Login via Accounts still works, so that must be tested before merging this PR in

## Applicable Issues

Closes #1761


ryanjjung:
I love a PR with this many red lines! :)

The only thing that should go from the Pulumi code is the same environment variables you've removed from the .env files here. That's a task I can throw on my backburner if you like.

davinotdavid (approved):
Thanks for the great cleanup!

I think we should cut a ticket for removing the unused env vars from the pulumi side and also to rename the FXA_ALLOW_LIST to ADMIN_ALLOW_LIST or something to be less confusing.

I will rebase the PR to fix the conflicts and merge if that's ok.

PR #1776, opened 2026-08-04
Update dependency packages [August 4th 2026]

davinotdavid:
## What changed?

- Updated some package dependencies:
  - `eslint-import-resolver-typescript` from `4.3.1` to `4.4.5`
  - `jsdom` from `28.0.0` to `30.0.1`
  - `postcss` from `8.5.10` to `8.5.23`
  - `typescript-eslint` from `8.65.0` to `8.66.0`
  - `vitest` from `4.1.8` to `4.1.10`

There was an extra vulnerability coming from the `brace-expansion` package that needed to be resolved by an `overrides` in `pnpm-lock.yaml` unfortunately as the upstream packages were not updates just yet.

## Why?

- So that we don't have security vulnerabilities caused by separate packages!

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -42,23 +42,23 @@
     "@vue/test-utils": "2.4.11",
     "autoprefixer": "10.4.12",
     "eslint": "9.39.5",
-    "eslint-import-resolver-typescript": "4.3.1",
+    "eslint-import-resolver-typescript": "4.4.5",
     "eslint-plugin-import": "2.32.0",
     "eslint-plugin-vue": "10.0.0",
     "globals": "17.3.0",
-    "jsdom": "28.0.0",
+    "jsdom": "30.0.1",
     "msw": "2.2.14",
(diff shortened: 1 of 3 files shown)


markstos (commented):
Thanks for this. 

The `jsdom` version bump shows a major version bump from 28 to 30, possibly a breaking change

Did you check the changelog on this to confirm if it's breaking for us or we are relying on our e2e tests to catch a potential regression there?

davinotdavid:
@markstos ty for the review!

Both! For the `jsdom` the only breaking changes cited in their release notes are minimum Node version bumps which are not a problem for our CI / Docker workflows as we are using version 24.x and LTS respectively.

However, to make it a bit more explicit, I've changed it in the `package.json` to be 24.15.0, [matching what's on their release notes for the 30.0 version breaking change](https://github.com/jsdom/jsdom/releases/tag/v30.0.0).

markstos (approved):

Issue #722, opened 2024-10-21
Support Multiple Scheduling Options


ryanleesipes:
**Is your feature request related to a problem? Please describe.**
Right now I only have my "General Availability" link to send to folks, which I've defined as a 30 minute meeting 3 days out. There are times when I want to send an "Urgent" link that would allow folks to grab my time on my calendar as short as 12 hours in advance. Additionally, I would like to have options for different lengths of meetings.

**Describe alternatives you've considered**
The current system means I have to change the settings of my only "availability", which isn't ideal because what I have set there isn't for special circumstances, it is intended for normal meetings.

**Additional Context**
You see this functionality in other tools, we really need a name for these - should see what @laurelterlesky thinks...

MelissaAutumn:
Technically these should be alternate schedules. Which we could support with some work, but we'd need some UI/UX for displaying multiple schedules for a user to edit.

Along with this we'll need to define what link we want to show in the various non-schedule places (e.g. what schedule is used in the nav bar, or account settings.) Or we could simply link to a landing page which shows a list of available options to book from.

Just some thoughts and ideas!

PR #1777, opened 2026-08-04
Add logs for slug conflict + set a stable order for schedules

davinotdavid:
## What changed?

- Added a Sentry call to log extra data for conflicting slugs preventing a schedule / availability to be saved
- Added ordering for schedules while getting the subscriber's schedule in get_by_subscriber

## Why?

This really shouldn't happen so this is an attempt of solving a potential data inconsistency (a user with two schedules) but given that we only currently support a single schedule, a generated slug should never conflict. Alas, it seems to be happening so my assumption is that somehow there are two (or more) schedules for the affected user and the unstable ordering of the (supposedly single) schedule returned by the backend caused a slug clash.

Really unlikely but very curious case 🤔 

## Limitations and Notes

We don't prevent the creation of two schedules in the backend at the moment as I believe there was a plan to support multiple schedules (ref https://github.com/thunderbird/appointment/issues/722). This is not tackled in this PR.

I can confirm this assumption if one would directly connect to the production DB and run: `SELECT owner_id, COUNT(*) FROM schedules GROUP BY owner_id HAVING COUNT(*) > 1;`

## Applicable Issues

Related with https://github.com/thunderbird/appointment/issues/1772

### DIFF
--- a/backend/src/appointment/database/repo/schedule.py
+++ b/backend/src/appointment/database/repo/schedule.py
@@ -30,6 +30,7 @@ def get_by_subscriber(db: Session, subscriber_id: int) -> list[models.Schedule]:
         db.query(models.Schedule)
         .join(models.Calendar, models.Schedule.calendar_id == models.Calendar.id)
         .filter(models.Calendar.owner_id == subscriber_id)
+        .order_by(models.Schedule.id)
         .all()
     )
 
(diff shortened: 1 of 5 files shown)


ryanjjung:
Confirmed there are 68 accounts with multiple schedules.

ryanjjung (approved):

PR #1779, opened 2026-08-05
Bump version to v1.6.0 + upgrade cryptography to 50.0.0

davinotdavid:
Changelog of cryptography looks safe from v50.0.0 and 49.0.0
https://cryptography.io/en/latest/changelog/#v50-0-0


markstos (commented):
There are 3 backwards incompatible changes in the changelog besides dropping Mac and Windows support. For each of these, we need to confirm we are not impacted:

BACKWARDS INCOMPATIBLE: Removed the deprecated PUBLIC_KEY_TYPES, PRIVATE_KEY_TYPES, CERTIFICATE_PRIVATE_KEY_TYPES, CERTIFICATE_ISSUER_PUBLIC_KEY_TYPES, and CERTIFICATE_PUBLIC_KEY_TYPES type aliases. Use PublicKeyTypes, PrivateKeyTypes, CertificateIssuerPrivateKeyTypes, CertificateIssuerPublicKeyTypes, and CertificatePublicKeyTypes instead. These were deprecated in version 40.0.

BACKWARDS INCOMPATIBLE: [ChaCha20](https://cryptography.io/en/latest/hazmat/primitives/symmetric-encryption/#cryptography.hazmat.primitives.ciphers.algorithms.ChaCha20) now treats the first 4 bytes of the nonce as a 32-bit little-endian block counter (as defined in [RFC 7539](https://datatracker.ietf.org/doc/html/rfc7539.html)) and tracks the number of bytes processed. Attempting to encrypt or decrypt more data than the counter allows before it would overflow now raises a [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError) rather than silently diverging from RFC 7539. Setting the counter portion of the nonce to zero allows encrypting up to 256 GiB with a given nonce.

BACKWARDS INCOMPATIBLE: Loading an X.509 certificate whose ECDSA or DSA signature AlgorithmIdentifier contains encoded NULL parameters now raises a [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError). Such certificates are invalid, but older versions of Java emitted them; previously they loaded with a deprecation warning.

davinotdavid:
@markstos there are no matches for these directly. `cryptography` isn't imported anywhere in backend/src, backend/test, or backend/scripts it's a dependency pulled in transitively.

  1. PUBLIC_KEY_TYPES/PRIVATE_KEY_TYPES aliases: no matches anywhere. Not affected.
  2. ChaCha20 nonce/counter behavior: no ChaCha20 usage anywhere. Not affected.
  3. X.509 NULL-parameter signature rejection: no x509/certificate handling in the code. The only cert-adjacent flows are:
    - JWT auth (auth.py, fxa.py, routes/auth.py): local tokens use HS256 (symmetric, JWT_ALGO=HS256 in .env), and FXA tokens are verified via jwt.PyJWK (JWK-based RSA key, not an X.509
  cert), this path never parses a certificate's signature AlgorithmIdentifier.
    - Google integration (external_connection.py, calendar.py) uses google.oauth2.credentials.Credentials for OAuth token refresh over HTTPS, no id_token cert/JWKS verification in this repo.

markstos (approved):
QA notes look good.

Issue #1766, opened 2026-07-21
`/setup`'s “Connect ThunderMail Calendar” option returns “We couldn't connect to your calendar”.


beedell-roke:
#### How To Reproduce It

~~~JS
#!/usr/bin/env xdg-open
window.location.assign('https://appointment.tb.pro/setup');
document.querySelector('.radio-group > label:nth-child(1)').click();
	/* “ThunderMail Calendar” */
document.querySelector('.base').click();
	/* “Continue” */
~~~

#### The Actual behavior

<blockquote>

~~~HTML
<div data-v-8e47f696="" data-v-9ae2fd6d="" class="critical notice notice-bar notice-bar" data-testid="notice-bar">
    <div data-v-8e47f696="" class="notice-bar-content">
        <span data-v-8e47f696="" class="icon"><!----><!----><!---->
            <svg data-v-8e47f696="" aria-label="Critical" width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
[...]

PR #1783, opened 2026-08-10
Retry Thundermail CalDAV auto-connection if not account not provisioned yet + surface better errors

davinotdavid:
## What changed?

- We are now actually surfacing the errors that are being sent from Accounts side while attempting to do the CalDAV auto-connection.
- In case there's a specific error where the account hasn't been provisioned yet in the Stalwart side, we do 3 retries before showing the error to give it a chance to work. After that we now show a specific error message about the account still being provisioned.

> [!NOTE]  
> To test this locally, you will need to run Accounts <-> Appointment in tandem

In the video attached I show how to reproduce the problem and the fix in action!

## Why?

- We were swallowing the errors with the `response.raise_for_status()` and showing a generic error message which prevents us to further diagnose what exactly went wrong.
- There might be a race condition where the Stalwart account hasn't been provisioned yet while attempting to make the Thundermail CalDAV connection. In these cases, we should give a little bit of grace for retrying but ultimately showing a more meaningful error message.

## Limitations and Notes

- This might not entirely solve the problem in the "Applicable Issues" section but at least would provide us with a bit more clarity on the error message. There's a hunch that the problem might be intermittent instability as well of attempting to do the Thundermail CalDAV connection while Accounts is still booting up from a deployment for example.
- The notice-bar design / message looks a bit odd but that's the wrapping done in the [component level itself in services-ui](https://thunderbird.github.io/services-ui/?path=/story/components-noticebar--type). To see it, shrink the window enough and the default text in it will wrap.

## Applicable Issues

Related with https://github.com/thunderbird/appointment/issues/1766

## Screenshots

https://github.com/user-attachments/assets/f8df75c5-c2ca-4e58-89f6-343624953cec

### DIFF
--- a/backend/src/appointment/defines.py
+++ b/backend/src/appointment/defines.py
@@ -31,6 +31,13 @@ DEFAULT_CALENDAR_COLOUR = '#c276c5'
 # List of Google CalDAV domains
 GOOGLE_CALDAV_DOMAINS = ['googleusercontent.com', 'google.com', 'gmail.com']
 
+# Code Accounts returns when a subscriber's Thundermail mailbox hasn't finished provisioning yet
+# (see thunderbird-accounts mail/views.py `appointment_caldav_setup`). It's a transient race, not
+# a permanent failure, so Appointment retries a few times with backoff before giving up.
+ACCOUNTS_MAIL_NOT_READY_CODE = 'mail_account_not_ready'
+ACCOUNTS_CALDAV_SETUP_MAX_ATTEMPTS = 3
+ACCOUNTS_CALDAV_SETUP_RETRY_DELAY_SECONDS = 2
+
 # Resolves to absolute appointment package path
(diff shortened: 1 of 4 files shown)


rwood-moz (approved):
Nice! Worked great for me:

- I pulled the branch locally
- Blew away my local appointment db/cache
- Started up my local accounts stack and appointment stack
- Signed into local appointment, it wants to connect a calendar
- Signed into local accounts, went into /admin, blew away my local primary email (with alias existing)
- Back in Appointment I chose to connect my Thundermail caldav calendar, it failed (I could see it was taking some time for the retries) and displayed the error message

Image

- Then I went back into Accounts /admin, added back the primary email, saved
- Back in Appointment continued to add Thundermail calendar again
- This time it connected successfully, and I was able to continue through entire Appointment FTUE and setup Appointment successfully

The code LGTML (nice tests too).

PR #1788, opened 2026-08-12
Skip Google Calendar polling while a push channel is live (#1607)

### LINKED ISSUE #1607: Remove pulling google events from Google if we already have a watch / notification channel open
After https://github.com/thunderbird/appointment/pull/1529 gets merged in, we will have [watch / notification channels](https://developers.google.com/workspace/calendar/api/guides/push) open which receive webhook calls when events gets changed remotely on the Subscriber's Google Calendar.

Assuming that our DB data is correct and up-to-date, we can have a huge performance boost by not having to pull remove events if we already have a channel open for that calendar.

Paraphrasing @MelissaAutumn here:

> Could you additionally cut a ticket to remove pulling the google events from Google's end if they have a watch channel open? Because then we can just pull from our db and assume the db is up-to-date. It's a bit out of scope here but would be a huge perf boost.

Ref: https://github.com/thunderbird/appointment/pull/1529#pullrequestreview-4032158837

alikhan126:
Closes #1607.

The booking page issues an uncached `freebusy.query` per Google connection on **every** availability request: every page load, every slot re-check, every visitor. This takes that off the hot path when a push channel is live.

> **Note on scope:** the `bust_cached_events` fix is split out into #1796 at review request. It's still in this branch's diff until that merges, then I'll rebase and it drops out of here.

## Why not "read the local database instead"

The issue proposes dropping the polling because push notifications keep the local DB current. The first half is right, the second isn't. `sync_google_calendar_changes` only acts on changed events that match an existing `Appointment`; everything else is discarded. There is no local table of remote calendar events, so reading the local DB would show subscribers as free during real meetings.

Mirroring Google's events locally was the alternative, and I rejected it on two grounds. One is documented: [sync tokens aren't compatible with most filters](https://developers.google.com/workspace/calendar/api/guides/sync), so a `syncToken` mirror can't be bounded with `timeMin`/`timeMax` and is unbounded in time. The other is an observation I could not find documented either way: incremental sync appeared to return recurring *masters* carrying `RRULE` rather than expanded instances, even with `singleEvents=true`. Either way it means reimplementing RRULE expansion, timezones and cancellation tombstones, where every bug is wrong availability.

## What this does instead

Keep asking Google, cache the answer, let push invalidate it. Google stays authoritative for expansion and timezones, so no correctness is reimplemented.

`is_push_active()` is the single decision point and is deliberately one-directional. Saying "no" when push is fine costs one API call; saying "yes" when push is dead shows stale availability. Trust needs a channel that exists, isn't near expiry, holds a sync token, and synced recently. Anything unknown answers no.

Four layers, each degrading into the one below, the bottom one being today's behaviour:

1. **Push** invalidates within seconds of a change.
2. **Reconciliation** (15 min) bounds how long a dropped notification goes unnoticed.
3. **The gate** stops trusting push after 6h.
4. **TTL** (1h) caps staleness if invalidation is missed entirely.

A stale read can't cause a double booking: `request_schedule_availability_slot` busts the cache and re-validates against the remote calendar before accepting.

## API volume, honestly

Repeated availability requests now make **0** Google calls instead of one each. But reconciliation re-syncs every watched channel every 15 minutes regardless of activity, which is roughly **96 `events.list` calls/day/channel**, unconditionally.

So this is a clear win for a booking page with real traffic, and a **net increase** for a subscriber whose page is viewed a handful of times a day. I don't think that can be optimised away safely: a calendar that changed but sent nothing is indistinguishable from one that didn't change, so the sweep can't be made conditional on observed activity without an independent signal. Happy to set `GOOGLE_RECONCILE_INTERVAL_SECONDS` to whatever you prefer; `MAX_SYNC_AGE` (6h) just needs to stay comfortably above it.

## Bugs fixed on this path

**410 `fullSyncRequired` discarded changes.** The handler fetched a fresh token and returned, dropping the lost delta. Recovery now follows what the [sync guide](https://developers.google.com/workspace/calendar/api/guides/sync) prescribes: wipe the derived store, then do a full sync, then re-establish the token. A failed scan aborts rather than re-tokening, because a fresh token's chain starts *after* the blackout and would make the gap permanent.

**A redundant expiration refresh, removed rather than fixed.** Review caught that the `X-Goog-Channel-Expiration` header was being parsed as epoch milliseconds when it is an RFC-1123 string. Fixing the parse was the obvious response, but a channel's expiration is set at creation and never moves (Google has no renewal mechanism; a channel is replaced, not extended), so the header restated a value we already held and could not have changed. The refresh is gone.

**Sync failures looked like successful empty syncs.** `list_events_sync` swallowed non-410 errors and returned `([], None)`, which stamped the channel healthy, skipped the bust, and hid it from reconciliation (which filters on the same column). It propagates now and the task's existing autoretry handles it.

## What changed

| Area | Change |
|---|---|
| `controller/google_watch.py` | `is_push_active()`, `push_backed_calendar_ids()` |
| `controller/calendar.py` | Cached busy-time reads for push-backed calendars, gated on `GOOGLE_INVITE_ENABLED` |
| `routes/webhooks.py` | Records that a notification arrived (`last_notification_at`) |
| `tasks/google.py` | Cache bust on change, 410 recovery, `reconcile_google_channels` |
| `models.py` + migration `e5f6a7b8c9d0` | `last_synced_at`, `last_notification_at` |
| `celery_app.py` | Renewal hourly (was ~6 days), reconciliation every 15 min |

### Configuration

| Variable | Default | Meaning |
|---|---|---|
| `REDIS_EVENT_EXPIRE_SECONDS_PUSH` | `3600` | Cache TTL when push is live |
| `GOOGLE_RECONCILE_INTERVAL_SECONDS` | `900` | How stale a channel may get before re-sync |
| `GOOGLE_CHANNEL_RENEW_INTERVAL_SECONDS` | `3600` | Renewal cadence |
| `GOOGLE_FULL_RESYNC_WINDOW_DAYS` | `90` | Window re-scanned each way after a 410 |

All four are in `.env.example` and `.env.test`. `MAX_SYNC_AGE` (6h) and `CHANNEL_EXPIRY_MARGIN` (5 min) are constants in `google_watch.py`.

## Testing

482 passed / 1 skipped, `ruff` clean. Migration verified on a disposable Postgres: applies, idempotent, reverses cleanly.

The failure modes were exercised against a sandboxed Google Calendar twin during development. **Three bugs in the first revision came from trusting that sandbox over Google's documentation**: the expiration header format, message-number sequencing, and fractional-second timestamps. All three were caught in review by @MooseTheRebel and @davinotdavid against real Google traffic. Anything read off the wire now carries a docs link next to it and a test fed the literal string from those docs, and where the docs are silent I've said so rather than filling the gap from the sandbox.

I also re-ran the suite with each fix reverted one at a time to confirm the new tests fail rather than agreeing with the code. Two didn't and were wrong, so those were fixed too.

## Deliberately not handled

- **Channel coverage unchanged.** Still only a schedule's default calendar, only under `GOOGLE_INVITE_ENABLED`. Widening multiplies channel count and quota, which is a policy call beyond this issue.
- **Reconciliation still syncs channels serially in one beat task.** Fanning out per channel needs deduplication against concurrent webhook-triggered syncs, and there's no locking primitive in the codebase yet, so fan-out without one would make the existing race more likely. Better as a follow-up.
- **No distributed lock on concurrent syncs.** Residual cost is duplicate API calls, never incorrect state. `SELECT … FOR UPDATE` is the obvious step, but SQLite ignores it so it'd ship untested.
- **CalDAV untouched** (no push mechanism), **frontend untouched** (no API shape changed).

### DIFF
--- a/backend/src/appointment/controller/apis/google_client.py
+++ b/backend/src/appointment/controller/apis/google_client.py
@@ -1,7 +1,7 @@
 import logging
 import os
 import uuid
-from datetime import datetime
+from datetime import datetime, timezone
 from enum import StrEnum
 
 import sentry_sdk
@@ -21,6 +21,30 @@ from ...exceptions.calendar import (
     FreeBusyTimeException
 )
 from ...exceptions.google_api import GoogleScopeChanged, GoogleInvalidCredentials
+
+
+def parse_rfc3339_utc(value: str) -> datetime:
+    """Parse an RFC-3339 UTC timestamp from the Calendar API into a naive UTC datetime.
+
+    Google documents these fields as RFC-3339, which permits fractional seconds.
+    The previous strict '%Y-%m-%dT%H:%M:%SZ' parse raised ValueError on any value
+    that carried them, taking down the whole availability lookup; that was
+    observed against a contract-accurate Calendar twin returning '...T17:52:58.000Z'.
+
+    Returns naive UTC to match what the rest of the busy-time pipeline and the
+    CalDAV connector produce.
+    """
+    try:
+        parsed = datetime.fromisoformat(value)
+    except (TypeError, ValueError):
+        # Fall back to the historical strict format so any shape that used to
+        # parse still parses identically.
+        parsed = datetime.strptime(value, DATETIMEFMT)
+
+    if parsed.tzinfo is not None:
+        parsed = parsed.astimezone(timezone.utc).replace(tzinfo=None)
+
+    return parsed
 
 
 class SendUpdates(StrEnum):
@@ -193,8 +217,8 @@ class GoogleClient:
                         # Transform to datetimes to match caldav's behaviour
                         items += [
                             {
-                                'start': datetime.strptime(entry.get('start'), DATETIMEFMT),
-                                'end': datetime.strptime(entry.get('end'), DATETIMEFMT),
+                                'start': parse_rfc3339_utc(entry.get('start')),
+                                'end': parse_rfc3339_utc(entry.get('end')),
                             }
                             for entry in busy
                         ]
--- a/backend/src/appointment/controller/apis/google_client.py
+++ b/backend/src/appointment/controller/apis/google_client.py
@@ -1,7 +1,7 @@
 import logging
 import os
 import uuid
-from datetime import datetime, timezone
+from datetime import datetime
 from enum import StrEnum
 
 import sentry_sdk
@@ -21,30 +21,6 @@ from ...exceptions.calendar import (
     FreeBusyTimeException
 )
 from ...exceptions.google_api import GoogleScopeChanged, GoogleInvalidCredentials
-
-
-def parse_rfc3339_utc(value: str) -> datetime:
-    """Parse an RFC-3339 UTC timestamp from the Calendar API into a naive UTC datetime.
-
-    Google documents these fields as RFC-3339, which permits fractional seconds.
-    The previous strict '%Y-%m-%dT%H:%M:%SZ' parse raised ValueError on any value
-    that carried them, taking down the whole availability lookup; that was
-    observed against a contract-accurate Calendar twin returning '...T17:52:58.000Z'.
-
-    Returns naive UTC to match what the rest of the busy-time pipeline and the
-    CalDAV connector produce.
-    """
-    try:
-        parsed = datetime.fromisoformat(value)
-    except (TypeError, ValueError):
-        # Fall back to the historical strict format so any shape that used to
-        # parse still parses identically.
-        parsed = datetime.strptime(value, DATETIMEFMT)
-
-    if parsed.tzinfo is not None:
-        parsed = parsed.astimezone(timezone.utc).replace(tzinfo=None)
-
-    return parsed
 
 
 class SendUpdates(StrEnum):
@@ -217,8 +193,8 @@ class GoogleClient:
                         # Transform to datetimes to match caldav's behaviour
                         items += [
                             {
-                                'start': parse_rfc3339_utc(entry.get('start')),
-                                'end': parse_rfc3339_utc(entry.get('end')),
+                                'start': datetime.strptime(entry.get('start'), DATETIMEFMT),
+                                'end': datetime.strptime(entry.get('end'), DATETIMEFMT),
                             }
                             for entry in busy
                         ]
@@ -425,10 +401,14 @@ class GoogleClient:
                     if e.status_code == 410:
                         logging.info('[google_client.list_events_sync] Sync token expired, full sync needed')
                         return None, None
+                    # Anything else must propagate. Returning the partial items
+                    # collected so far is indistinguishable from "nothing
+                    # changed", which lets the caller stamp the channel healthy
+                    # and stop polling on the strength of a failed sync.
                     logging.warning(
                         f'[google_client.list_events_sync] Request Error: {e.status_code}/{e.error_details}'
                     )
-                    break
+                    raise
 
         return items, next_sync_token
 
--- a/backend/src/appointment/controller/calendar.py
+++ b/backend/src/appointment/controller/calendar.py
@@ -31,6 +31,7 @@ from sqlalchemy.orm import Session
 from .. import utils
 from ..defines import REDIS_REMOTE_EVENTS_KEY, DATEFMT, DEFAULT_CALENDAR_COLOUR, FALLBACK_LOCALE, APP_ENV_DEV
 from .apis.google_client import EventStatus, GoogleClient, ResponseStatus, SendUpdates
+from . import google_watch
 from ..database.models import CalendarProvider, BookingStatus
 from ..database import schemas, models, repo
 from ..controller.mailer import Attachment
@@ -38,6 +39,19 @@ from ..exceptions.calendar import TestConnectionFailed, RemoteCalendarAuthentica
 from ..exceptions.validation import RemoteCalendarConnectionError
 from ..l10n import l10n
 from ..tasks.emails import send_invite_email, send_pending_email, send_rejection_email, send_cancel_email
+
+
+def push_cache_expiry() -> int:
+    """Cache TTL, in seconds, for data covered by a live push channel.
+
+    This is a safety net, not the primary refresh mechanism -- a push
+    notification busts the cache as soon as the calendar changes. It bounds how
+    long a *silently* dropped notification can go unnoticed, so it trades
+    staleness against API volume. Reads are the only consumer; the booking write
+    path re-validates against Google regardless (see routes.schedule), so a stale
+    read cannot produce a double booking.
+    """
+    return int(os.getenv('REDIS_EVENT_EXPIRE_SECONDS_PUSH', 3600))
 
 
 class RemoteEventState(Enum):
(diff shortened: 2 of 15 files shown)


alikhan126:
@davinotdavid @kewisch @rwood-moz 
Need your eyes on this one. :shipit:

davinotdavid:
@alikhan126 First, thank you for the contribution / PR. We appreciate the interested in Appointment and the effort on making this.

However, this is quite big of a change for a core piece of the application and will require some time for the team to parse through for feedback. We will get to it when time allows for a deep dive in the changes.

In the meantime, you're welcome to join our Matrix chat for contributors. It's where we discuss development and help each other out:

https://matrix.to/#/#tb-services:mozilla.org

backend/src/appointment/routes/webhooks.py
```diff
@@ -76,6 +77,52 @@ def google_calendar_notification(
...
+        logging.warning(f'[webhooks.google_calendar] Unparsable X-Goog-Message-Number: {raw!r}')
+        return None
+
+
+def _parse_expiration(raw: str | None) -> datetime | None:
+    """Parse the X-Goog-Channel-Expiration header (Unix milliseconds, as a string)."""
+    if not raw:
+        return None
+    try:
+        return datetime.fromtimestamp(int(raw) / 1000, tz=timezone.utc)
```
MooseTheRebel: Hey @alikhan126, great work on this — the architecture here is really nice.

For the sake of helping with code review, I copied what you have and created a draft (mostly for discussion, your work is obviously the primary PR).  Anyhow, I drafted a mirror of your work via thunderbird/appointment#1794 . Just offering a friendly code review here — I'm hoping this is helpful and doesn't cause any difficulty or slow things down!

One thing I wanted to flag: `_parse_expiration` here parses `X-Goog-Channel-Expiration` as Unix milliseconds (`datetime.fromtimestamp(int(raw) / 1000, ...)`), but [Google's docs](https://developers.google.com/workspace/calendar/api/guides/push) describe this header as "expressed in human-readable format," with an example like `Tue, 19 Nov 2013 01:13:52 GMT` — not an epoch integer. If that's accurate, this would throw on every real notification and silently fall back to `None` via the `except` below, so the stored expiration would never get refreshed from push traffic.

Might be worth double-checking against a real Google push notification to confirm the header format, just in case! I ended up parsing it as an RFC-1123 date string (`email.utils.parsedate_to_datetime`) on my branch — happy to compare notes if useful.

backend/src/appointment/tasks/google.py
```diff
@@ -89,29 +147,120 @@ def sync_google_calendar_changes(self, channel_id: str):
...
+      2. re-scan a bounded forward window and re-apply RSVP/cancellation state,
+         so bookings that changed during the blackout are not missed,
+      3. establish a fresh token and re-stamp the watermark.
+
+    If step 3 fails the watermark is left alone, so is_push_active() ages the
+    channel out and read paths return to polling -- stale, never silently wrong.
+    """
+    log.info(f'[tasks.google] Sync token invalidated for channel {channel.channel_id}, running full resync')
+
+    _bust_subscriber_cache(calendar)
```
MooseTheRebel, line 251: Another small one from the same read-through (again, just hoping to be useful and not a bother!) - in `_recover_from_invalid_sync_token`, the cache gets busted first, then the resync scans a forward-only window (`now` → `now + window_days`).

Two thoughts on this, offered gently since I'm not sure which behavior is intended:

- Busting the cache before the resync/re-token step succeeds means a failed resync leaves the cache cold with no push-derived data to fall back on until the next cycle — whereas busting only after a fresh token is confirmed would keep the previous cache alive a little longer as a safety net.
- The forward-only window (`now` → `now + window_days`) wouldn't catch changes to events that started before `now` but are still ongoing or were modified during the blackout — a `now - window_days` → `now + window_days` range would cover that.

Neither of these seemed like they'd break the core guarantee (push-trusted calendars still fall back to polling once `is_push_active` goes false), so easily could be a non-issue depending on how Google's `list_events` windowing works elsewhere in the app. 

Again, I have to say you've done an awesome job here regardless — this whole push-caching design is a nice piece of engineering.

backend/src/appointment/routes/webhooks.py
davinotdavid: Thanks @MooseTheRebel ! I can verify that this is indeed not an epoch integer and throws in the current code.

I've tested this with a real Google account and it returns me a human-readable format like `Mon, 24 Aug 2026 20:15:26 GMT`.

alikhan126: @MooseTheRebel thanks for catching this and a really sharp catch.

backend/src/appointment/controller/apis/google_client.py
davinotdavid: If there's any failure that is not `status_code == 410`, the `list_events_sync` returns `([], None])`which makes `sync_google_calendar_changes`'s `changed_events` be `[]` which is not `None` so then it keeps the cache.

I believe we should bust the cache for any `status_code` here.

davinotdavid: If there's any failure that is not `status_code == 410`, the `list_events_sync` returns `([], None])`which makes `sync_google_calendar_changes`'s `changed_events` be `[]` which is not `None` so then it keeps the cache.

I believe we should `raise` here instead of `break` perhaps.

backend/src/appointment/routes/webhooks.py
```diff
@@ -76,6 +77,52 @@ def google_calendar_notification(
...
+        if message_number > previous_message_number + 1:
+            logging.warning(
+                f'[webhooks.google_calendar] Missed {message_number - previous_message_number - 1} '
+                f'notification(s) on channel {channel_id}; the sync token will cover the gap'
+            )
+        elif message_number <= previous_message_number:
+            logging.info(
+                f'[webhooks.google_calendar] Duplicate/out-of-order notification {message_number} '
+                f'on channel {channel_id} (already saw {previous_message_number}); syncing anyway'
+            )
```
davinotdavid: I don't think that these logs are worth surfacing. I don't see an action that we can do with this information and the numbers are not sequential anyway [according to the docs](https://developers.google.com/workspace/calendar/api/guides/push):

> X-Goog-Message-Number: Integer that identifies this message for this notification channel. Value is always 1 for sync messages. Message numbers increase for each subsequent message on the channel, but **they're not sequential**.

Testing locally gives me a wild variety of numbers there that got me confused for a little while more than it helped to be honest.

backend/src/appointment/controller/calendar.py
```diff
@@ -155,16 +230,65 @@ def __init__(
         if google_tkn:
             self.google_token = Credentials.from_authorized_user_info(json.loads(google_tkn), self.google_client.SCOPES)
 
+    def _push_backed_remote_ids(self, calendar_ids: list) -> set:
```
davinotdavid, line 242: This should be gated by the `GOOGLE_INVITE_ENABLED` flag as live channels are not created if the flag is not on so that we can save a few trips to the DB (even though those would return 0).

backend/src/appointment/controller/apis/google_client.py
```diff
@@ -23,6 +23,30 @@
...
+        parsed = datetime.fromisoformat(value)
+    except (TypeError, ValueError):
+        # Fall back to the historical strict format so any shape that used to
+        # parse still parses identically.
+        parsed = datetime.strptime(value, DATETIMEFMT)
+
+    if parsed.tzinfo is not None:
+        parsed = parsed.astimezone(timezone.utc).replace(tzinfo=None)
+
+    return parsed
```
davinotdavid: I don't think this is an existing problem. For me, Google Calendar's `freebusy.query` is sending exactly the strict format (for example `2026-08-19T14:30:00Z` in my testing now, no fractional seconds) so perhaps this was a problem with your twin-testing environment / Veris? 🤔 

In that case we can also remove the paired tests.

davinotdavid (changes requested):
In the future, I'd love to see issues broken down and tackled in separate PRs. For instance, this PR here aims to solve two different bits:

1. `bust_cached_events` potentially leaving stale entried / `scan_iter` instead of single `scan` batch.
2. Skip Google Calendar polling while a push channel is live

They are related / cross paths with each other but it would make for a much leaner PR review.

Also, even though there are defaults for the usage of the new env vars, we should add them to the various `.env` files.

backend/src/appointment/routes/webhooks.py
alikhan126: @MooseTheRebel 
You're both right that it was broken. I started to fix the parse, then realised the feature wasn't worth keeping.

A channel's expiration is set when the channel is created and never moves, since there's no renewal mechanism and a channel gets replaced rather than extended. We already store that value from the watch response, and `update_expiration` rewrites it when `renew_google_channels` swaps the channel. So the header was restating something we already had and that couldn't have changed. Refreshing from it wrote the same value on every notification, and mostly existed as a second chance to parse it wrong.

So `_parse_expiration` is gone rather than fixed, and `record_notification` just stamps `last_notification_at` now.

(The original mix-up, since it's an easy one to repeat: `setup_watch_channel` parses `expiration` off the watch API *response body*, where it genuinely is epoch millis. I copied that same parse onto the header, which is RFC-1123.)

backend/src/appointment/tasks/google.py
alikhan126: Thanks, both good ones.

Window: agreed, now `now-90d → now+90d`.

Bust ordering: keeping it as is, and I think the docs back it. The [sync guide](https://developers.google.com/workspace/calendar/api/guides/sync) says a 410 "should trigger a full wipe of the client's store and a new full sync", which is the order we have. `is_push_active` also isn't false at that point: a 410 doesn't touch `last_synced_at`, so the gate keeps trusting the channel for up to 6h and we'd carry on serving cache built from a delta we know we lost. Comment added on the function.

Your comment did turn up a real bug though. If the resync scan failed we swallowed the error and grabbed a fresh token anyway. That token's chain starts after the blackout, so anything changed during it is gone for good, and reconciliation can't help because nothing's left that knows it's missing. I had a test asserting that was correct, which is the worse half. It aborts and re-raises now, keeping the old token so the retry re-enters recovery instead of skipping the gap.

backend/src/appointment/controller/apis/google_client.py
alikhan126: You're right, and it's worse than just keeping the cache. `changed_events` comes back `[]`, so we skip processing, but `record_sync(db, channel, None)` still stamps `last_synced_at`. So `is_push_active()` stays true, the cache never gets busted, and `get_stale()` skips the channel too, since it filters on that same column. A channel that keeps failing to sync looks healthy indefinitely.

Changed to `raise`. The task's existing autoretry handles the retry, and once it gives up the stale watermark ages the channel out. Also added a check for a missing `nextSyncToken`, which was getting recorded as a successful empty sync.

My test had `list_events_sync.return_value = ([], 'token-v2')` hardcoded, so it was never going to catch this. It runs the real pagination loop with an actual HttpError now.

backend/src/appointment/routes/webhooks.py
alikhan126: Yeah, fair. "Not sequential" makes the gap check meaningless, it'd fire on normal delivery. I claimed the opposite in the description off the back of the sandbox I developed against, which does number them sequentially.

Removed the logging, `_parse_message_number`, the param, and the `last_message_number` column and its migration entry. Kept `last_notification_at`, labelled observability only.

backend/src/appointment/controller/apis/google_client.py
alikhan126: Fair enough, removed along with the tests.

Went back to the docs and my justification was wrong too: the `freebusy` reference doesn't specify a format for the response `start`/`end` fields at all, only for `timeMin`/`timeMax` on the request. So "RFC-3339 permits fractional seconds" was me filling a gap in the docs with whatever the sandbox emitted.

alikhan126:
Thanks both @MooseTheRebel  and @davinotdavid , this was a useful review. Everything's addressed, details in the threads.

On splitting: you're right, and I'll go with your preference @davinotdavid . The `bust_cached_events` fix is now #1796, and it's worth a look on its own, because checking the SCAN docs made it look worse than I'd described. MATCH is applied after a batch is retrieved, so when our keys are a small share of the DB the first call often comes back empty with a non-zero cursor, and the old code read that as "nothing cached" and deleted nothing at all. It stays in this branch's diff until #1796 merges, then I'll rebase and it drops out of here.

Env vars: added all four to `.env.example` and `.env.test`. Left out of the pulumi configs because the equivalent existing settings aren't there either and rely on the same code defaults, but happy to add them if you'd rather be explicit.

One bug neither of you caught, found chasing Moose's 410 comment. If the recovery scan failed we swallowed it and re-established the sync token anyway. The new token starts after the blackout, so anything changed during it was gone for good. An existing test of mine asserted that was correct. Both fixed.

| Item | Resolution |
|---|---|
| `X-Goog-Channel-Expiration` format | Refresh removed as redundant, not re-parsed |
| Forward-only resync window | Widened to `now±90d` |
| Bust cache after token success | Keeping as is, reasoning in thread |
| `list_events_sync` swallows errors | Raises now, plus a guard for missing `nextSyncToken` |
| Message-number logging | Removed, with the column and migration entry |
| `GOOGLE_INVITE_ENABLED` gating | Fixed, plus a related lookup in `list_events` |
| `parse_rfc3339_utc` | Removed with its tests |
| Env vars | Added to both `.env` files |
| Split the PR | Redis fix split out as #1796 |
| 410 recovery losing the delta | Recovery aborts instead of re-tokening |

Three of these share one cause. I developed against a sandboxed Google twin and took its behaviour as the contract, and there was no test anywhere near the header parsing: 39 tests, none at the layer where all three bugs were. Anything read off the wire now has a docs link beside it and a test using the literal string from those docs, and where the docs are silent I've said so instead of filling the gap from the sandbox. I also reran the suite with each fix reverted one at a time to check the new tests actually fail. Two of mine didn't, and were wrong.

Description is rewritten too, including an honest note that reconciliation costs roughly 96 `events.list` calls/day/channel, so this is a win for a busy booking page and a net increase for a quiet one.

482 passed / 1 skipped, ruff clean, migration checked on a throwaway Postgres.

markstos:
I'm closing this PR in favor of the PR series by @MooseTheRebel which builds on this with a co-author credit. 

While we need to work on an updated LLM contribution policy, I think the original submission here would be an example of the kind  which might not make the cut-- a 1,000+ line PR with multiple correctness problems and a description and even follow-up comment which all appear to be completely LLM-generated. I appreciate this section from the Rust LLM policy:

>  ## Mechanically copy-pasting LLM output is a waste of time

> We will often get people who respond to review comments by copy-pasting them into their LLM, then copy-pasting its response back onto GitHub. Bluntly: this is a waste of everyone's time. If we wanted an LLM's opinion, we could have asked it ourselves. We want to hear your thoughts, not a machine's.

> Moreover, this is a breach of trust between the reviewer and the author. Our assumption when we review is that we're talking to a real person who wants to do their best work. Pasting LLM text creates suspicion: does the author actually care? Is there a person here at all?

PR #1793, opened 2026-08-17
Fix crash parsing fractional-second timestamps in free/busy

MooseTheRebel:
Update timestamp parsing (supporting the ISO-8601 subset Python supports).  Also, added test coverage for both the happy path and several error paths.

Originally based on https://github.com/thunderbird/appointment/pull/1788 

<details><summary>Previously part of a stacked PR series</summary>
<p>

**1 of 4** in a stacked PR series mirroring @alikhan126's design in #1788 — huge credit to them for the original architecture (trust gate, cache-then-invalidate-on-push, reconcile sweep). This series splits it into one commit per PR for easier review.

## Series

1. thunderbird/appointment#1793 — Fix crash parsing fractional-second timestamps in free/busy **(current)**
2. MooseTheRebel/appointment#1 — Fix Redis cache-bust to always cover full keyspace
3. MooseTheRebel/appointment#2 — Add push-channel trust gate and wire sync pipeline
4. MooseTheRebel/appointment#3 — Serve busy-time reads from cache when push active

See thunderbird/appointment#1794 for the combined view of all 4 commits together, plus a writeup of two behavioral differences from #1788 worth a look.

</p>
</details>


markstos (approved):
Thanks for the contribution. Code, docs, and tests are all clear. 

I'm in favor of this. I was going to object about the long code comment, but then I read it and found the context useful. 

## QA Log

 - Review diff and test coverage
 - Reviewed what the new dep does https://pypi.org/project/hypothesis/  I'm cautious about adding new deps, but one does seem like it could add some testing value if we start to using it for some other things.

MooseTheRebel:
@markstos Sorry for the slow reply 😞 Please let me know if there's anything I can do to improve this PR. 

I'm happy to remove `hypothesis` and use the **pyproject.toml** and **uv.lock** from the `main` branch - I only added it to mimic 1788.  Normally I wouldn't adjust a project's dependencies.  

One idea, is that we could roll our own hypothesis-like classes (like the factories that you already have in `backend/test/factory/` ).  Of course, there would be limitations, but it could be a good stepping stone. 🙂 (Also note, the decision about hypothesis would impact https://github.com/MooseTheRebel/appointment/pull/2 ), but I'm happy to update that as well.

markstos:
About the dep: I agree about removing it. It seems like Pytest [parametrize](https://docs.pytest.org/en/stable/example/parametrize.html) could be used instead.

MooseTheRebel:
@markstos Updated this PR 🙂

backend/src/appointment/controller/apis/google_client.py
```diff
@@ -23,6 +23,30 @@
 from ...exceptions.google_api import GoogleScopeChanged, GoogleInvalidCredentials
 
 
+def parse_rfc3339_utc(value: str) -> datetime:
```
devmount: This function isn't exactly doing what the name tells. It actually parses ISO-8601 (well, the ISO-8601 subset Python supports), which is more general than RFC-3339. So rather than reflecting the use case, I would call it like what it actually does:

```suggestion
def parse_iso8601_utc(value: str) -> datetime:
```

Also can we move this to utils? It isn't exactly a standard made by Google, so I would rather have it available at a neutral place for other providers as well if needed.

backend/src/appointment/controller/apis/google_client.py
```diff
@@ -23,6 +23,30 @@
...
+    """Parse an RFC-3339 UTC timestamp from the Calendar API into a naive UTC datetime.
+
+    Google documents these fields as RFC-3339, which permits fractional seconds. The previous
+    strict '%Y-%m-%dT%H:%M:%SZ' parse raised ValueError on any value that carried them, taking
+    down the whole availability lookup; that was observed against a contract-accurate Calendar
+    twin returning '...T17:52:58.000Z'.
+
+    Returns naive UTC to match what the rest of the busy-time pipeline and the CalDAV connector
+    produce.
+    """
```
devmount: I suggest making that comment a bit more general and less verbose. Something like:


```suggestion
    """Parse an ISO-8601 timestamp (supporting RFC-3339, fractional seconds) into a naive UTC datetime.
    Useful when calendar providers return ISO-8601 conform timestamps instead of plain datetime strings."""
```

backend/src/appointment/controller/apis/google_client.py
```diff
@@ -23,6 +23,30 @@
...
+    try:
+        parsed = datetime.fromisoformat(value)
+    except (TypeError, ValueError):
+        # Fall back to the historical strict format so any shape that used to parse still
+        # parses identically.
```
devmount: I think this is a typical AI-context comment drawing in unnecessary time context. I would rephrase it:


```suggestion
        # Fall back to string parsed time in the defined default format
```

devmount (changes requested):
Thank you for fixing this, that's awesome! I tried that out locally and it worked great with my Google calendar 👏🏻 

However, I found some inconsistencies noted below. Let's discuss them or if you agree, please integrate them. Thanks so much for your time!

MooseTheRebel:
@devmount I've applied your suggestions , along with some other cleanup 🙂 

Thanks for the review!

devmount (approved):
lgtm now, thanks a lot!

Approving now to unblock the merge. But please wait with the merge until @davinotdavid or somebody confirmed all tests / linter etc. are still green 🙏🏻 (I didn't have the time anymore to check that)

devmount:
@aatchison @markstos For some reason, our validation workflows extended in #1774 still don't seem to trigger here.

MooseTheRebel:
> @aatchison @markstos For some reason, our validation workflows extended in #1774 still don't seem to trigger here.

If it helps over the next week or so, I'm happy to open a PR with a one character spellcheck fix or similar.

davinotdavid:
I've checked out this PR / branch locally and manually ran tests / code coverage / lint check and formatting check.

They all pass except for the formatting check `ruff format` but those also don't pass in `main` right now, unfortunately. I've opened up a PR for fixing that in the `main` branch here: https://github.com/thunderbird/appointment/pull/1806.

markstos:
@aatchison Can you check the workflow issue?

MooseTheRebel:
Hoping to be helpful, I've merged `main` into this branch and run `uvx ruff@latest check . --fix` in the `"backend/"` (it didn't change any files). 🙂

davinotdavid:
Workflows seems to be working after the rebase, thanks @MooseTheRebel ! Merging.

PR #1794, opened 2026-08-17
Mirror of #1788: Skip Google Calendar polling while a push channel is live

MooseTheRebel:
## Summary

This is a revised take on #1788 , but realistically you can use this purely a code review 🙂

Obviously, @alikhan126 deserves all the credit on the PR.  I mirrored the original design and implementation, which this branch builds directly on. The architecture here is the same: same trust gate (`is_push_active`), same cache-then-invalidate-on-push design, same reconcile sweep, same wiring into `get_busy_time`. I did run a bit of extra static analysis and formatting (`ruff`), so there may be some cosmetic differences.

Two real issues did surface in my review, and I want to flag both explicitly:

### 1. `_parse_expiration` in `webhooks.py` parses a different format

- **#1788**: treats `X-Goog-Channel-Expiration` as Unix milliseconds (`datetime.fromtimestamp(int(raw)/1000, ...)`)
- **This PR**: treats it as an RFC-1123 date string (`email.utils.parsedate_to_datetime(raw)`)

Google's [push notifications guide](https://developers.google.com/workspace/calendar/api/guides/push) documents this header as "Date and time of notification channel expiration, expressed in human-readable format," with the example `X-Goog-Channel-Expiration: Tue, 19 Nov 2013 01:13:52 GMT` — not an epoch integer. So the #1788 version would likely throw on every real notification and silently fall back to `None` (never refreshing the stored expiration from push traffic). This looks like a genuine bug in #1788 that got fixed somewhere along the way on the branch this PR was built from, rather than an intentional change I made on purpose. Worth double-checking against real Google push traffic either way.

### 2. `_recover_from_invalid_sync_token` — resync window and cache-bust ordering differ

- **#1788**: busts the cache first, then re-scans a forward-only window (`now` → `now + 90d`).
- **This PR**: re-scans a wider window (`now - 90d` → `now + 90d`), and only busts the cache after a new sync token is successfully obtained — so a failed token re-establishment skips the bust entirely.

Skipping the cache-bust after a failed token re-establishment is arguably fine in practice, since `is_push_active` would already be false at that point and we'd stop serving from cache regardless. Still, this is a real behavioral difference between the two PRs, not just a rewording, so I'm calling it out rather than assuming the two are equivalent.

### In summary

Neither of the above items break the feature's core guarantee (push-trusted calendars get served from cache, everything else polls), and all 484 tests pass here. 

I think both are worth a second look, and maybe some added test coverage — especially the first, since it's a correctness question against Google's real API. 

@alikhan126, I'm curious what you think?

## Alternatively

If you'd rather review this commit-by-commit, I broke the same 4 commits into a stacked PR series instead. Each one is based on the last, so its diff shows only that commit's changes. Step 1 targets `thunderbird/appointment:main` directly; Steps 2-4 target each other from my own fork (cross-repo PRs can't set a fork branch as the base, so this was the only way to keep the stack diffable):

- Step 1 (→ main): thunderbird/appointment#1793 — fractional-second timestamp fix
- Step 2 (→ Step 1): MooseTheRebel/appointment#1 — Redis cache-bust fix
- Step 3 (→ Step 2): MooseTheRebel/appointment#2 — push-channel trust gate
- Step 4 (→ Step 3): MooseTheRebel/appointment#3 — cached busy-time reads

If this is the approach you'd rather take, here's how Steps 2-4 would get brought in once thunderbird/appointment#1793 merges: once it lands, I'll retarget MooseTheRebel/appointment#1's base from `feat/1607-01-fractional-timestamp-fix` to `thunderbird/appointment:main` (a GitHub UI/CLI edit, no new commits needed) and open it against the upstream repo so it can be merged the same way thunderbird/appointment#1793 was. Steps 3 and 4 would then follow the same pattern in order — each retargeted to `main` and merged only after the one before it lands — so the stack unwinds one PR at a time instead of needing to move all four at once.


MooseTheRebel:
Closing this PR, in favor of the series that starts in https://github.com/thunderbird/appointment/pull/1793

PR #1801, opened 2026-08-20
test(e2e): non-destructive @deployment-analysis sign-in smoke

aatchison:
A freight-verification smoke for the EKS/Kargo pipeline (platform-infrastructure#1046), mirroring the accounts one (#1146). Kargo clones this repo at `main` and runs `deployment-analysis-e2e` against a freshly-promoted tb-dev deploy; red = freight not verified.

## What it does
OIDC sign-in via the tbpro realm (reusing `navigateToAppointmentAndSignIn`), then asserts the dashboard renders. **Non-destructive** — no booking, no email, no settings/availability mutation, and deliberately not `@stage-sanity`.

## Changes
- `tests/desktop/deployment-analysis-signin.spec.ts` — the `@deployment-analysis` spec
- `playwright.config.ts` — standalone `deployment-analysis` project (firefox-only, no `desktop-setup` dep, no `storageState`)
- `package.json` — `deployment-analysis-e2e` (+ `-headed`) scripts
- `.env.tb-dev.example` — the four vars it consumes

## Verified
`pnpm install --frozen-lockfile` + `playwright test --list --project=deployment-analysis` → discovers exactly this one test.

PR #1802, opened 2026-08-20
fix(e2e): @deployment-analysis asserts authenticated landing, not the NavBar

aatchison:
Follow-up to #1801. Running the smoke as a real Kargo freight gate — in-cluster, against the shared tb-dev test user — surfaced a false-FAIL that the local runs (yours and mine) missed because they used an already-set-up user.

## What happened
A user who hasn't completed appointment first-time-setup lands on **/setup**, where the desktop NavBar (`.nav-items-container`) isn't rendered — so the `navBarDashboardBtn` assertion timed out on a *healthy* deploy. Set-up users land on `/dashboard` (NavBar present), which is why it passed locally.

Confirmed in-cluster: with a fresh user, sign-in succeeds and lands on `https://appointment.tb-dev.thunderbird.dev/setup`.

## Fix
Assert the authenticated **landing route** (`/dashboard` or `/setup`) instead of a dashboard DOM element. Both are auth-gated — a failed sign-in reaches neither — so this proves OIDC sign-in succeeded independent of FTUE state (which a shared/Neon-reset test user needs). Still fail-closed.

## Verified
In-cluster AnalysisRun against this branch with the shared user landing on `/setup`: **1 passed (21.7s)**.

Refs platform-infrastructure#1046

### DIFF
--- a/test/e2e/tests/desktop/deployment-analysis-signin.spec.ts
+++ b/test/e2e/tests/desktop/deployment-analysis-signin.spec.ts
@@ -32,12 +32,12 @@ test.describe('deployment analysis sign-in smoke', () => {
     // but that title is also present on the signed-OUT app, so it is NOT sufficient on its own.
     await navigateToAppointmentAndSignIn(page);
 
-    // The real gate: an element that ONLY renders for an authenticated subscriber. The desktop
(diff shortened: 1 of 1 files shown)


rwood-moz (approved):
Ah good call, I also assumed the test user would already have been setup in advance. LGTM!

PR #1803, opened 2026-08-21
Fix Redis cache invalidation for multi-page case

### LINKED ISSUE #1798: Cache invalidation can leave stale Redis entries because the SCAN cursor is ignored
`BaseConnector.bust_cached_events()` makes one Redis `SCAN` call beginning at cursor 0, deletes the keys returned in that page, and ignores the returned cursor.

The problem is that Redis does not guarantee that one `SCAN` call returns all matching keys. So, matching cached entries may remain accessible until their TTL expires.

In particular, the booking path expects its post-invalidation availability check to be authoritative so the current behavior can result in stale entries being served.

 **Implementation considerations**

 The smallest fix is to iterate `SCAN` until its cursor returns to zero. A complete iteration is O(N) in the total Redis keyspace, although it is incremental and does not block Redis like `KEYS`.

If The `O(N)` scan in production becomes to costly, there are longer term cache redesigns that could be considered, such as generate-based invalidation.

## Related PRs

There's already a PR opened for this fix here:
https://github.com/MooseTheRebel/appointment/pull/1

I went ahead and and opened an issue in case we wanted to discuss alternate cache designs /now/ but it seems safe to take a "good now, better later" approach and applier the straightforward fix linked above and defer cache invalidation redesign until later. 

CC: @MooseTheRebel

markstos:
## What changed and why?

 Fixes #1798 for reasons described there-in.

## Notes

This has been a collaboration with @MooseTheRebel who in turn built
on work by @alikhan126.

## QA log

 * Added test coverage

### DIFF
--- a/backend/src/appointment/controller/calendar.py
+++ b/backend/src/appointment/controller/calendar.py
@@ -29,7 +29,14 @@ from enum import Enum
 from sqlalchemy.orm import Session
 
 from .. import utils
-from ..defines import REDIS_REMOTE_EVENTS_KEY, DATEFMT, DEFAULT_CALENDAR_COLOUR, FALLBACK_LOCALE, APP_ENV_DEV
+from ..defines import (
+    REDIS_CACHE_BUST_BATCH_SIZE,
+    REDIS_REMOTE_EVENTS_KEY,
+    DATEFMT,
+    DEFAULT_CALENDAR_COLOUR,
+    FALLBACK_LOCALE,
+    APP_ENV_DEV,
+)
 from .apis.google_client import EventStatus, GoogleClient, ResponseStatus, SendUpdates
 from ..database.models import CalendarProvider, BookingStatus
 from ..database import schemas, models, repo
@@ -119,9 +126,9 @@ class BaseConnector:
 
         deleted = 0
         batch = []
-        for key in self.redis_instance.scan_iter(match=match, count=500):
+        for key in self.redis_instance.scan_iter(match=match, count=REDIS_CACHE_BUST_BATCH_SIZE):
             batch.append(key)
-            if len(batch) >= 500:
+            if len(batch) >= REDIS_CACHE_BUST_BATCH_SIZE:
                 deleted += self.redis_instance.delete(*batch)
                 batch = []
         if batch:
--- a/backend/src/appointment/controller/calendar.py
+++ b/backend/src/appointment/controller/calendar.py
@@ -105,19 +105,14 @@ class BaseConnector:
 
         return True
 
-    def bust_cached_events(self, all_calendars=False) -> int:
+    def bust_cached_events(self, all_calendars=False) -> bool:
         """Delete cached events for a specific subscriber/calendar.
         Optionally pass in all_calendars to remove all cached calendar events for a specific subscriber.
 
-        Walks the whole keyspace with scan_iter -- a single SCAN batch is not guaranteed to be
-        complete, so a bust could silently leave entries behind. Two callers depend on this being
-        exhaustive: the booking write path re-checks availability immediately after busting, and
-        push notifications use it to invalidate on change.
-
-        Returns the number of keys deleted.
+        Returns whether any entries were deleted.
         """
         if self.redis_instance is None:
-            return 0
+            return False
 
         timer_boot = time.perf_counter_ns()
         match = f'{REDIS_REMOTE_EVENTS_KEY}:{self.get_key_body(only_subscriber=all_calendars)}:*'
@@ -134,7 +129,7 @@ class BaseConnector:
 
         sentry_sdk.set_measurement('redis_bust_time', time.perf_counter_ns() - timer_boot, 'nanosecond')
 
-        return deleted
+        return deleted > 0
 
 
 class GoogleConnector(BaseConnector):
--- a/backend/src/appointment/controller/calendar.py
+++ b/backend/src/appointment/controller/calendar.py
@@ -105,28 +105,36 @@ class BaseConnector:
 
         return True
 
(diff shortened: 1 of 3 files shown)


backend/src/appointment/controller/calendar.py
```diff
@@ -105,28 +105,31 @@ def put_cached_events(
...
-        if len(ret[1]) == 0:
-            return False
-
-        # Expand the list in position 1, which is a list of keys found from the scan
-        self.redis_instance.delete(*ret[1])
+        match = f'{REDIS_REMOTE_EVENTS_KEY}:{self.get_key_body(only_subscriber=all_calendars)}:*'
+
+        deleted = 0
+        batch = []
+        for key in self.redis_instance.scan_iter(match=match, count=500):
```
devmount: Why the 500? Is this a common page size?

devmount (commented):
Just commenting for now since I'm not really a redis expert: Thanks so much for this! Worked fine locally for me. Just one little question below for understanding.

backend/src/appointment/controller/calendar.py
MooseTheRebel, line 49:
```suggestion
# Number of keys to SCAN per Redis round-trip, and the max batch size passed to a single DELETE call.
_REDIS_SCAN_BATCH_SIZE = 500

```

backend/src/appointment/controller/calendar.py
MooseTheRebel: Maybe good to swap both uses of 500 for a module-level variable (like `_REDIS_SCAN_BATCH_SIZE`) https://github.com/thunderbird/appointment/pull/1803/changes#r3832128882 ?

backend/src/appointment/controller/calendar.py
```diff
@@ -105,28 +105,31 @@ def put_cached_events(
...
+            batch.append(key)
+            if len(batch) >= 500:
```
MooseTheRebel: See module-level variable idea (`_REDIS_SCAN_BATCH_SIZE`) https://github.com/thunderbird/appointment/pull/1803/changes#r3832128882 ☝️

backend/test/unit/test_cache_bust.py
```diff
@@ -0,0 +1,50 @@
...
+        prefix = f'{REDIS_REMOTE_EVENTS_KEY}:{connector.get_key_body(only_subscriber=True)}'
+        target_keys = [f'{prefix}:{i}' for i in range(501)]
+        redis_instance = EmptyFirstPageRedis(target_keys, ['unrelated-key'])
+        connector.redis_instance = redis_instance
+
+        deleted = connector.bust_cached_events(all_calendars=True)
+
+        assert deleted is True
+        assert redis_instance.store == {'unrelated-key': 'value'}
+        assert redis_instance.delete_batch_sizes == [500, 1]
```
MooseTheRebel, line 50: Just an idea, you could add this test to guard (future / regression) against accidentally wiping the wrong calendar's cache:

```suggestion
        assert redis_instance.delete_batch_sizes == [500, 1]
        
    def test_all_calendars_false_only_busts_that_calendar(self):
        subscriber_id = 1
        target_calendar = BaseConnector(subscriber_id=subscriber_id, calendar_id=10)
        other_calendar = BaseConnector(subscriber_id=subscriber_id, calendar_id=20)

        target_prefix = f'{REDIS_REMOTE_EVENTS_KEY}:{target_calendar.get_key_body(only_subscriber=False)}'
        other_prefix = f'{REDIS_REMOTE_EVENTS_KEY}:{other_calendar.get_key_body(only_subscriber=False)}'
        target_keys = [f'{target_prefix}:{i}' for i in range(3)]
        other_keys = [f'{other_prefix}:{i}' for i in range(3)]

        redis_instance = EmptyFirstPageRedis(target_keys, other_keys)
        target_calendar.redis_instance = redis_instance

        deleted = target_calendar.bust_cached_events(all_calendars=False)

        assert deleted is True
        assert redis_instance.store == dict.fromkeys(other_keys, 'value')
        
```

backend/src/appointment/controller/calendar.py
markstos: Good callout and recommendation. Yes, this magic number should be replaced with a module-level constant, `CACHE_BUST_BATCH_SIZE`.

markstos: @devmount to answer the "why", 500 is a  compromise between round trips and oversized commands.

markstos: I've pushed a commit to address this.

backend/src/appointment/controller/calendar.py
markstos: Good direction. In this project, the convention is to use `defines.py` with no leading underscore. I've pushed a commit with that direction:

https://github.com/thunderbird/appointment/blob/main/backend/src/appointment/defines.py

devmount: Perfect, thanks Mark!

backend/src/appointment/controller/calendar.py
devmount: I see, thanks a lot for making that a proper constant.

backend/test/unit/test_cache_bust.py
markstos: It's an interesting idea, but also relatively low-risk to overprune.

Because the current design using linear scans is O(n), a future design may go a different direction with cache design with something like using a generational key to allow O(1) pruning. I imagine at the time we are doing a cache overhaul, we'll consider adding more test coverage then, as especially as LLMs are so eager to write coverage for changes and so fast at doing it.

devmount (approved):
Lgtm!

PR #1804, opened 2026-08-25
Remove the time format setting

### LINKED ISSUE #901: 24hr format not applied to availability start and end time settings

### LINKED ISSUE #1795: TIME FORMAT
## System

- OS: 
- Browser Version

devmount:
## What changed?

This PR removes the time format setting from the Users settings page completely. Also 5 tests were added to ensure the correct auto-detection of the browsers time format (generated by Claude Code):

- 12h locale (en-US) -> hh:mma
- 24h locale (de-DE) -> HH:mm
- en-GB -> HH:mm (catches the "English != 12h" edge case)
- invalid locale -> falls back to config.defaultHourFormat when set
- invalid locale, no config -> falls back to hh:mma (12h default)

[Davi] Updated `WeekCalendar` component to make use of the time format setting.

## Why?

https://github.com/thunderbird/appointment/issues/901#issuecomment-4852916770

## Limitations and Notes

- I found one dead language key on the way (unrelated) and deleted it too.

## Applicable Issues

Closes #901
Closes https://github.com/thunderbird/appointment/issues/1795

## Screenshots

No time format option anymore:

image

24 hours:

Screenshot 2026-08-25 at 10-55-33 Thunderbird Appointment

AM / PM:

Screenshot 2026-08-25 at 10-53-44 Thunderbird Appointment

### DIFF
--- a/backend/src/appointment/database/models.py
+++ b/backend/src/appointment/database/models.py
@@ -97,11 +97,6 @@ class ColourScheme(enum.Enum):
     light = 'light'
 
 
-class TimeMode(enum.Enum):
-    h12 = 12
-    h24 = 24
-
-
 # Use ISO 8601 format to specify day of week
 class IsoWeekday(enum.Enum):
     monday = 1
@@ -177,7 +172,6 @@ class Subscriber(HasSoftDelete, Base):
     language = Column(encrypted_type(String), nullable=False, default=FALLBACK_LOCALE, index=True)
     timezone = Column(encrypted_type(String), index=True)
     colour_scheme = Column(Enum(ColourScheme), default=ColourScheme.system, nullable=False, index=True)
-    time_mode = Column(Enum(TimeMode), default=TimeMode.h12, nullable=False, index=True)
     start_of_week = Column(Enum(IsoWeekday), default=IsoWeekday.sunday, nullable=False, index=True)
 
     # Only accept the times greater than the one specified in the `iat` claim of the jwt token
--- a/backend/src/appointment/database/schemas.py
+++ b/backend/src/appointment/database/schemas.py
@@ -27,7 +27,6 @@ from .models import (
     ExternalConnectionType,
     MeetingLinkProviderType,
     ColourScheme,
-    TimeMode,
     IsoWeekday,
 )
 from .. import utils, defines
@@ -322,7 +321,6 @@ class SubscriberIn(BaseModel):
     secondary_email: str | None = None
     language: str | None = FALLBACK_LOCALE
     colour_scheme: ColourScheme = ColourScheme.system
-    time_mode: TimeMode = TimeMode.h12
     start_of_week: IsoWeekday = IsoWeekday.sunday
 
 
--- a/backend/src/appointment/migrations/versions/2025_01_15_1340-4a15d01919b8_add_config_fields_to_subscribers_table.py
+++ b/backend/src/appointment/migrations/versions/2025_01_15_1340-4a15d01919b8_add_config_fields_to_subscribers_table.py
@@ -5,10 +5,11 @@ Revises: 0c99f6a02f3b
 Create Date: 2025-01-15 13:40:12.022117
 
 """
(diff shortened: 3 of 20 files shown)


davinotdavid (changes requested):
Removal code LGTM, thanks!

However, the main point of the issues linked was that changing the time format didn't actually update the week calendar view. If we are removing the option, we should guarantee that the time slot labels are updated accordingly.

I believe this would be done somewhere in `frontend/src/views/DashboardView/components/WeekCalendar.vue` around L222.

Also, since this is also a visual change, would you mind sharing screenshots as part of the QA process of this PR?

davinotdavid (approved):
LGTM! Since the change that I was referring to was small / straightforward and @devmount synced with me on this, I've added a new commit to this PR.

Looking forward to land this patch this week!

rwood-moz:
Hey @devmount

So just so I understand this change correctly, the 24 hour user preference setting has been removed, and instead Appointment will read the user's local time format preference from their browser locale settings?

So in Firefox about:config I set the pref `intl.regional_prefs.use_os_locales` to `true` and restarted Firefox. Then in my MacBook `System Settings => General => Date & Time` I turned on the option toggle for `24 hour time`.

However when I sign back into Appointment (local dev stack with this branch) the calendar still shows me 12 hour time format; and my bookee share link also shows 12 hour format. Am I misunderstanding this change or maybe doing something wrong? Thanks :)

davinotdavid:
@rwood-moz Hi Rob! I can reply to this one perhaps!

So the setting is now derived from the browser's locale, using the [Intl.DateTimeFormat.prototype.resolvedOptions()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/resolvedOptions) API.

A good snippet to try out in the console devtools of your browser of choice is:

```
const options = new Intl.DateTimeFormat("de-CH", { hour: 'numeric' }).resolvedOptions(); 

options.hour12 --> This will be true / false depending on the locale set
```

Please replace "de-CH" with whichever [ISO language code](https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes) preferred. In Firefox I believe you can also change the preferred language in the about:config as well!

rwood-moz:
> Please replace "de-CH" with whichever [ISO language code](https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes) preferred. In Firefox I believe you can also change the preferred language in the about:config as well!

Ahhh I see, and certain locales support/use 24 hour time. In Firerfox I just had to go into `Settings => Languages` and I changed the `Preferred language => English (GB)` and that worked. Now back in Appointment I see 24 hr time on the main dashboard calendar and also on the share link/bookee page too. Nice, thank you for helping me understand that better. I'm just going to run the E2E tests now on this branch before approving.

rwood-moz (approved):
Yep LGTM and the E2E tests are good. Note that the E2E tests will pass regardless of the locale/time format because the Appointment `Bookings` page lists the appointments in 12 hour format still. If that ever is to change, then we will need to update the E2E tests to force Playwright to use the `EN (US)` locale instead of the current system setup. We'll leave it as/is for now anyway since that's not an issue ATM.

davinotdavid:
@rwood-moz perfect, thank you for testing! Merging!

devmount:
@rwood-moz @davinotdavid Thanks so much for finishing this! 🤗

PR #1806, opened 2026-08-25
Ruff format + add --check to ruff format step in validate.yml

davinotdavid:
Currently, `main` fails a `ruff format --check` but it doesn't error in CI since we are silently formatting with `ruff format`.

These formatting changes are not backtracked into the code itself so we ended up in a place were lots of files are not formatted with ruff.

In this PR I've formatted all the files and added `--check` to the CI step so that this is enforced.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -84,7 +84,7 @@ jobs:
 
       - name: Backend formatting check
         run: |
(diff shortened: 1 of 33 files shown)


markstos (approved):
I'm in favor.

## QA Log

- Reviewed notes. Trusting that the diff is all formatting, as advertised.

PR #1812, opened 2026-09-01
Add minimal OpenGraph tags

### LINKED ISSUE #629: Add OpenGraph meta tags
Following https://github.com/thunderbird/appointment/pull/627, we should consider how our site looks when it is shared

davinotdavid:
## What changed?

Added minimal / simple OpenGraph tags

## Why?

We currently don't have any OpenGraph tags so any Appointment links shared don't render any extra UI. This may cause recipients to be wary of them.

## Limitations and Notes

- As of right now we can't do a booking page specific open graph tags because we are doing a SPA approach without any server side rendering. Similarly, i18n would also be challenging in this scenario.

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/629

## Screenshots

Example in Slack (tunneling localhost through ngrok):

image


frontend/index.html
```diff
@@ -8,6 +8,22 @@
...
+
+    <!-- Open Graph -->
+    <meta property="og:type" content="website" />
+    <meta property="og:site_name" content="Thunderbird Appointment" />
+    <meta property="og:title" content="Thunderbird Appointment" />
+    <meta
+      property="og:description"
+      content="Plan less. Do more. Thunderbird Appointment makes it easy to find a time to meet. So you can ditch the admin and streamline your day."
+    />
+    <meta property="og:image" content="/android-chrome-512x512.png" />
```
devmount, line 24: I believe this is specified to be an absolute URL to work as preview image (see https://ogp.me/). We could just do:

```suggestion
    <meta property="og:image" content="https://appointment.tb.pro/android-chrome-512x512.png" />
```

I checked, this is indeed publicly available. But we'd have to hardcode the URL here or find a way to inject it (which might be tricky because this needs to be working with crawlers getting the image for social previews).

Aaaand: it doesn't look very good 😅 It's kinda blurry and the square might not work very well for social previews...

Image

Maybe we can ask design for a more suitable preview image?

devmount (commented):
@davinotdavid Thank you for adding og tags! I found an issue with the `og:image` tag, but I'm not sure if it's a blocker - so only commenting for now.

frontend/index.html
davinotdavid: Interesting! I thought that it would attempt to resolve from the /public assets for relative links? The test screenshot that I did was serving my localhost through ngrok and copying the ngrok link into Slack but I wonder if the behavior would be different for a more realistic setup.

Agreed on the image, not ideal for sure so asking Design on some input. Thanks!

devmount: Ah I'm sure most of the platforms will do that lookup, it just seems like it was never specified that way, so I'm not sure if we can/should rely on it.

image

I mean, we can always just try it out and see, if we need to update this in the future. Nothing critical.

davinotdavid: Hmmm alright, I think that's for the best and the image is assumed to exist once this gets deployed to production so all good. The only caveat from this PR is that I've updated the code to be `appointment.tb.pro/` and that will not work in stage before it gets to prod (once it's there it should work fine as you've mentioned).

davinotdavid:
@devmount Updated the og:image tag and the image itself for a better one according to design input. Thanks for the review!

frontend/index.html
devmount: Oh that's right, won't work on stage. But we don't need links to stage shared anyway 😂 Thanks!

devmount (approved):
Thanks a lot, can't find anything else. Let's go!

PR #1814, opened 2026-09-02
Add robots.txt and update sitemap.txt

### LINKED ISSUE #628: For SEO purposes, do we need anything else in sitemap? Would we benefit from robots.txt?
Following https://github.com/thunderbird/appointment/pull/627, we should see if we can benefit from adding other routes. Also, we don't have a robots.txt, which could be helpful for requesting that they don't scour unnecessary pages

davinotdavid:
## What changed?

- Updated the sitemap.txt from the old .day domain to the .tb.pro
- Added a robots.txt file with some disallow rules (primarily for the /user/ booking links)

## Why?

This is mostly to reduce search crawler visibility for public pages like the /user/ booking links.

## Limitations and Notes

We moved away from the cloudfront-rewrite function in favor of nginx but just for completeness, made the change there too. The current nginx configuration should already cover and copy the new robots.txt file over.

## Applicable Issues

Fixes https://github.com/thunderbird/appointment/issues/628

### DIFF
--- a/frontend/public/robots.txt
+++ b/frontend/public/robots.txt
@@ -1,0 +1,23 @@
+User-agent: *
+
+# Booker pages render a subscriber's name and booking slug
+Disallow: /user/
+
+# Contains a single-use auth token in the URL
+Disallow: /post-login/
+
+# Authenticated areas
+Disallow: /dashboard
+Disallow: /availability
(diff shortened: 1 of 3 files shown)


devmount (approved):
Looking great, thanks for taking care of that!

PR #1817, opened 2026-09-04
Disable buttons while refreshing link in Refresh Link modal 

### LINKED ISSUE #1559: Refresh Link modal should block clicking "Save" more than once
Currently the Refresh Link modal on the availability settings page does not have a loading state or feedback once you click on "Save". This led to me clicking on "Save" a few times not realizing it was doing anything. 

Perhaps we can disable it after the first click? (Same with cancel as they cannot undo that request.)

Image

davinotdavid:
## What changed?

- Blocking the cancel / confirm buttons while loading since we can't actually `cancel` / rollback the action after you clicked confirm.
- Added error handling in case any of the two API requests related to this action fails.

## Why?

- Users were able to spam click confirm and trigger a lot of refreshes. This is quite common as it may take a while for the operation to complete.

## Applicable Issues

Closes https://github.com/thunderbird/appointment/issues/1559

## Screenshots

[Video in action]

https://github.com/user-attachments/assets/6339d0df-7526-4230-95a6-276bcc11f44f

[Error in case refreshing the link fails]

Screenshot 2026-09-04 at 8 48 24 AM

[Error in case refreshing the link works but we fail to reload the user info]

Screenshot 2026-09-04 at 8 48 51 AM

### DIFF
--- a/frontend/src/locales/de.json
+++ b/frontend/src/locales/de.json
@@ -33,6 +33,8 @@
     "noConnectedCalendars": "Kalendereinrichtung zum Fortfahren erforderlich. {0}.",
     "noConnectedCalendarsLink": "Zu den Einstellungen wechseln.",
     "noSpecialCharacters": "Bitte die Sonderzeichen entfernen und nochmal versuchen.",
+    "refreshLinkError": "Dein Link konnte nicht aktualisiert werden. Bitte nochmal versuchen.",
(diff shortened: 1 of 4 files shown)


rwood-moz (approved):
Code LGTM. Gave it a try locally and it worked for me.

devmount (approved):
Thanks Davi! The German sounds natural, great job 👏🏻

PR #1822, opened 2026-09-08
Improve availability page performance

devmount:
## What changed?

This PR contains a fix for long loading times on the availability page.

## Why?

## How to test

4. See the `http://localhost:5000/me/calendars?only_connected=false` called twice
5. Now check out this branch and repeat steps 2-4
6. See the `http://localhost:5000/me/calendars?only_connected=false` called only once

## Limitations and Notes

Background: The parent AvailabilityView/index and it's child component AvailabilitySettings/AvailabilityCalendarSelect both were using the `availabilityStore` without waiting for it to be initialized.

There is no easy way to test this (I think) so I didn't write a regression test.

## Applicable Issues

Related to #1371 

## Screenshots

No UI changes

### DIFF
--- a/frontend/src/views/AvailabilityView/index.vue
+++ b/frontend/src/views/AvailabilityView/index.vue
@@ -15,6 +15,7 @@ import { DateFormatStrings } from '@/definitions';
 import AvailabilitySettings from './components/AvailabilitySettings/index.vue';
 import BookingPageDetails from './components/BookingPageDetails/index.vue';
 import BookingPageLink from './components/BookingPageLink/index.vue';
+import LoadingSpinner from '@/elements/LoadingSpinner.vue';
 
 const { t } = useI18n();
 const dj = inject(dayjsKey);
@@ -181,6 +182,9 @@ export default {
         </div>
       </div>
     </form>
+    <div v-else class="availability-loading-container">
+      <loading-spinner />
+    </div>
 
     <div class="footer-save-panel" v-if="isDirty">
       <link-button @click="onRevertChanges" :disabled="savingInProgress">
@@ -229,6 +233,12 @@ export default {
   grid-template-columns: 1fr;
   gap: 2rem;
   margin-block-end: 2rem;
+}
+
+.availability-loading-container {
+  display: flex;
+  justify-content: center;
+  align-items: center;
 }
 
 .page-content-right {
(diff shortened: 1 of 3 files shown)


frontend/src/views/AvailabilityView/index.vue
```diff
@@ -164,7 +164,7 @@ export default {
       </template>
     </notice-bar>
 
-    <form ref="availability-form" @submit.prevent>
+    <form v-if="isLoaded" ref="availability-form" @submit.prevent>
```
davinotdavid: Hmm I think the v-if would cause a big layout shift here right? I wonder if it would be best to instead show a blocking loading spinner of sorts, like what we are doing in the Dashboard for the WeekCalendar?

devmount: Good catch, there is a small shift:

https://github.com/user-attachments/assets/3ffb117b-7672-4812-b79f-5f3f2e6d4d00

I'll add a spinner

devmount: @davinotdavid Alright, added a loading spinner now (it shows very short, but who knows how long that is on production - so this was a great addition, thanks!):

https://github.com/user-attachments/assets/ec1314a2-fa4c-4786-a022-ba01abaa9d92

devmount:
Whoopsi I accidentally pushed changes from my potentially third PR on the availability performance coming soon. Reverted it, sorry about the commit mess 😅

frontend/src/views/AvailabilityView/index.vue
davinotdavid: Oh sorry that's not what I meant, if we were to load either the loading spinner or the Availability, it would be kind of worse layout-shift wise since we would be showing a loading and _then_ replacing it with the full content (it would be nicer for the user though as they would at least have some feedback for slower connections).

What I meant was to still load the form as usual but overlay the loading on top of it, like we are doing in the WeekCalendar here:


https://github.com/user-attachments/assets/1172c19f-9e11-4366-85ce-3e7ae14b49ab

devmount: Oh I see, well yes, that is indeed better 😂 I'm on it

devmount: @davinotdavid Alright, I now removed the v-if completely as well as the onMounted from the AvailabilityCalendarSelect that was interfering and not necessary.

Here is the page overlay spinner now:


https://github.com/user-attachments/assets/ac3c12ce-57ea-4106-82fe-f5b83093e783

davinotdavid (approved):
LGTM, thanks! Tested locally and checked code diff.

PR #1823, opened 2026-09-08
Improve availability page performance: Initialize stores in parallel

devmount:
## What changed?

This PR contains another fix for bad performance on the availability page. This time I improved a consecutive await calls on 3 stores in the availability-store and made them run in parallel.

## Why?

## How to test

4. See the different `http://localhost:5000/*` (calendars, external connections, schedule) requests running after each other
5. Now check out this branch and repeat steps 2-4
6. See the `http://localhost:5000/*` requests happening at the same time:
    image

You can look at the overall page loading time average. This reduced the loading time by around 200ms on my machine.

## Limitations and Notes

I removed all the `if (!is*StoreLoaded.value)` conditions, since the stores themselves already do that check, resulting in a no-op in case it's already loaded. But please double check that 😇 

There is no easy way to test this (I think) so I didn't write a regression test (again 😅).

## Applicable Issues

Related to #1371 

## Screenshots

No UI changes

### DIFF
--- a/frontend/src/stores/availability-store.ts
+++ b/frontend/src/stores/availability-store.ts
@@ -32,23 +32,12 @@ export const useAvailabilityStore = defineStore('availability', () => {
     const scheduleStore = createScheduleStore(fetch);
     const externalConnectionStore = createExternalConnectionsStore(fetch);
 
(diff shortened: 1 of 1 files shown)


davinotdavid (approved):
Oh neat! Good catch, thanks!

PR #1824, opened 2026-09-08
Improve general request performance

devmount:
## What changed?

This is the third (and last for today I promise 😂) PR improving the availability page performance. This one is not scoped to the availability page in particular, but to EVERY single request we made, that needs a database lookup.

We use `get_db()` as a per-request dependency, which is fine. But we were calling SQLAlchemy's `create_engine()` via `get_engine_and_session()` there, causing to open a whole new connection pool without looking for an already existing one.

This PR sets a global engine and session (like we do with the redis instance) and checks, if that already exists before creating a new one.

## Why?

## How to test

4. See the different `http://localhost:5000/*` (calendars, external connections, schedule) requests and their timings
    image
6. Now check out this branch and repeat steps 2-4
7. See the `http://localhost:5000/*` requests being faster:
    image

## Limitations and Notes

This is a very heavy impact change. So please double check if I did that right! Existing tests were running fine though.

## Applicable Issues

Related to #1371 

## Screenshots

No UI changes

### DIFF
--- a/backend/src/appointment/dependencies/database.py
+++ b/backend/src/appointment/dependencies/database.py
@@ -3,7 +3,7 @@ import os
 import time
 from typing import Optional
 
-import sentry_sdk.metrics
+import sentry_sdk
 from redis import Redis, RedisCluster
 from redis.backoff import ExponentialBackoff
 from redis.retry import Retry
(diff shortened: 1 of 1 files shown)


davinotdavid (approved):
LGTM!

## QA Log

- Code diff
- Manually tested locally and it does seem to improve performance!

PR #1827, opened 2026-09-09
Consolidate E2E desktop and mobile tests into one source

rwood-moz:
## What changed?
Consolidating the `/e2e/tests/desktop` and `/e2e/tests/mobile` tests into one common source in `/e2e/tests`. The tests will remain running on each of the desktop, android, and iOS platforms as they do now, but just use common test source.

`/e2e/tests/desktop/book-appointment.spec.ts` and `/e2e/tests/mobile/book-appointment.spec.ts` to => `/e2e/tests/book-appointment.spec.ts`

`/e2e/tests/desktop/settings-account.spec.ts` and `/e2e/tests/mobile/settings-account.spec.ts` to => `/e2e/tests/settings-account.spec.ts`

`/e2e/tests/desktop/settings-connected-apps.spec.ts` and `/e2e/tests/mobile/settings-connected-apps.spec.ts` to => `/e2e/tests/settings-connected-apps.spec.ts`

`/e2e/tests/desktop/settings-language.spec.ts` and `/e2e/tests/mobile/settings-language.spec.ts` to => `/e2e/tests/settings-language.spec.ts`

`/e2e/tests/desktop/settings-start-of-week.spec.ts` and `/e2e/tests/mobile/settings-start-of-week.spec.ts` to => `/e2e/tests/settings-start-of-week.spec.ts`

`/e2e/tests/desktop/settings-theme.spec.ts` and `/e2e/tests/mobile/settings-theme.spec.ts` to => `/e2e/tests/settings-theme.spec.ts`

The remaining `/e2e/tests/desktop/*.spec.ts` files currently only exist for `desktop` runs, so just moved those remaining tests (including auth.desktop.setup.ts) to `/e2e/tests`.

## Why?
For the tests that exist both on desktop and mobile, the code is mostly duplicated. Consolidating into one source for easier maintenance less code. The tests can follow different code paths on android/ios/desktop where required by checking the Playwright test info project name at runtime.

## Limitations and Notes
I worked with ChatGPT's Codex to develop this patch. I iterated many times, reviewed all the code changes and tested thoroughly.

For the book-appointment test I am also changing it so that when running on desktop browsers it will NOT be signed in when it selects the time slot on the book appointment page (this is the same current behaviour as mobile).

[...]

PR #1833, opened 2026-09-11
Fix book-appointment test when logout confirmation is not required

### LINKED ISSUE #1816: Consolidate desktop and mobile E2E tests into common source
Currently the E2E tests are split in `/tests/e2e/desktop` and `/tests/e2e/mobile` with a lot of duplicate code in both. Consolidate the tests into one source in `/tests/e2e` and use Playwright's test info project name to determine at run time if the test is running on mobile or desktop, and use appropriate code paths as necessary.

rwood-moz:
## What changed

Follow-up to the changes made in #1827 to fix an intermittent log-out issue: The book-appointment test signs out of Appointment as we want to use the bookee page and select a time slot when not signed in; after navigating to the /logout URL the test is expecting the logout confirmation page to appear (with a Logout button) but turns out it doesn't appear every time.

Updated the desktop logout flow in `book-appointment.spec.ts` to support both valid Keycloak outcomes:

- If the logout confirmation page appears, click the **Logout** button.
- If logout completes automatically, continue when the **Sign In** page appears.
- Always wait for the stable Sign In page before navigating to the public booking page.

## Why

Keycloak does not consistently display the logout confirmation page. The test previously required the Logout button, causing intermittent timeouts when Keycloak completed logout automatically.

## Limitations

This handling applies only to desktop runs because mobile tests already begin unauthenticated. The synchronization relies on the current Keycloak Logout and Sign In controls.

## Notes

No Playwright tags, platform coverage, or booking behavior changed. The test still verifies the public booking flow while signed out and signs back in afterward to confirm the appointment was created.

I used Codex AI to write this patch.

## Applicable Issues
Fixes #1816.

## QA Log
Ran the updated book-appointment test on:

- Firefox desktop on my local dev stack ✅ 
- Firefox desktop against Appointment production ✅ 
- Chromium desktop [in BrowserStack](https://automate.browserstack.com/projects/Thunderbird+Appointment/builds/Appointment+Nightly+Tests+Chromium+Desktop/319?tab=sessions&bls_localTimezone=true) against production ✅ 
- Android Chrome on a Google Pixel 10 device [in BrowserStack](https://automate.browserstack.com/projects/Thunderbird+Appointment/builds/Appointment+Nightly+Tests+Android+Chrome/338?tab=sessions&bls_localTimezone=true) against production ✅

### DIFF
--- a/test/e2e/tests/book-appointment.spec.ts
+++ b/test/e2e/tests/book-appointment.spec.ts
@@ -72,13 +72,33 @@ test.describe('book an appointment', () => {
       // and must provide the booker's name and email address.
       await page.goto(`${APPT_URL}logout`);
 
-      // Appointment's OIDC logout redirects to Keycloak, which requires the
-      // user to confirm the logout before following post_logout_redirect_uri
-      // back to Appointment. Waiting for the button also ensures that the
-      // redirect to Keycloak has completed before the test tries to continue.
+      // Depending on the current Keycloak session, logout either stops on a
+      // confirmation page or completes immediately and returns to Sign In.
+      // Wait for either valid outcome instead of requiring the optional
+      // confirmation button, which makes this branch intermittent.
       const keycloakLogoutButton = page.getByRole('button', { name: 'Logout', exact: true });
-      await expect(keycloakLogoutButton).toBeVisible({ timeout: TIMEOUT_60_SECONDS });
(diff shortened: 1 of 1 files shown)


davinotdavid (approved):
Oh good catch on this, thanks! Reviewed the diff and lgtm!

PR #1835, opened 2026-09-14
Pull in APPT_BOOKEE_NAME secret to fix book-appt test on desktop CI

### LINKED ISSUE #1834: Book appointment test fails in CI because of missing GHA secret
In #1827 the book-appointment test was changed so that the test would be signed out of Appointment when it accesses the bookee page and selects the time slot. This test is failing in CI because the APPT_BOOKEE_NAME env var is not being pulled into the GHA desktop workflows, so the test doesn't fill out the name of the person requesting the time slot, and the test fails.

rwood-moz:
## What changed?
Fix the book-appointment E2E test on desktop platforms in CI by pulling in the APPT_BOOKEE_NAME env var (GHA secret).

## Why?
In https://github.com/thunderbird/appointment/pull/1827 the book-appointment test was changed so that the test would be signed out of Appointment when it accesses the bookee page and selects the time slot. That means on the booking request page, the user email and user name must be filled in when requesting a time slot. The test is failing in CI because the APPT_BOOKEE_NAME env var is not being pulled into the GHA desktop workflows, so the test doesn't fill out the name of the person requesting the time slot. The user email is pulled in so that field is fine.

On mobile GHA workflows in CI the APPT_BOOKEE_NAME is already pulled in (because on mobile we were already signed out on the select timelsot page) so it's just the desktop workflows that need to be updated.

## Applicable Issues
Fixes #1834.

## QA Log
Ran the book-appointment test in production on this branch from my machine. Also triggered the nightly E2E test Firefox desktop GHA workflow on this branch and the test passed [in BrowserStack](https://automate.browserstack.com/projects/Thunderbird+Appointment/builds/Appointment+Nightly+Tests+Firefox+Desktop/327?testListView=spec).


davinotdavid (approved):
LGTM! Checked code diff and browserstack seems happy!

PR #1839, opened 2026-09-16
Fix js-yaml E2E test dependencies dependabot alerts

### LINKED ISSUE #1838: Fix js-yaml dependabot alert from E2E dependencies
Address a bunch of [dependabot alerts](https://github.com/thunderbird/appointment/security/dependabot?q=is%3Aopen+manifest%3Atest%2Fe2e%2Fpnpm-lock.yaml+package%3Ajs-yaml) regarding an unsafe version of js-yaml being pulled in via the E2E test dependencies.

rwood-moz:
## What changed?
Add a pnpm override for js-yaml E2E test dependencies that are flagged by dependabot as unsafe.

## Why?
Address a bunch of [dependabot js-yaml alerts](https://github.com/thunderbird/appointment/security/dependabot?q=is%3Aopen+manifest%3Atest%2Fe2e%2Fpnpm-lock.yaml+package%3Ajs-yaml).

## Limitations and Notes
Used Codex AI for this patch.

## Applicable Issues
Fixes #1838.

## QA Log
Ran an Appointment E2E test from my machine with these updated packages:

- On Firefox desktop on my machine ✅ 
- On Chromium desktop [in BrowserStack](https://automate.browserstack.com/projects/Thunderbird+Appointment/builds/Appointment+Tests+Chromium+Desktop/2?tab=sessions&testListView=spec) ✅ 
- On Android Chrome [in BrowserStack](https://automate.browserstack.com/projects/Thunderbird+Appointment/builds/Appointment+Tests+Android+Chrome/57?tab=sessions&bls_localTimezone=true) ✅


davinotdavid (commented):
It seems that we need `js-yaml@3: 3.15.2` just for `js-yaml-cloudformation-schema@1.0.0`, is that right? The other peer dependencies seems to use `4.3.2`.

Just sanity checking but would it be possible to use just 4.3.2? Feels weird to have both versions installed.

rwood-moz:
> It seems that we need `js-yaml@3: 3.15.2` just for `js-yaml-cloudformation-schema@1.0.0`, is that right? The other peer dependencies seems to use `4.3.2`.
> 
> Just sanity checking but would it be possible to use just 4.3.2? Feels weird to have both versions installed.

No we need to keep both versions because yep one dependency uses 3.X and the other 4.X, I don't think it's safe to force 4.3.2 on both in case that causes a breaking change. We had to do the same for [Accounts here also](https://github.com/thunderbird/thunderbird-accounts/pull/1317/changes).

davinotdavid (approved):
I see, thanks for clarifying! Sad that we can't consolidate dependencies but all good!

PR #1843, opened 2026-09-22
feat(backend): add beat container role and SKIP_DB_MIGRATIONS flag to entry.sh

aatchison:
entry.sh dispatches on CONTAINER_ROLE, defaults to api, and the api role can skip migrations.

- CONTAINER_ROLE: api (default when unset), worker, beat, flower. Any other value exits 1.
- worker runs without embedded beat. beat is its own role; run one instance.
- SKIP_DB_MIGRATIONS=true (case insensitive) skips update-db in the api role. Anything else runs it.
- The final process runs under exec so SIGTERM reaches uvicorn or celery.
- uvicorn args are a bash array; shellcheck passes.
- docker-compose.yml: adds celery-beat, drops the ulimits workaround from celery-worker.
- backend/README.md documents both variables.

Manual step: add a CONTAINER_ROLE=beat service to legacy ECS (pulumi) before this ships, or scheduled tasks stop on stage and prod.

Refs thunderbird/appointment-deploy#19
Refs thunderbird/appointment-deploy#20


davinotdavid (approved):
lgtm!

## QA Log

- Code diff
- Manually ran containers locally

